Issue #637, opened 2014-09-29
dupe search results


fommil:
ensime-search is giving me loads of dupes. Probably just a silly brain fart in SearchService.

PR #758, opened 2015-01-02
Only flag indexer ready after commit

rorygraves:
As discussed, adding the tests highlighted a race condition which I believe this fixes.
@fommil  want to take a gander  - also see inline comment about results.

### DIFF
--- a/src/main/scala/org/ensime/core/Indexer.scala
+++ b/src/main/scala/org/ensime/core/Indexer.scala
@@ -14,8 +14,6 @@ import scala.reflect.io.{ AbstractFile, ZipArchive }
 
 // legacy types
 case class TypeCompletionsReq(prefix: String, maxResults: Int) extends RPCRequest
-case class SourceFileCandidatesReq(enclosingPackage: String, classNamePrefix: String) extends RPCRequest
-case class AbstractFiles(files: Set[AbstractFile]) extends RPCRequest
 
 //@deprecated("there is no good reason for this to be an actor, plus it enforces single-threaded badness", "fommil")
 class Indexer(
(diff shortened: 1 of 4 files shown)


src/test/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -66,6 +66,33 @@ class BasicWorkflow extends FunSpec with Matchers {
...
+        // public symbol search - java.io.File
+
+        val javaSearchSymbol = project.rpcPublicSymbolSearch(List("java", "io", "File"), 2)
+        javaSearchSymbol match {
+          case SymbolSearchResults(List(
+            TypeSearchResult("java.awt.image.ImageObserver", "ImageObserver", 'class,
+              Some(_)),
+            TypeSearchResult("java.io.File", "File", 'class,
+              Some(_)),
+            TypeSearchResult("org.ensime.util.FileUtilSpec", "FileUtilSpec", 'class, None))) =>
```
rorygraves: @fommil These indexer results seem wrong - two items:
1. It asks for 2 items and gets 3
2. Search for "java" "io" "File" returns j.a.i.ImageObserver and FileUtilSpec

I accept I may not understand the protocol spec but this feels a bit random.

src/main/scala/org/ensime/core/Indexer.scala
```diff
@@ -53,11 +51,6 @@ class Indexer(
   }
 
   override def receive = {
-    case SourceFileCandidatesReq(enclosingPackage, classNamePrefix) =>
```
rorygraves: This was dead and uncallable code so I ripped it out

src/test/scala/org/ensime/intg/BasicWorkflow.scala
fommil: 1. that looks like a bug. The max number must not be going through to lucene. Also see #637 
2. this is actually reasonable: "io" will hit the `ImageObserver` via CamelCase... but `java.io.File` should be a much stronger hit.

fommil:
good catch! I was too close to it to see it.

src/test/scala/org/ensime/intg/BasicWorkflow.scala
rorygraves: I assumed on first reading that it was an expression like j.i.File where the subparts could be camel or partials but all would have to match.
I will raise a new ticket for the count break (I looked, it appears to pass all the way to Lucene)

PR #762, opened 2015-01-03
Replace regexp based localization with AST-based in Completion.

aemoncannon:
* Add more tests.
* Fixes at least two outstanding completion issues.
* Gives us a more stable base for improving completion.

### DIFF
--- a/src/main/scala/org/ensime/core/Completion.scala
+++ b/src/main/scala/org/ensime/core/Completion.scala
@@ -4,8 +4,9 @@ import akka.pattern.Patterns
 import akka.util.Timeout
 import org.ensime.model.CompletionInfoList
 import scala.collection.mutable
-import scala.concurrent.Await
+import scala.concurrent.{ Future, Await }
 import scala.reflect.internal.util.{ SourceFile, BatchSourceFile }
+import scala.reflect.runtime.universe.{ showRaw }
 import org.ensime.util.Arrays
 import org.ensime.model.{ CompletionInfo, CompletionSignature, SymbolSearchResults }
 import scala.concurrent.duration._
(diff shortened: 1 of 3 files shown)


src/main/scala/org/ensime/core/Completion.scala
```diff
@@ -110,132 +114,92 @@ trait CompletionControl {
...
-  // to parenthesis. Ignores the possibility of comments and strings.
-  private def parenBalanced(s: String): Boolean = {
-    var i = 0
-    var p = 0
-    while (i < s.length) {
-      val ch = s.charAt(i)
-      if (ch == '(') p += 1
-      else if (ch == ')') p -= 1
-      i += 1
+    val defaultPrefix = IdentRegexp.findFirstMatchIn(preceding) match {
```
rorygraves: I accept I may be showing complete failure to understand the code, but surely you want the lastMatchIn - you are a a point P and you want the nearest Ident backwards, rather than the first match from 100 characters previously?

rorygraves:
You should squash the commits.

src/main/scala/org/ensime/core/Completion.scala
aemoncannon: The regex is anchored by an '$', so this gives the desired effect, whether or not it's the best way. Where can I find this 'lastMatchIn' ?

rorygraves: Oh I see - told you I had missed something ;) (the $ anchor).  Not sure there is a lastMatchIn, ;)

aemoncannon:
Squashed.

fommil:
cool... does this mean we can do tests like the scala-ide?

rorygraves:
I was thinking about this last night - the code (whilst correct) does not parse in my brain nicely - even knowing the behaviour I find it non intuitive to read - how about extracting a method so instead of 

```scala
IdentRegexp.findFirstMatchIn(preceding) match {
```

you have 

```scala
def nearestPrecedingMatch(regex: Regex, preceding: String) regex.findFirstMatchIn(preceding)
```
and add a comment on the regex expressions to say 'Use $ to find the last matching case'

aemoncannon:
Hmm, might that be a little misleading, since what we really want is the
*first* match that happens to extend to the end (That is, the longest valid
prefix)?

aemoncannon:
@fommil What are these fancy scala ide tests of which you speak?

rorygraves:
@aemoncannon I think @fommil comment is actually in the wrong place.  Knowing that we share a bit of logic with the scala-ide and knowing the recently release version 4.0, I thought it might be worth exploring whether there were any fixes we could port over.

The inspection lead to two thoughts:
1) They have a lot of tests around the debugging and PC interactions - which we could potentially borrow.
2) It might be possible to bring over PC interaction interface closer in line with theirs making sharing easier.

tldr; scala-ide has a large number of tests and fies we might be able to reuse to improve ensime.

They have a really neat set of tests which we could

fommil:
@aemoncannon rory was looking at the eclipse tests recently and seen some cool stuff that they do to allow tests to be completely driven by the example sources (e.g. completions and their expected responses all defined in comments).

rorygraves:
@aemoncannon Fair enough

rorygraves:
@aemoncannon  I'm happy to merge unless @fommil  has any more PR comments?

fommil:
:shipit:

aemoncannon:
woo!
@fommil @rory Completely source-driven tests sounds like a great idea!

aemoncannon:
@fommil Though I think this change is orthogonal to testing strategy.

@fommil @rorygraves RE: scala-ide - I think we should seriously consider going through the pain of changing our source license. Otherwise a lot of nice sharing will have to go out the window. Even this source, which borrows an overall approach from scala-ide, is probably treading on thin ice.

fommil:
I thought we were using their derived licence for everything that borrows from them?

aemoncannon:
@fommil - Ah. Yes, you're right. I need to update LICENSE.txt.

aemoncannon:
@fommil  Done.

PR #768, opened 2015-01-05
close #613 until it is fixed upstream

### LINKED ISSUE #613: is vfs2 ignoring directories that are deleted and recreated?
it looks like an `sbt clean compile` is enough to wipe out the classfile watcher logic... is vfs2 really ignoring recreated folders? Epic fail if it is.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4684656-is-vfs2-ignoring-directories-that-are-deleted-and-recreated?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github).
</bountysource-plugin>

fommil:
Each file watcher now consists of two watchers:

1. doing what we want it to do (watch for source/classfile changes)
2. a workaround looking for recreations of the folders that we are watching, resetting `1.` if that happens

to workaround an upstream bug.

### DIFF
--- a/src/main/scala/org/ensime/indexer/FileWatchers.scala
+++ b/src/main/scala/org/ensime/indexer/FileWatchers.scala
@@ -1,11 +1,14 @@
 package org.ensime.indexer
 
+import java.io.File
 import akka.event.slf4j.SLF4JLogging
 import org.apache.commons.vfs2._
 import org.apache.commons.vfs2.impl._
 import org.ensime.config._
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
+
+import pimpathon.file._
 
 trait ClassfileListener {
   def classfileAdded(f: FileObject): Unit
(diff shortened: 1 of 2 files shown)


rorygraves:
Having apparently pushed up coverage above 75% maybe you should update the build expectations?
I'm ignoring the fact that travis borked again.

src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -57,12 +53,61 @@ class ClassfileWatcher(
...
+    log.info("Setting up new file watchers")
+    // must remove then add to avoid leaks
+
+    for {
+      d <- config.targetClasspath
+    } {
+      fm.removeFile(d)
+      fm.addFile(d)
+
+      // for-comp inside for-comp madness, w00t!
```
rorygraves: I think you can flatten this into one by doing 

_ <- fm.removeFile(d)
_ <- fm.addFile(d)

Not sure if that is better ;)

fommil:
god damn you, travis

rorygraves:
@fommil  I'm tempted to simply disable it for CI builds at this point and only keep it for the release builds, it basically fails every time now.

fommil:
it's even failing for the release builds. Such a pretty service, such terrible server instances.

rorygraves:
bite the bullet and configure jenkins to do the publishes?

fommil:
yeah, let's move to Jenkins 100%.

rorygraves:
ok, will sort that out over the next few days.

src/main/scala/org/ensime/indexer/FileWatchers.scala
fommil: nope, you can't chain that sort of thing (it would have been nice, because it would work for the whole block)

fommil: oh, hold on... I think it might work with `=` instead of `<-`

rorygraves:
@fommil  I have disabled travis PR building - planning on updating the travis build to simply do the build and deploy (no test), so it is only responsible for the deploy.  On a positive note this means that we can switch of the apt-get stuff and do faster builds.

fommil:
leaving testing on is good.

Issue #772, opened 2015-01-07
don't auto complete in comments / documentation


fommil:
@aemoncannon I'm not sure how easy/hard this would be to fix. If we're passing the cursor location along with the object to complete, we should be able to ask the presentation compiler if we're in a comment block or not.

aemoncannon:
Hmmm, this depends a lot on what compiler does with the docs. Will look
into it.

@aemoncannon <https://github.com/aemoncannon> I'm not sure how easy/hard
this would be to fix. If we're passing the cursor location along with the
object to complete, we should be able to ask the presentation compiler if
we're in a comment block or not.

—
Reply to this email directly or view it on GitHub
<https://github.com/ensime/ensime-server/issues/772>.

fommil:
we might also be able to leverage the `scala-mode2` comment detection

fommil:
it would be pretty sweet if we could turnon flyspell completion in comments.

fommil:
also, when  completing in strings it would be good if it was restricted to the interpolated scala bits only

aemoncannon:
@fommil Working well for you?

fommil:
Yes it is, thank you!

PR #783, opened 2015-01-10
Don't generate completions inside string literals. Takes care of one aspect of ensime/ensime-server#772

aemoncannon:


src/main/scala/org/ensime/core/Completion.scala
```diff
@@ -66,7 +66,7 @@ trait CompletionControl {
 
     val context = x.get match {
       case Left(tree) => {
-        logger.debug("Completing at tree:" + showRaw(tree))
+        logger.debug("Completing at tree:" + showRaw(tree).take(200))
```
rorygraves: Can I suggest you add:
```scala
if(logger.debugEnabled)
  logger.debug("Completing at tree:" + showRaw(tree).take(200))
```
Which saves the string generation (I'm guessing it can be huge) if its not needed.

rorygraves:
@aemoncannon Change looks really good - a potential minor tweak - take it if you fancy  - otherwise go ahead and merge (or make change and merge)

aemoncannon:
That's a good call. The strings are sometimes huge. summaryString looks like it was made exactly for this purpose.

rorygraves:
Cool stuff

PR #791, opened 2015-01-17
Travis fixes

fommil:


fommil:
@rorygraves I've turned Travis back on for PRs because it should be able to handle the load now :smile:

rorygraves:
Cool - will be interesting to see how it performs :)

fommil:
it has caching now, so should be quite good

Issue #155, opened 2011-04-06
Don't write files in place on completion request


hjz:
ensime-write-buffer-to-temp-file seems to be missing from ensime.el in 5.0

aemoncannon:
@edani
FYI I'm working on a cleanup of the completion stuff. Mostly just separating auto-complete backend from company backend. Afterward I want to give this a serious look.

edani:
@aemoncannon If you can wait a few days, I made more server-side changes and I need to push client changes to go with them.

aemoncannon:
@edani You mean wait on the re-org?

edani:
@aemoncannon  I meant wait on looking at this issue (#155). I'll have an update on this shortly.

aemoncannon:
@edani
Okey dokey.

FWIW I just ran these benchmarks:

(defun bench-file ()
    (benchmark-run 1000
      (ensime-eval `(swank:typecheck-file ,buffer-file-name))))


(defun bench-contents ()
    (benchmark-run 1000
      (ensime-eval `(swank:typecheck-file ,buffer-file-name
,(ensime-get-buffer-as-string)))))


;; Result of 3 runs each, where each run is: (total-elapsed-time #of-GCs
#total-GC-time)

;; to disk
(2.057487 2 0.14408899999999925)
(2.194499 3 0.2171789999999998)
(2.070298 2 0.1403800000000004)

;; by contents
(11.508381 23 1.6244739999999993)
(11.635637000000001 23 1.584938000000001)
(11.490751 22 1.5078890000000005)

aemoncannon:
This was on Analyzer.scala with ~11k characters.

rorygraves:
Wow, thats quite a difference, I guess this is encoding and decoding the buffer in emacs?

aemoncannon:
I suspect its the string creation yeah, but we can tease out the culprit
with profiler

fommil:
when I finally get round to moving us fully over to S-Express for the protocol, the contents-over-the-wire might be better.

rorygraves:
I'm guessing it's the enacts rather than jvm side.

aemoncannon:
Here's a dump from the profiler (doesn't add up to 100 because of keyboard
handling etc that's not part of the bench). The gist is that it *looks*
like it's spending most of the time waiting on the socket.

   - bench-contents                                            1826  47%
      - let                                                      1826  47%
       - list                                                    1826  47%
        - -                                                      1826  47%
[...]

Issue #552, opened 2014-08-12
no fork version of the ensime sbt server starter


fommil:
The current server starter uses an sbt start script that then forks the server.
This means creating two `java` instances before the compilers are started, which may be considered expensive on lower-end machines.

It would be good to pass the java options (prefixed with `-J`) to `sbt` (assuming paulp's script is used) or to instantiate sbt to create a start script, which is then called.

fommil:
I was thinking about using https://github.com/sbt/sbt-start-script for this but it copies the dependencies.

A cleaner way to do it might be for the callout to sbt to simply write out the classpath to a file, then read that into emacs and call `exec` directly.

That approach will require attaching a second process to a `comint` after its first process terminates. @aemoncannon @edani is that possible?

aemoncannon:
I have no idea! : ) but I bet its doable.

fommil:
sweet, I'll look into the sbt side and share a branch with you both when my elisp foo runs out.

edani:
I like that idea too. Also the great thing is that the classpath can be cached so e.g. in the emacs test suite there's no need to run stb more than once.

fommil:
@edani @aemoncannon 

This should produce a file `classpath` which contains the OS-dependent string for the classpath part.

https://github.com/fommil/ensime-emacs/tree/one_process

fommil:
sorry, missed a bit... pushing change now

fommil:
actually that's missing the tools.jar, but I think I can fix that.

aemoncannon:
Oh man that would be sweet.

fommil:
@aemoncannon @edani have either of you had any time to look at my branch? I did some research on comint-mode and couldn't find any documentation on how to add a callback to the process for when it dies. Also, I noticed that we return the process from the function that starts the server, and that might be used elsewhere. However, there is no reason to do this because comint allows looking up the process associated to a buffer separately, so we could return the buffer instead.

fommil:
If I could work out how to attach a callback to a process when it finishes, I could polish this off.

aemoncannon:
@fommil look for set-process-sentinel

fommil:
awesome! I can finish this off now. If anybody fancies taking it, please ping me. I'm moving house soon anyway, so likely AWOL for a few weeks.

fommil:
[...]

PR #794, opened 2015-01-18
Protocol changes to support #155

edani:
See the commit message for details. This goes towards fixing #155 and a few other places where we don't want to save sources to disk.

I don't have ensime-emacs changes ready yet, I'd like @aemoncannon's opinion on this first.

### DIFF
--- a/src/main/scala/org/ensime/core/Completion.scala
+++ b/src/main/scala/org/ensime/core/Completion.scala
@@ -50,7 +50,6 @@ trait CompletionControl {
     }
 
     val constructing = ConstructingRegexp.findFirstMatchIn(preceding).isDefined
-    logger.debug("In constructing context: " + constructing)
 
     val (src, p, patched) = if (defaultPrefix.isEmpty) {
       // Add a fake prefix if none was provided by the user. Otherwise the
--- a/src/main/scala/org/ensime/server/protocol/swank/SwankProtocol.scala
+++ b/src/main/scala/org/ensime/server/protocol/swank/SwankProtocol.scala
@@ -22,9 +22,16 @@ class SwankProtocol(actorSystem: ActorSystem,
   implicit val rpcExecutionContext = actorSystem.dispatchers.lookup("akka.swank-dispatcher")
 
   /**
-   * Protocol Version: 0.8.10 (Must match version at ConnectionInfo.protocolVersion)
+   * Protocol Version: 0.8.11 (Must match version at ConnectionInfo.protocolVersion)
    *
    * Protocol Change Log:
+   *   0.8.11
+   *     Some calls now take (:file "filename" :contents "xxxx")
+   *       as a way to transmit source files:
+   *         - swank:typecheck-file
+   *         - swank:completions
+   *     Made the "reload" parameter of swank:completions optional
+   *     Added swank:format-one-source
    *   0.8.10
    *     Remove the config argument from init-project. Server loads
    *     configuration on its own.
@@ -130,6 +137,17 @@ class SwankProtocol(actorSystem: ActorSystem,
   ////////////////////
   // Datastructures //
   ////////////////////
+
+  /**
+   * Doc DataStructure:
+   *   SourceFileInfo
+   * Summary:
+   *   A source file to be analyzed by the presentation compiler.
+   *   (
+   *     :file    //String:A filename, absolute or relative to the project
+   *     :contents  //String, optional: if present, this string is loaded instead of the actual file's contents.
+   *   )
+   */
 
   /**
    * Doc DataStructure:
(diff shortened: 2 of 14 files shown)


edani:
@aemoncannon, here is another benchmark I tried, I think it reflects better what we're trying to measure. Place the cursor in a completion-worthy spot then:

```elisp
(defun bench-file ()
  (benchmark-run 100
    (ensime-write-buffer)
    (ensime-eval
     `(swank:completions
       ,buffer-file-name
       ,(ensime-computed-point)
       10
       t
       t))))
```

```elisp
(defun bench-contents-in ()
  (let ((tmp (make-temp-file "c:/tmp/test")))
    (benchmark-run 100
      (ensime-write-buffer tmp)
      (ensime-eval
       `(swank:completions
         (:file ,buffer-file-name :contents-in ,tmp))
         ,(ensime-computed-point)
         10
         t
         t)))))
```

```elisp
(defun bench-contents ()
  (benchmark-run 100
    (ensime-eval
     `(swank:completions
       (:file ,buffer-file-name :contents ,(ensime-get-buffer-as-string))
       ,(ensime-computed-point)
       10
       t
       t))))
```

Crucially, `bench-file` and `bench-contents-in` actually write the buffer to disk!  It turns out that, on Windows at least, this is where a lot of the time is spent.  Also I used `swank-completions` in the benchmark because it does more work and that's the one cal that's used several times per second.

In my tests, performance is a toss-up between the three methods when I write to a SSD and `bench-contents` performs better when I write to a HDD. I'll test on linux tomorrow, I'm curious to see if in-memory tmpfs makes a difference.

fommil:
good stuff!

As this involves a change in the protocol, we'll have to have the emacs client and 2.10, 2.9 branches ready to take the changes ASAP to avoid unfixable breakages.

BTW, will the client complain if the protocol version mismatches? I can imagine the most common problem will be that sbt updates the server but the user hasn't updated on MELPA yet, so the client is sitting with the old protocol version.

I'll put together a 2.10 and 2.9 branch today since you're taking the process/buffer (hard!) bits of #552

fommil:
@edani 2.10 and 2.9 are all up to date. Once the review is finished, could you please prepare a 2.10 and 2.9 branch of this and raise PRs with them, that would be great because we can merge all four at the same time. If it's a problem, I can raise PRs for you, just let me know (I'm probably going to be AWOL for a week or two with a house move)

aemoncannon:
@edani Good catch about writing to disk in the bench. I'll re-run your
benches on Mac when I get a chance.

If speed is comparable, put me in the pro-contents camp.

edani:
@fommil  The protocol changes should be backward compatible. I'll double check. In any case I'll make the 2.9 and 2.10 PRs.

fommil:
ok cool. What about the bump in version number, doesn't that necessitate the server upgrading at the same time?

I don't actually know what the client does if the version doesn't match.

aemoncannon:
@fommil I don't think we act on it at all right now.

src/main/scala/org/ensime/core/Completion.scala
```diff
@@ -50,7 +50,7 @@ trait CompletionControl {
     }
 
     val constructing = ConstructingRegexp.findFirstMatchIn(preceding).isDefined
-    logger.debug("In constructing context: " + constructing)
+    // logger.debug("In constructing context: " + constructing)
```
rorygraves: If this line is junk, can we simply remove it rather than commenting it out?

src/main/scala/org/ensime/server/protocol/swank/SwankProtocol.scala
```diff
@@ -695,7 +715,7 @@ class SwankProtocol(actorSystem: ActorSystem,
        *   swank:unload-all
        * Summary:
        *   Remove all sources from the presentation compiler. Reset the project's symbols
-       *    to the ones defiled in .class files.
+       *    to the ones defined in .class files.
```
rorygraves: lol, I quite like the original

rorygraves:
This patch looks good to me.

edani:
My plan is to have the client use the new protocol if the protocol version is >= "0.8.11" , and the old protocol otherwise, so we don't need to increase the server version. But we can if we want to.

BTW I just finished benchmarking on my ancient linux box (which I consider the bare minimum to do any scala development)  . Writing to tmpfs has a slight edge (not much) over sending contents over the socket. Writing to hard disk is 1.5x to 2x slower depending on file size.

src/main/scala/org/ensime/server/protocol/swank/SwankProtocol.scala
edani: unload-all, defiler of symbols.

fommil:
I had turned off typecheck when idle and only today turned it back on again... wow, this is an awesome feature!

PR #797, opened 2015-01-18
Scala 2.10

fommil:


fommil:
wow, there are waaaay more commits than I expected!

rorygraves:
@fommil  - yep, we have all been busy ;), and yes I often disagree with wartremover.

fommil:
Interesting, travis passed but jenkins still isn't set up for the older branches.

I think we can declare the new travis docker instances a success!

fommil:
ha! no I spoke too soon. Looks like the ParserSpec failed with a transient really complex sexpression to parse.

PR #798, opened 2015-01-18
Scala 2.9

fommil:
always painful...


fommil:
ok, **this** time the jenkins build isn't set up to handle the specificities of 2.9 :smile:

fommil:
I found an sbt problem... we're not building the 2.9.2 branch properly thanks to the sbt caching magic. I'll add a patch in a moment.

fommil:
yay

fommil:
@rorygraves maybe we should turn off Jenkins now that Travis seems to be working again?

rorygraves:
I think we should leave it on for now - there is a bunch of other stuff I would like to do which would require dependent builds and cleverness that even the new funky travis doesn't support.

rorygraves:
but still, pretty cool that using docker travis is now reliable.

fommil:
ok, thanks for that coveralls.

PR #811, opened 2015-01-29
Resuse some of the tests from Scala-IDE debugging support

rorygraves:
I think this improves the debug tests a bit, although it also highlights some shortcomings.
Longer term I think we might want to borrow a bunch of the debugging logic from scala-ide as they have a lot of neat bits to handle the edge cases and make stepping more intuitive for scala code.


fommil:
```
- should inspect a String local - h *** FAILED ***
[info]   java.lang.IllegalStateException: Unexpected value for h: DebugStringInstance("test",List(DebugClassField(0,value,char[],Array['t', 'e', 's',...]), DebugClassField(1,offset,int,0), DebugClassField(2,count,int,4), DebugClassField(3,hash,int,0), DebugClassField(4,serialVersionUID,long,-6849794470754667710), DebugClassField(5,serialPersistentFields,java.io.ObjectStreamField[],Array[]), DebugClassField(6,CASE_INSENSITIVE_ORDER,java.util.Comparator,Instance of String$CaseInsensitiveComparator)),java.lang.String,662)
[info]   at org.ensime.intg.DebugTest.inspectVar(DebugTest.scala:53)
```

aemoncannon:
Great stuff, Rory!

rorygraves:
Gahh, annoying the fields change order on different will have to add a bit more hackery.

aemoncannon:
that might be us (me) doing something silly like throwing them in a hashset
or hashmap. i'd expect the jdi order to be deterministic...

rorygraves:
Ah good point. I will do some more checks.

rorygraves:
@aemoncannon The jdi may be consist, but it is not consistent between jvm versions.

aemoncannon:
Womp womp

fommil:
cool! But maybe we ought to start separating tests into unit and integration... the CI now takes about 45 minutes :confused:

PR #818, opened 2015-02-06
Fix search results, issue stemmed from collecting results of different q...

### LINKED ISSUE #763: public symbol search count bug
From ```BasicWorkflow```

The below request returns 3 results rather than the requested 2, and I believe the order should be weighted so that java.io.File comes first.

``` scala
val javaSearchSymbol = project.rpcPublicSymbolSearch(List("java", "io", "File"), 2)
        javaSearchSymbol match {
          case SymbolSearchResults(List(
            TypeSearchResult("java.awt.image.ImageObserver", "ImageObserver", 'class,
              Some(_)),
            TypeSearchResult("java.io.File", "File", 'class,
              Some(_)),
           TypeSearchResult("org.ensime.util.FileUtilSpec", "FileUtilSpec", 'class, None))) =>
}
```

jschlather:
...ueries. Instead of doing multiple queries do a single query that ORs the original queries together.

This should resolve #763. Unit tests appear to be passing, I could write some more if there's some concern.

I also wonder if the lucene searching wouldn't benefit from something like using a lucene SearcherManager or doing near real-time searching by passing the indexwriter to the indexreader.

### DIFF
--- a/src/test/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/src/test/scala/org/ensime/intg/BasicWorkflow.scala
@@ -71,9 +71,8 @@ class BasicWorkflow extends FunSpec with Matchers {
         val javaSearchSymbol = project.rpcPublicSymbolSearch(List("java", "io", "File"), 2)
         javaSearchSymbol match {
           case SymbolSearchResults(List(
-            TypeSearchResult("java.io.File", "File", 'class,
-              Some(_)))) =>
-
+            TypeSearchResult("java.io.File", "File", 'class, Some(_)),
+            TypeSearchResult("java.awt.image.ImageObserver", "ImageObserver", 'class, Some(_)))) =>
           case _ =>
             fail("Public symbol search does not match expectations, got: " + javaSearchSymbol)
         }
--- a/src/test/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/src/test/scala/org/ensime/intg/BasicWorkflow.scala
@@ -71,11 +71,8 @@ class BasicWorkflow extends FunSpec with Matchers {
         val javaSearchSymbol = project.rpcPublicSymbolSearch(List("java", "io", "File"), 2)
         javaSearchSymbol match {
           case SymbolSearchResults(List(
-            TypeSearchResult("java.awt.image.ImageObserver", "ImageObserver", 'class,
-              Some(_)),
             TypeSearchResult("java.io.File", "File", 'class,
-              Some(_)),
-            TypeSearchResult("org.ensime.util.FileUtilSpec", "FileUtilSpec", 'class, None))) =>
+              Some(_)))) =>
 
           case _ =>
             fail("Public symbol search does not match expectations, got: " + javaSearchSymbol)
--- a/src/main/scala/org/ensime/core/Indexer.scala
+++ b/src/main/scala/org/ensime/core/Indexer.scala
@@ -24,8 +24,8 @@ class Indexer(
       name => name.fqn.endsWith("$") || name.fqn.endsWith("$class")
     }.map(typeResult)
 
-  def oldSearchSymbols(query: String, max: Int) =
-    index.searchClassesFieldsMethods(query, max).flatMap {
+  def oldSearchSymbols(terms: List[String], max: Int) =
+    index.searchClassesFieldsMethods(terms, max).flatMap {
       case hit if hit.declAs == 'class => Some(typeResult(hit))
       case hit if hit.declAs == 'method => Some(MethodSearchResult(
         hit.fqn, hit.fqn.split("\\.").last, hit.declAs,
@@ -47,7 +47,7 @@ class Indexer(
             sender ! ImportSuggestions(suggestions)
 
           case PublicSymbolSearchReq(keywords, maxResults) =>
-            val suggestions = keywords.flatMap(oldSearchSymbols(_, maxResults))
+            val suggestions = oldSearchSymbols(keywords, maxResults)
             sender ! SymbolSearchResults(suggestions)
         }
       } catch {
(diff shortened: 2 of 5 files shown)


src/test/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -71,11 +71,8 @@ class BasicWorkflow extends FunSpec with Matchers {
         val javaSearchSymbol = project.rpcPublicSymbolSearch(List("java", "io", "File"), 2)
         javaSearchSymbol match {
           case SymbolSearchResults(List(
-            TypeSearchResult("java.awt.image.ImageObserver", "ImageObserver", 'class,
```
rorygraves: @jschlather Have we broken the test here - looking up from java.io.File looks like it used to return sensible results and now returns nothing?

src/test/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -71,11 +71,8 @@ class BasicWorkflow extends FunSpec with Matchers {
...
-              Some(_)),
             TypeSearchResult("java.io.File", "File", 'class,
-              Some(_)),
-            TypeSearchResult("org.ensime.util.FileUtilSpec", "FileUtilSpec", 'class, None))) =>
```
rorygraves: Admittedly this line should never be here - the test code should not see the ensime source (I believe thats an open bug).

src/test/scala/org/ensime/intg/BasicWorkflow.scala
jschlather: @rorygraves I don't think so. It's a search call for at most 2 results that used to return three results, with the top result for java io file being java.awt.image.ImageObserver. It now returns just java.io.File, which is I assume what you'd want this search to return.

rorygraves: @jschlather Fair point - @fommil @aemoncannon  - Are we missing something subtle on the expected behaviour of this search method?

fommil:
Great PR! This will really improve search for everybody :smile: If you could please tidy up that one little method, and do a quick check if we are making the same mistake in the other endpoints, I'll merge and try to get it onto all branches ASAP.

src/test/scala/org/ensime/intg/BasicWorkflow.scala
fommil: if you do a search for `java io` I would expect `java.awt.image.ImageObserver` to come back (`io` matches camel case the `ImageObserver`). If you can add that as an additional search test I'm confident that there will be no regressions in expected behaviour.

jschlather:
@fommil @rorygraves 

ImageObserver - Okay, so there was another issue with publicSymbolSearch counts that I missed that was causing this not to return. In Indexer.scala when oldSearchSymbols prunes any results that are fields. I can fix this in one of two ways:
1. Change add(FieldIndexT, Occur.SHOULD) to Occur.MUST_NOT in the query, presumably then I would rename the function. publicSymbolSearch seems to be the only consumer of this, so I don't think this would break anything. But maybe we want this functionality in general, so I could also make a new function for it.
2. I could do something strange like pull back 2n results and then take n results after pruning.

tieBreakerMultiplier - Unfortunately DisjunctionMaxQuery is a java class so I can't pass a named parameter into its constructor. I can either pass 0f or leave it as is (or I guess wrap it in a scala function...)

Unit Tests - I'm also have a strange issue on my Ubuntu 14.04 desktop that I wasn't having on OS X, wherein the file locations aren't returning for functions. So instead of getting `TypeSearchResult("java.io.File", "File", 'class, Some(_))` I get `TypeSearchResult("java.io.File", "File", 'class, None)` this is happening on master for me. I don't know if anyone else has had this problem.

Other places where this could be happening - I looked around a little in the codebase, but I'm not that familiar with it so I don't really know where I should be looking. When you say other endpoints are you just referring to other places where we're querying Lucene or any rpc call? It looks like we do something similar when deleting documents, but I'm not quite sure.

fommil:
0f is fine 

Re the failing test, do you definitely have to java sources installed? It might be failing to resolve the source file in the zip. Although I thought we died early if that was the case. 

Re method changes, if it changes to behaviour such that the name and docs no longer are true, then change them as well. No point leaving a dead method around... if we want it we can always resurrect it.

fommil:
Other endpoints, I mean the other methods in the same file

jschlather:
Okay, installing the jdk-sources fixed my unit test issues.

I'll change the method then.

Looking at Indexer.scala we do have a similar issue for ImportSuggestionsReq. I think I can only partially fix this one. We're pruning based on endsWith in the oldSearchTypes function. In order to replicate a query with that functionality in Lucene, you have to store the string reversed because it doesn't allow queries beginning with a wildcard. I can at least switch it to do a single query if you'd like as a part of this PR.

rorygraves:
@jschlather Sorry this fell off my radar - I think this looks good @fommil  - any more comments/do you want to kick the tires a bit more before merge?

fommil:
I was waiting for @jschlather to change the method, unless he wants to do it in a separate PR.

fommil:
@jschlather btw, the reason we call these methods "old" is because we were hoping to come up with new ones that have more well defined behaviour. But that involves changing the protocol and that is blocked on my current branch being completed (making it a lot easier to add new request/response types).

jschlather:
@fommil @rorygraves Okay, was a bit busy last week and didn't have time to finish this until today. So I pushed another commit which I think addresses the remaining issues. There's still the similar issue with searching on types, but I think it'd be better to address that in a separate PR.

fommil:
thanks for this @jschlather ! I think this has turned into a change in expected behaviour, so once this turns green we should kick off the emacs integration tests and see if they are happy with this.

Once I've finished with my protocol changes, it should be a lot easier to add new protocol endpoints, so it should be very easy to add in separate searches for types, fields and methods --- at which point we might want to resurrect the older combo behaviour.

fommil:
https://travis-ci.org/ensime/ensime-emacs/builds/50419444

PR #819, opened 2015-02-06
First pass on hosting doc jars from ensime.

aemoncannon:
Note:
* Jars are never expanded to disk. Entries are read out on demand.
Todo
* Might want to parameterize the java.* doc url.


fommil:
This is going to be soooo sweet! :grin: 

awesome, you're using spray can :smile: I really hope that we can move to spray-can websockets post-1.0.

BTW, please try to stick to the parts of spray-can that we can use in scala-2.9, or we will have to spawn off an experimental branch. It should be fine, I don't think you'll be doing anything too unusual.

How come we need commons-io? If something is transitively pulling in Guava, you might want to have a look at it instead.

aemoncannon:
@fommil Thanks
re: plugin. yep. removed accidental lines in later commit
re: commons-io. I just used it out of habit. Can avoid I think.
re: i'll have a look at the spray docs and see about version compatibility

fommil:
@rorygraves just found this http://spray.io/documentation/1.1.2/spray-io/ (but may not be relevant here). I've definitely used spray-can in 2.9 so it'll be API changes more than anything, but might need a custom resolver.

rorygraves:
@aemoncannon Awesome stuff btw!

fommil:
@aemoncannon btw, have a look at pimpathon for util related things. It is more or less replacing guava/commons-io convenience methods for me, by pimping core java classes. Of course, Guava adds additional functionality so still has value.

fommil:
sorry, slippery mouse! Meant to click "comment" :blush:

aemoncannon:
@fommil looks like pimpathon is missing a getExtension helper. we are already pulling in guava, so i'll just use that i suppose

aemoncannon:
An unanticipated difficulty: I want to be able to link directly to an arbitrary scala member, for example, the current code can generate the following:
http://localhost:60531/scala-library-2.11.4-javadoc.jar/index.html#scala.Option@getOrElse[B>:A](default:=>B):B
Which works!  but, on closer inspection, scaladoc is sloppy about its link generation, and doesn't follow consistent rules. For example, sometimes Iterable is fully qualified, and sometimes not. And sometimes a type parameter will be randomly replaced with "Repr" in the link text.

Sigh.

Looks like re-using scaladoc code might be the only reliable way to work this -- but i'm not sure they support the one-thing-on-demand use case. Will look into it further.

fommil:
heh, so you'll have to pull in scaladoc as a dependency and call it to generate the URL name?

If it can be used programmatically, maybe we can use it to dynamically generate project doc pages. It would be awesome to include that as part of method name completion actually... I miss that from IntelliJ Java coding days (although, admittedly scala devs never actually document the methods anyway taking a more "the code is the documentation" approach)

aemoncannon:
Yeah something like that.
I don't have a ton of hope as it (scala.tools.nsc.doc.model.ModelFactory) doesn't look like it was meant to be used incrementally. 

If that fails, plan B is to load the html on the server and use a known prefix to find the anchor tag. The search will be imprecise on overloads, but that's not a huge loss IMO.

fommil:
for scala, the other hack is to populate the search box of the scaladoc browser with the FQN

aemoncannon:
Hah interesting

Btw its just the member suffix that's hard. Can get to top level type HTML
reliably.

fommil:
oh god, I'm pretty sure I've written the Regex Of Doom to do exactly this in Java: https://github.com/paul-hammant/paranamer/blob/master/paranamer/src/java/com/thoughtworks/paranamer/JavadocParanamer.java#L149

aemoncannon:
Hah!

fommil:
I'm surprised that you need to extract the docs onto the filesystem, I thought you'd able to intercept the requested URL and then just load/server that content. Is this just a quick hack to get it up and running?

aemoncannon:
@fommil I was a little concerned about perf for pages with lots of elements. Anyway just tried it and it moves along fine, besides being much cleaner. Nice!

aemoncannon:
@fommil Btw, RE: my previous concerns about building the link urls. It turns out they were mostly unfounded. A little digging revealed scaladoc using Symbol#signatureString to make their links. I do the same now and things are working nicely.

aemoncannon:
Squashed, removing lots of little model changes that are no longer required.

fommil:
@aemoncannon is this in a state where I can use it this coming week and kick the tyres? (I'm working on a 2.11 branch of our codebase at work, instead of the older 2.9 one :smile: I don't expect this to be a permanent situation, though)

fommil:
I noticed you added some binary resources. There is no need to do this because we can use SBT to download any java/scaladoc we want from Maven Central and then pass the file handle to the tests as a resource. Have a look in `build.sbt` to see how we do this for source files, we may as well do the same for all the docs and have access to them as well. A test covering looking up the java/scala stdlib would also be useful.

aemoncannon:
@fommil yes kick the tires! It's getting pretty good IMO. You'll need to
wire up some elisp

aemoncannon:
@fommil cool will look into that jar thing

aemoncannon:
Some elisp:

(defun ensime-rpc-doc-uri-at-point (file point)
  (ensime-eval
   `(swank:doc-uri-at-point ,file ,point)))

(defun ensime-show-doc-for-symbol-at-point ()
  "Browse to documentation for the symbol at current point."
  (interactive)
  (let* ((url (ensime-rpc-doc-uri-at-point buffer-file-name (point))))
    (if url (browse-url (url-encode-url url))
      (message "No documentation found."))))

fommil:
there are a lot of exceptions in my project because my own javadocs are not available. I'd expect just a one line WARN about the missing file instead of a stacktrace, since it's an expected error case.

fommil:
```
16:25:12.049 [default-akka.actor.default-dispatcher-4] ERROR akka.actor.ActorSystemImpl - Uncaught fatal error from thread [default-akka.actor.default-dispatcher-2] shutting down ActorSystem [default]
java.lang.NoClassDefFoundError: spray/can/Http$
	at org.ensime.core.DocServer.preStart(DocServer.scala:75) ~[ensime_2.11.jar:0.9.10-SNAPSHOT]
```

Maybe because it can't find my javadocs?

fommil:
(ignore the failing actor message, I forgot to delete my classpath file and it didn't have the correct classpath for your branch)

fommil:
this is very very cool :smile:

It worked great for some methods I tried but failed for 

```
http://localhost:53419/akka-actor_2.11-2.3.9-javadoc.jar/akka/actor/Scheduler.html#schedule(scala.concurrent.duration.FiniteDuration,%20scala.concurrent.duration.FiniteDuration,%20akka.actor.ActorRef,%20scala.Any)
```

(going to the top of the file instead of the method). It looks like this is to do with the addition of implicit parameters, it should be this:

```
http://localhost:53419/akka-actor_2.11-2.3.9-javadoc.jar/akka/actor/Scheduler.html#schedule(scala.concurrent.duration.FiniteDuration, scala.concurrent.duration.FiniteDuration, akka.actor.ActorRef, java.lang.Object, scala.concurrent.ExecutionContext, akka.actor.ActorRef)
```

aemoncannon:
oh i bet i know why. akka's docs are javadocs right? we're probably failing
to convert scala.Any to java world

aemoncannon:
@fommil these are great for the test suite, so keep them coming as you find
them.

fommil:
@aemoncannon btw are scaladocs really just javadocs under the hood? I was surprised with how plain they looked.

aemoncannon:
@fommil at what level? it seems they tried to keep the basic jar structure
as close as possible, but there are differences in the link formats. i bet
there's a   way to get scaladocs for akka, but I don't know.

fommil:
@aemoncannon I mean when ENSIME serves them it doesn't have all the gloss that scaladocs typically have.

aemoncannon:
@fommil Screenshot? That's news to me

fommil:
I'm at my home computer now (I set this up at work), but this is what I mean by the "gloss" of scaladocs http://www.scala-lang.org/api/current/?_ga=1.17448380.489730132.1397315774#scala.Some

whereas when I browsed to the docs in ENSIME it was just the plain content without any stylesheets and without the package/types list, search bar etc (which I believe is in a different frame)

aemoncannon:
Ok. something wonky going on - I see full gloss on my local machine. What
browser are you using?

fommil:
@aemoncannon chromium on ubuntu. I'll try again tomorrow... I'm in lisp/shell/make scripts tonight to finally build these emacsens.

aemoncannon:
@fommil 
Your scheduler(...) uri should work now. Also added a bit more request logging, so hopefully we can see why the css is not loading.

fommil:
sweet.

Sorry, my branch has resulted in this breaking. Do you need a hand with how the protocol works now?

I think you're just adding one new wire format so your implementation just needs to be the same as the others (this will go away at some point) and then your case class formatter should be optional but you can explicitly create one for a (probably negligible) performance boost.

e.g.

```scala
implicit val docMessageFormat = SexpFormat[DocMessage] // for a case class called DocMessage
```

fommil:
@aemoncannon I've sent you a PR with all the correct merge choices. I'm continuing the tradition of always running the bleeding edge :smile:

fommil:
Keybinding suggestion: `C-c C-v d`

fommil:
yay! the scheduler is indeed working now.

Also, I was obviously talking total rubbish about the scaladoc gloss.... akka must just publish javadocs even for the scala API.

aemoncannon:
Question: how do we decide which java version to use for documentation purposes? The local java runtime? What's the best way to grab the version?

aemoncannon:
Related question: right now I call out to the web for core javadocs. Should we instead suck in a copy via sbt(does that exist)?

fommil:
@aemoncannon I see what you mean. It seems that the javadocs are not distributed with the JDK, although openjdk on debian has an `openjdk-7-doc` that will extract the html pages under a `doc/api` folder. For Oracle's distributions, it can be downloaded, but a license must be ticked. e.g. http://www.oracle.com/technetwork/java/javase/documentation/jdk8-doc-downloads-2133158.html I don't know what the story is for OS X.

Probably the cleanest way to do it is to have a top level entry in the config file (like the sources) for the docs, and to use it in the server if it is there... otherwise point at the web. That keeps the server logic nice and simple (although this one can be a zip, jar or a directory - luckily VFS handles that well).

Then in the sbt plugin we could make it a user-definable function to return the doc zip, jar or root directory given a JDK directory. If we document it right, it should be ok and many people might consider local JDK docs to be a niche requirement so we can maybe give it a page of its own instead of cluttering the quick start guide.

fommil:
btw, I just realised I had two commits sitting on my machine that I didn't push. They basically just skipped jars that didn't exist (no exception noise) and a faster way to check if something was java/scaladoc that didn't read the whole file. I'll push them tomorrow and send you another PR.

fommil:
@aemoncannon regarding java version, remember we always run under the same JDK as the project :wink: so you can just do a `scala.util.Properties.javaVersion`

aemoncannon:
Oh, awesome. Javadoc 8 has decided to arbitrarily change their link format. : \
http://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#max-int-int-

fommil:
uugh, that probably means my paranamer stuff is broken. You think their Javadoc conventions changing is bad? Try keeping up with the Garbage Collector logs: https://github.com/fommil/lions-share/blob/master/analysis/src/main/scala/com/github/fommil/lion/gc/parser.scala

aemoncannon:
This just gets stupider and stupider #copyOf(byte[], int) becomes #copyOf-byte:A-int-
whhhhhhyyyyyy

aemoncannon:
At least we're not alone http://git.eclipse.org/c/jdt/eclipse.jdt.core.git/commit/?id=cd169c7695d941d4918a15490c18d5c9ea399e1c

aemoncannon:
Incidentally, is all of eclipse written in this c-in-java style? Must be a performance sensitive call...

aemoncannon:
Here's the original javadoc bug from 2003 (predates generics): https://bugs.openjdk.java.net/browse/JDK-4663254

I guess valid url syntax is a worthy goal.. but still annoying.

aemoncannon:
@fommil @rorygraves @edani I'm probably going to squash and merge in the next day or so. Last looks welcome!

aemoncannon:
Haha. Here's a joyous thing:
scaladoc uses @usecase comment annotations to substitute kid-safe signatures in the docs. http://stackoverflow.com/questions/26132459/why-are-scaladoc-method-signatures-wrong

Happily, these usecase signatures are also used for the link anchors. As a result, it's impossible to generate the links without analyzing the doc comments : (

Grepping through scala-library I find 75 instances of @usecase annotations. My current plan is to just bake in a hand-built hash: real-to-fake. I can't think of a better way : (

aemoncannon:
https://issues.scala-lang.org/browse/SI-9168

fommil:
uugh. Is the annotation available in the classfile at least? We could index this annotation (although that feels like too much work to workaround this)

aemoncannon:
think it's docs only. well, it's actually only 65 instances. and they're
all in scala.collections

edani:
@aemoncannon   This looks very good from my side! Regarding the "usecase" signatures, a workaround may be to scan the document for links and return the first match that contains the method. Even without that your PR is super useful!

fommil:
I still forgot to send you the small fixes I had on my machine, doh! It's irrelevant anyway.

aemoncannon:
Little status update:
Still plugging away. I'm moving to a more comprehensive testing strategy, where the test will actually look at the html content and verify that the anchor exists as computed. The @usecase thing made me realize there are many ways that future doc formats could/will break our stuff, and I want some alarms to go off when that happens.

aemoncannon:
@edani Btw, I pretty much used your plan for @usecase handling. I included a whitelist of known usecases though, and limited to the scala library, as this approach could cause problems with overloads if used in general.

My hope is that either A) Use of @usecase doesn't spread too much, or B) scaladoc is fixed so that the anchor's are not derived from the usecase signature.

aemoncannon:
Squashed.

fommil:
actually, this PR is too useful for me to be nitpicky about the testing stuff... I'll merge it in my branch. Thank you! :smile:

aemoncannon:
woo!

rorygraves:
Awesome.

PR #841, opened 2015-02-24
Modularise

fommil:
I'm happy for this to be merged as-is, but if the review takes a while I might start adding more commits to the end. The only thing left here is to move all the swank stuff into its separate module, but I can do that in a smaller PR if this is looking good to people.

TL;DR

1. split project into api, sexpress, swank, testing, server
2. redesign how example projects are used in tests, and move all such tests to the `IntegrationTest` config
3. create a new suite of scalatest `Fixture`s to be used in integration testing: effectively playing the role that Spring Testing would provide in Java.

With the new integration testing framework, it means we can effectively spawn up an integration test against any arbitrary sbt/git project and write automated tests around it.

I also explicitly set the S-Express license to be LGPL... if anybody has a problem with that I can be convinced to go back to Apache. It's completely compatible to do that because it's now going to be distributed as a separate jar on the classpath, plus I think it's more in keeping with the Emacs philosophy (our front-end is GPL).


testing/debug/src/main/scala/stepping/ForComprehensionListObject.scala
```diff
@@ -4,7 +4,7 @@ object ForComprehensionListObject {
 
   def main(args: Array[String]) {
 
-    val l= List(new Object(), "deux", "quatre", "huit")
```
aemoncannon: Might want to exempt these test sources from auto-formatting?

aemoncannon:
lgtm overall. i love the test fixtures

testing/debug/src/main/scala/stepping/ForComprehensionListObject.scala
fommil: aah, for stepping in multi parameter lists. good point. I'm not sure how to do that.

fommil:
interestingly, this branch is running faster on travis already than master, which is good. I've not even sped up the compile yet by moving all the implicit stuff into `swank`! (believe me, you'll all be pleased when I do that... the compile time is getting crazy)

rorygraves:
cool

fommil:
I rebased at lunchtime today, but the tests are failing now... I'm probably doing something wrong in the new version of the tests, probably the sourceAtPos stuff.

@aemoncannon once this test goes in, I think it might be worthwhile to refactor the DocServer tests so that the testing focus is on the lookup from FQNs as that seems to be the atomic thing to test.

aemoncannon:
@fommil You mean, as opposed to pulling symbols from sources? That would leave one tiny corner uncovered (we actually undo implicit conversions, e.g. we go to Option#map instead of TraversableOnce#map), but otherwise would be fine. Maybe could break it apart into two or something.

fommil:
@aemoncannon yes exactly. We could add a few lookups to the BasicWorkFlow test to cover that layer.

I'll try to get my tests passing then merge very soon so that it reduces the risk of merge hell for people.

fommil:
@aemoncannon actually, if we're using the `IsolatedPresentationCompilerFixture` you could even do a few in-place tests in the same file as isolated tests that only need an `EnsimeConfig`.

fommil:
yeah, he is a damn jerk...

fommil:
This is green, but the DebugTests are timing out in their coverage run so the coverage looks low.

Hmm, I'll see about turning off parallelisation or something.

fommil:
ok, we're green on local emacs tests as well.

I recommend merging ASAP to avoid nightmare merge situations

aemoncannon:
sgtm!

fommil:
gimme a few runs to disable parallel testing on travis then we should be good to go.

btw, emacs23 is no longer compatible. We should just bump the requirements to emacs24.

fommil:
apparently not running in parallel has nothing to do with it.

fommil:
screw you, coverage jerk!

PR #853, opened 2015-02-28
build stability on Travis

fommil:
also, rely on coveralls to tell us if the coverage failed or not (the thresholds are now options)

### DIFF
--- a/project/EnsimeBuild.scala
+++ b/project/EnsimeBuild.scala
@@ -23,8 +23,11 @@ object EnsimeBuild extends Build with JdkResolver {
     scalaVersion := "2.11.5",
     version := "0.9.10-SNAPSHOT"
   )
-  val isTravis = sys.env.get("TRAVIS").isDefined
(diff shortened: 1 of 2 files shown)


fommil:
@rorygraves I've tried lots of different combinations, but for some reason the `DebugTest` keeps timing out when travis runs the coverage. No idea why, maybe a memory limit thing?

rorygraves:
Yeah, I don't think they are stupidly big instances, ensime and debug instance probably push the limits.
will double check by beating the jenkins instance around so they run there properly and see what happens.

fommil:
@rorygraves probably just running locally with different memory settings is a fine tests, but I'm not even sure that's the problem.

rorygraves:
I more mean the entire VM  rather than the processes  - will ahve a play.

fommil:
I disabled the jenkins hooks because they were red lighting on everything.

rorygraves:
I believe that was because the scoverage output location moved - will fix.

fommil:
the multi-module changed the script that needs to be run

PR #857, opened 2015-03-01
travis tests serially

fommil:


fommil:
this (and other PR) failed due to hanging emacs debug test.

fommil:
I tested a version of ensime-server with the

```
Debugger VM args:  -classpath "/tmp/ensime_test_proj_20585Kmu/target/scala-2.11/classes:/tmp/ensime_test_proj_20585Kmu/test-target"
```

And I'm now amazed that this ever worked :smile: looks like a genuine bug (possibly in the emacs test) because the `.ensime` doesn't have the scala-library on its dependency list.

```elisp
(:root-dir "/tmp/ensime_test_proj_20585Kmu/" :cache-dir "/tmp/ensime_test_proj_20585Kmu/cache/" :name "test" :scala-version "2.11.5" :java-home "/usr/lib/jvm/jdk-6-oracle-x64" :subprojects ((:name "test" :module-name "test" :source-roots ("/tmp/ensime_test_proj_20585Kmu/src/main/scala/") :depends-on-modules nil :target "/tmp/ensime_test_proj_20585Kmu/target/scala-2.11/classes/" :test-target "/tmp/ensime_test_proj_20585Kmu/test-target/")))
```

fommil:
btw, this is probably the same thing in https://github.com/ensime/ensime-emacs/pull/97

fommil:
the latest commit will fail until we merge my emacs branch

PR #860, opened 2015-03-02
Random code cleanups

rorygraves:
There is a single functional change hidden within a large number of random code cleanups.
The functional change is to create a future for application start so that an app using Ensime as a library can await the server being ready.

Other than that its just random cleaning up of warnings, imports etc.


fommil:
oh bugger, this will fail. I'll try to fix the build tonight.

fommil:
looks like we're really stressing out the sbt-coveralls plugin.

aemoncannon:
@rorygraves lgtm. do you have a tool to do all this cleanup? or are you just very thorough : )

rorygraves:
@aemoncannon My dirty secret is that I still use IntelliJ.  When opening a file it highlights stuff like this, it also has an 'inspection' mode and you can apply the fixes automatically.  So I occasionally run the full suite and do a bit of a purge (its very low brain input, so I do it when I can't find enough energy to do something more meaningful).

aemoncannon:
@rorygraves infidel! ;-)  these are great changes.

PR #891, opened 2015-03-14
Travis to Shippable

fommil:
This is to trigger the Shippable builders. I'll merge when it's all working.


fommil:
yay! Green lights with Shippable.

aemoncannon:
Awesome. 21mins

fommil:
I think the emacs tests are the slow bit now because we no longer need to throttle the scala tests for Travis crapness. Do you think we could do a few of the emacs ones in parallel?

aemoncannon:
Maybe just run a few emacs binaries in parallel? The test harness assumes
serial execution i believe. Would be nice if we could get per instance
logs...

fommil:
also, eric's idea to share the ensime cache will give us a big speed up.

fommil:
that last successful build was using our docker image without any caching :smile:

btw, if you want to try out our docker image, install https://docs.docker.com/installation and then type

```
docker run -i -t ensime/ensime
```

(might need `sudo` if you haven't set up user perms). Emacsen are in `/usr/lib/emacs-24.[1234]` and we have various java, paulp's `sbt` and a basic populated ivy cache in `/root/.ivy2` / `/root/.sbt`

fommil:
ok, before I merge this I want to create a tagged "1.0" image and then have the "latest" image rebuild when there are any changes to `ensime-server` and all I'll do is run `updateClassifiers` and publish a new image (this is really easy to do with docker).

That way we'll have a super optimal ivy cache for any incoming PRs!

fommil:
ok, this is looking good! One thing left to do: I have to get my build node working, otherwise everything will fail 50% of the time. I'll do that, and get `ensime-emacs` working, then merge.

rorygraves:
Awesome stuff!

PR #895, opened 2015-03-18
Typed debug thread ids + logic seperation

rorygraves:
Two changes for the price of one:

1. Move from ints encoded as strings for debug threadIds (its been a bugbear of mine for a while).
2. Finish the work to have a startup future (needed for the ensime as API changes.

Part of (2) included moving the message backlog and replay logic out of the Project class into the protocol layer for clarity.

Also moved rpcPeekUndo() to be option rather then Either where the error case was a protocol string.

### DIFF
--- a/api/src/main/scala/org/ensime/model/Model.scala
+++ b/api/src/main/scala/org/ensime/model/Model.scala
@@ -2,9 +2,6 @@ package org.ensime.model
 
 import java.io.File
 import org.ensime.util.FileEdit
-import scala.collection.mutable
-
-import org.ensime.config._
 
 sealed trait EntityInfo {
   def name: String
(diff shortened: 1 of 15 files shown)


server/src/it/scala/org/ensime/fixture/IsolatedTestKitFixture.scala
```diff
@@ -0,0 +1,12 @@
+package org.ensime.fixture
+
+trait IsolatedTestKitFixture extends TestKitFixture {
```
rorygraves: This was moved because its initial location was causing weird compiler issues.

rorygraves:
@fommil Shippable failed with some weird error

fommil:
oh shippable won't work until my branch is merged

server/src/main/scala/org/ensime/server/Server.scala
```diff
@@ -44,19 +45,25 @@ object Server {
...
   }
 
-  def initialiseServer(config: EnsimeConfig): Server = {
+  /**
+   * Initialise a server based on the given ensime config.
+   * @param config The ensime config
+   * @return A tuple of Server instance and initialisation future.
+   * @see Project.start()
+   */
+  def initialiseServer(config: EnsimeConfig): (Server, Future[Unit]) = {
```
aemoncannon: Preferring British english here I see : )

aemoncannon:
lgtm

server/src/main/scala/org/ensime/server/Server.scala
rorygraves: Of course!  Or as we call it - the correct spelling ;)

rorygraves:
@fommil  Please review the third commit around object ids.
@aemoncannon The last commit on this should mean all object ids are strings on the emacs side avoiding the error.

aemoncannon:
awesome!

rorygraves:
@fommil Shippable seems to be timing out with this PR?

@fommil  @aemoncannon  it would be good if you could give this branch a kick around.

fommil:
that's twice the startup of the server hangs in the emacs integration tests. can you check locally (after rebasing) that all is well?

rorygraves:
@fommil Wanna take a quick look and merge.  Looks like my other commit (the init changes) interacted badly with the emacs test - will take another look once this has been pulled.

fommil:
:smile:

PR #896, opened 2015-03-19
Add symbol-by-name call. 

aemoncannon:
Subsumes member-by-name call. 
This call and doc-uri-for-symbol, now using the symbol signatureString to disambiguate overloaded methods.

Why?
* Gives us an unambiguous way to refer to symbols which we can immediately use in the type inspector to fix jumping to source for method overloads.
* Reduces reliance on type ids.


rorygraves:
lgtm - a couple of minor comments.

fommil:
can you please rebase and squash? That should let shippable kick off a build

rorygraves:
@aemoncannon Does this need a matching emacs change at the same time -if not go ahead and merge :)

PR #901, opened 2015-03-21
shippable updates

fommil:
and hopefully this will be using the new "latest" docker image, which has pre-cached ivy jars. w00t!


fommil:
yay, caching worked. That should shave 3 mins off the build time. All we need now is for the emacs tests to reuse the lucene/h2 indexes, and for sbt to fix the caching once and for all https://github.com/sbt/sbt/issues/1930

PR #927, opened 2015-04-08
Startup future and message caching changes

rorygraves:
This is a resubmission of the earlier work - hopefully I have addressed the emacs killing problems.
Do not merge until this has actually been given a full workout in Emacs.


rorygraves:
Hmm, seems to hang the emacs tests - more investigation needed.

Issue #390, opened 2014-06-19
refactor swank parser layer


fommil:
(possibly parboiled, possibly with Spray JSON or something in the middle)

minor breaking protocol changes may be tolerable for the sake of consistency (but releases need to be well timed with the backports branches and the emacs client)

fommil:
Just a quick update on this, I think I'm ready to put in the first PR as the tests are green.

1. ~~rewrite SwankProtocolConversions to use S-Express~~
2. rewrite SwankProtocol to use S-Express (kill off old SExp/WireFormat code)
3. change the protocol to use the "natural" S-Express transformations (lots of emacs work)
4. use Shapeless-powered 

once that's all done, the entire Swank protocol layer will probably collapse to a couple of lines AND the wire format should be easily swappable with another implementation: e.g. JSON or XML.

fommil:
@rorygraves had a good idea: release 1.0 sooner without any protocol changes at this point. I'm in favour of that, as it means we can change the protocol on the unstable branch, impacting less people, and not have to worry about backporting to scala 2.9.

rorygraves:
More that if as part of 1.0 we were going to make a number of tweaks ot the protocol it might be worth release 0.9.1 stable (now) - 0.9.2 stable (+protocol tweaks) ... (1.0.0 stable) once we were 1.0 feature complete..

fommil:
back on this after a hiatus into the land of DevOps...

fommil:
I'm working on converting all the incoming messages to use S-Express

fommil:
I'm also trying to generalise the marshalling so that we can plug in a JSON marshaller

fommil:
WIP available at https://github.com/fommil/ensime-server/tree/sexp_input I feel like I'm making progress here. Swappable protocols is very nearly in sight!

fommil:
@rorygraves one thing that really comes to mind here is that we seem to have two APIs: one is async and message based (what `SwankProtocol` is becoming) and the other is synchronous (`EnsimeApi`).

I'd really like us to just have **one** API --- the `Protocol` one --- but that will mean that programmatic callers will either have to be actor based or use the ask pattern when calling into ENSIME. I don't think it's really our place to add a trivial ask pattern layer in the off-change that someone wants to use the API that way, do you have a strong opinion one way or the other?

rorygraves:
[...]

PR #928, opened 2015-04-09
SWANK refactor progress toward #390

fommil:
This PR might seem a bit weird, but it's working towards a noble goal.

It shouldn't change the protocol in any way, but to simplify things I removed some of the deprecated endpoints.

The current protocol handling code is layered something like this:

1. raw socket => String
2. String => SExp (old implementation)
3. Sexp => raw parameters
4. raw parameters => synchronous methods
5. ... S-Exp primadonna actors ENSIME ...
6. domain model => S-Express
7. Sexp => String
8. String => socket

but we'd like to get to

1. socket => WireFormat (JSON / XML / S-Exp)
2. WireFormat => domain model
3. ... protocol independent reactive ENSIME ...
4. domain model => Sexp
5. Sexp => socket

and unfortunately this PR seems to muddy the water by bringing us to

1. raw socket => Sexp
2. Sexp => domain model
3. model => synchronous methods
4. ... S-Exp primadonna actors ENSIME ...
5. domain model => Sexp
6. Sexp => socket

but at least we're converging on the desired architecture :-)

Lots of String => File and CanonFile => File as a result of canonisation now occurring at the protocol level. Ideally .canon would only be used when information is entering the system. I might have broken things for OS X.

Also, the `SwankProtocolConversions.scala` file is really becoming a monstrosity (sorry!). It exists entirely to implement the current SWANK protocol and this simplify reflects that we marshall every object in a slightly different way. Hopefully, once we get 1.0 out the door, we can look at using the "natural S-Exp form" which will involve a lot of changes on the Emacs side (nothing functional, just data structures), and this file will vanish thanks to shapeless.


rorygraves:
Awesome stuff.

rorygraves:
Seems to work find on Mac - @aemoncannon  @edani  Any comments before I merge this?

fommil:
don't forget to read `SwankProtocolConversions.scala` via a mirror http://en.wikipedia.org/wiki/Medusa

rorygraves:
It also makes me sad the Ensime dislikes one of its own files - too many implicits....

fommil:
dag namnit

fommil:
crap, that was an accidental push. I'll revert.

rorygraves:
Ok, pulling the trigger

PR #939, opened 2015-04-16
Improve the source location resolution for type requests.

rorygraves:


server/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -72,10 +72,10 @@ class BasicWorkflow extends WordSpec with Matchers
       // public symbol search - java.io.File
 
       val javaSearchSymbol = project.rpcPublicSymbolSearch(List("java", "io", "File"), 30)
-      assert(javaSearchSymbol.syms.exists(sr => sr match {
```
rorygraves: This is just a cleanup

server/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -94,8 +94,7 @@ class BasicWorkflow extends WordSpec with Matchers
       val typeByIdOpt: Option[TypeInfo] = project.rpcTypeById(intTypeId)
       val intTypeInspectInfo = typeByIdOpt match {
         case Some(ti @ BasicTypeInfo("Int", `intTypeId`, 'class, "scala.Int", List(), List(), Some(_), None)) =>
-          // TODO in inspectType, pos is Some(EmptySourcePosition()) hack to make flow work
-          ti.copy(pos = Some(EmptySourcePosition()))
+          ti
```
rorygraves: No longer need to suppress the source info

server/src/main/scala/org/ensime/model/ModelBuilders.scala
```diff
@@ -39,29 +39,33 @@ trait ModelBuilders { self: RichPresentationCompiler =>
...
   }
 
-  def locateSymbolPos(sym: Symbol, needPos: PosNeeded): Option[SourcePosition] =
-    if (sym == NoSymbol || needPos == PosNeededNo) None
+  def locateSymbolPos(sym: Symbol, needPos: PosNeeded): Option[SourcePosition] = {
+    if (sym == NoSymbol || needPos == PosNeededNo)
+      None
     else if (sym.pos != NoPosition) {
-      if (needPos eq PosNeededYes)
+      if (needPos == PosNeededYes || needPos == PosNeededAvail) {
```
rorygraves: Here we have the source position and yet we would only return it if it was 'Yes' rather than 'Yes' and 'if available'

server/src/main/scala/org/ensime/model/ModelBuilders.scala
```diff
@@ -191,8 +195,8 @@ trait ModelBuilders { self: RichPresentationCompiler =>
       }
       def basicTypeInfo(tpe: Type): BasicTypeInfo = {
         val typeSym = tpe.typeSymbol
-        val symPos = locateSymbolPos(
-          if (typeSym.isModuleClass) typeSym.sourceModule else typeSym, needPos)
+        val symbolToLocate = if (typeSym.isModuleClass) typeSym.sourceModule else typeSym
```
rorygraves: Just factored out the pos into two lines to make it easier to read.

server/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -164,14 +163,14 @@ class BasicWorkflow extends WordSpec with Matchers
...
-          BasicTypeInfo("package$", _: Int, 'object, "org.example.package$", List(), List(), Some(EmptySourcePosition()), None),
-          BasicTypeInfo("package$", _: Int, 'object, "org.example.package$", List(), List(), Some(EmptySourcePosition()), None)))) =>
+          BasicTypeInfo("Bloo", _: Int, 'class, "org.example.Bloo", List(), List(), Some(_), None),
+          BasicTypeInfo("Bloo$", _: Int, 'object, "org.example.Bloo$", List(), List(), Some(_), None),
+          BasicTypeInfo("CaseClassWithCamelCaseName", _: Int, 'class, "org.example.CaseClassWithCamelCaseName", List(), List(), Some(_), None),
+          BasicTypeInfo("CaseClassWithCamelCaseName$", _: Int, 'object, "org.example.CaseClassWithCamelCaseName$", List(), List(), Some(_), None),
+          BasicTypeInfo("Foo", _: Int, 'class, "org.example.Foo", List(), List(), None, None),
+          BasicTypeInfo("Foo$", _: Int, 'object, "org.example.Foo$", List(), List(), Some(_), None),
+          BasicTypeInfo("package$", _: Int, 'object, "org.example.package$", List(), List(), None, None),
+          BasicTypeInfo("package$", _: Int, 'object, "org.example.package$", List(), List(), None, None)))) =>
```
rorygraves: Interesting (as weith the old code - two "package$" items - they have different type ids.

server/src/main/scala/org/ensime/model/ModelBuilders.scala
edani: Would you mind doing a performance check for this? Sending all known symbol positions can produce huge messages. Good examples are inspecting  the RichPresentationCompiler class or the org.ensime package.

rorygraves: Hmm, good point - I had considered the calculation performance cost, but not the message sizes/performance impact there. I will have a play.

rorygraves:
@fommil Any idea why shippable didn't trigger for this?

fommil:
I tried this out on `RichPresentationCompiler` and it wasn't noticably any slower than before... I think the benefits of having the sources definitely outweight the potential for it to be a second slower on really big classes.

PR #940, opened 2015-04-17
Upgrades

fommil:


fommil:
rebased

fommil:
ping

edani:
I can get to this 2 or 3 days from now

fommil:
fwiw I've been using this at my work with no problems for a few days

rorygraves:
Been good for me too.

edani:
I'm seeing problems with the test suite:
- on Windows I get several test failures. I'll check if any of them indicate problems with the server
- on Linux the "Breakpoints" integration test hangs. This is a regression from the master branch. I'm using OpenJDK 7, I'll see if the problem persists with Oracle's JDK.

fommil:
Don't forget to rerun gen-ensime for integration tests

edani:
I did run gen-ensime  ande deleted .ensime_cache.  I get the same results with OpenJDK 7 and Oracle's JDK 1.7.

Here is the smoking gun: in `server/it-output.log` I see messages like
11:46:36.579 WARN   org.ensime.indexer.SearchService - failed to insert List(FqnSymbol(None, file:///disk/home/eric/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.11.4.jar, <.......>
) class org.h2.jdbc.JdbcSQLException: The database has been closed [90098-187]
```
This causes the indexer to fail, and the test times out waiting for an "indexer ready" message.

This happens with most integration tests but not 100% of the time which makes me suspect a race condition.  Also the indexer seems to work fine when I start ensime from emacs.

**Comment by fommil** [comment]
huh, weird. The DB would only be closed if it wasn't being used in that test anymore (in which case these sorts of errors can be safely ignored)... but if the DB is being closed **before** the test finishes that sounds like we've had a longstanding bug.

**Comment by edani** [comment]
Are you able to reproduce the problem? If not I'll see which commit introduced it.

**Comment by fommil** [comment]
Nope, works for me. I have a windows image I can run it on, I'll look into it in the next week or so. It'll be good for me have have it as part of my dev process anyway, must be frustrating that I keep breaking all the tests for you :-)

**Comment by edani** [comment]
The windows test problems don't look like a blocker, I'm able to run ENSIME manually without problems. I can't see how this PR could have introduced them  so they must have been there before. I'll just do a git bisect and fix them in a separate PR at some point.

I'd like to spend one more day on the linux test problem though. What JDK version do you use?

**Comment by fommil** [comment]
JDK 1.6 from SUN (intentionally, because this is still the official Scala JDK, believe it or not, and the official JDK for one of my work projects). I'm hoping we can make 1.7 the minimum requirement post 1.0 because that means we can use the fancy native filesystem handling hooks.

**Comment by fommil** [comment]
btw, turns out I won't be able to look at running in Windows for a few days... because ironically I'm getting **real** windows put in this week and the workers have put sheets over everything (including my desktop with the Windows images).

**Comment by rorygraves** [comment]
ok, something interesting going on with the indexing - was seeing failures on this branch (on Macbook).
Investigating it was timing out on the indexing - increasing them gave me this:
```
21:46:31.152 DEBUG  org.ensime.indexer.SearchService - indexing file:///Users/rorygraves/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.11.6.jar
...
21:52:00.920 DEBUG  org.ensime.indexer.SearchService - committing index to disk...
21:52:04.109 DEBUG  org.ensime.indexer.SearchService - ...done committing index
```
took 5.5 minutes (much longer than the default 3 minutes timeout current timeout).

edani:
Hrmm... it turns out that my /tmp partition was filling up during "it:test".  It must have been just large enough for the tests in the master branch.  So this can be merged as far as I'm concerned (maybe increase the timeout as @rorygraves suggested).

rorygraves:
Got to the bottom of this - turns out the analyser was not correct discarding the presentation compiler when ensime was shutdown.

@fommil  has just merged a change I added on top of his branch to fix.   Assuming it passs the test good to merge.

rorygraves:
It was basically struggling with memory which is why the test took so long - the bug has been in there a while.

aemoncannon:
nice catch

rorygraves:
There is a related issue in the SearchService which I am looking at.  PR coming.

PR #951, opened 2015-04-29
Fix escaping of files in Swank protocol tests

edani:
This fixes a few Windows test failures. The thing to remember is to always use `fileToWireString` when hard-coding protocol messages  (canonizes and escapes backslashes)


fommil:
I got a transient when I merged this to master: it was scalacheck finding a bug in sexpress. I'm going to be looking at that as part of 1.0.

rorygraves:
The good and the bad of randomised testing in one sentence.

fommil:
I think I have a better solution to the roundtripping of `File` in the tests. It should be pretty simple to use shapeless to go through and canonise all the `File`s in the endpoints, instead of tying this into the marshalling layer.

PR #952, opened 2015-04-30
convert 'declaredAs' symbols to case objects

esessoms:
Fix for issue #900 low-hanging fruit.


aemoncannon:
Radical

rorygraves:
@esessoms Awesome - this has been on my bugbear list for a while and I hadn't got to it ;)
Unfortunately we can't merge this yet as the the integration tests fail, I'm guessing you can `sbt clean test`, its probably underdocumented that we have a suite of integration tetss as well.  In this case  the ```BasicWorkflow``` integration test fails to compile as it is still trying to match against the symbols - 

```
[error] /root/workspace/src/github.com/ensime/ensime-server/server/src/it/scala/org/ensime/intg/BasicWorkflow.scala:49: fruitless type test: a value of type org.ensime.util.DeclaredAs cannot also be a Symbol
[error]         case Some(SymbolInfo("scala.Int", "Int", Some(_), BasicTypeInfo("Int", typeId, 'class, "scala.Int", List(), List(), _, None), false, Some(ownerTypeId))) =>
```

I normally run this:
```
sbt clean
sbt gen-ensime
sbt test it:test
```

@fommil Might also have some comments around ```RefactorLocation``` - I don't think we can fully roundtrip the messaging but that is definitely the goal.

fommil:
Cool. Yeah, some work to be done on the protocol here to keep it consistent with the client. 

Btw, did you shapeless allows sealed traits with case objects to be aggregated? I was thinking that would be a nice thing for these enum type things.

fommil:
FYI http://stackoverflow.com/questions/29676204/

esessoms:
Ah, yes, I did not know about the integration tests.  Will fix.  Thanks!

The comments in the header of SwankProtocolConversions.scala suggest avoiding shapeless.  Out of date?

rorygraves:
@esessoms  Cool.
I will have to leave it to @fommil  to comment on the shapeless comment.

fommil:
@esessoms yeah, I'll just do manual marshalling where we can't use shapeless. The entire JSON protocol (my branch) is pretty much automatically derived using shapeless 2.2 (out soon), I've been thinking about simplifying the SWANK protocol implementation with exactly the same approach. This file would be a **lot** cleaner and make it more obvious which steps need to be taken to simplify the wire protocol itself.

api/src/main/scala/org/ensime/util/RefactorLocation.scala
```diff
@@ -0,0 +1,11 @@
+package org.ensime.util
+
+object RefactorLocation {
```
fommil: any chance of doing a similar case object hierarchy here as well?

fommil:
great stuff! One small comment above, and also can you please rebase and squash your commits?

esessoms:
The squashed commit includes changes for RefactorLocation as well.  I'm not entirely happy with those particular changes and may need to revisit them, but at least everything is consistent now.

fommil:
Awesome!

fommil:
Also, it was very brave of you to go anywhere near the protocol file. I'd hoped to simply before anybody else had to look at it

esessoms:
LOL.  Well, the whole point was to start to learn my way around, and the protocols were actually very useful for that, so it's all good.  :)

PR #956, opened 2015-05-03
Various updates to improve test memory behaviour

rorygraves:
Tests were running out of memory - various changes to ensure that Ensime instances
clean up after themselves.  Signficantly speed up tests.

SearchService created a thread which was never cleaned - move to actor.
SearchService created its own threadpool - now use ActorSystem managed dispatcher.
Shutdown H2 when EnsimeServer is shutdown.
Shutdown the filewatchers on shutdown.
Improve the sharing of actor systems in tests.

Saves 2:20 on my laptop.
master - sbt clean test it:test  791.06s user 21.10s system 271% cpu 4:58.73 total
branch - sbt clean test it:test  595.71s user 17.52s system 384% cpu 2:39.41 total


rorygraves:
We are still leaking memory horribly (vfs I'm looking at you!)  But this makes it significantly better and should make the testing more stable as well.

rorygraves:
One for your review @fommil

rorygraves:
Updated PR to use a managed vfs rather than a global shared one (which was leaking memory)

rorygraves:
@fommil  - Back to you.

fommil:
sweet!

Issue #919, opened 2015-03-31
swank:debug-stop Does not actually stop the VM


rorygraves:
From Google groups message.

swank:debug-stop disposes the vm mirror (calls ```vm.dispose()```) it should probably call ```vm.exit(0)``` otherwise we are leaking processes.

See jdi documentation at https://docs.oracle.com/javase/7/docs/jdk/api/jpda/jdi/com/sun/jdi/VirtualMachine.html 

This needs an integration test and a fix.

fommil:
I guess it's unlikely that somebody would want to keep the JVM running after stopping a debugging session. Would this explain why our debug tests used to grind to a halt?

rorygraves:
Yep - that would be my guess.  It may well explain that test failure.

PR #957, opened 2015-05-05
Issue #919 - Stop debug vm on end of debug session.

rorygraves:
Debug sessions launched by Ensime were not terminated when Ensime finishes debugging.
This commit makes the debugged vm stop when we terminate the debug vm session.
This may be a little agressive for attached sessions.

### DIFF
--- a/core/src/main/scala/org/ensime/core/DebugManager.scala
+++ b/core/src/main/scala/org/ensime/core/DebugManager.scala
@@ -19,8 +19,9 @@ object DebugManager {
   )(
     implicit
     config: EnsimeConfig
-  ): Props = Props(classOf[DebugManager], broadcaster, config)
+  ): Props = Props(new DebugManager(broadcaster, config))
 }
+
 class DebugManager(
     broadcaster: ActorRef,
     config: EnsimeConfig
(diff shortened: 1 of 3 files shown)


aemoncannon:
Cool

edani:
@rorygraves  I think for attached sessions, exit() shouldn't be called, just dispose().

rorygraves:
@edani Yeah fair comment - will look at it when i get the chance.

rorygraves:
lol, went to fix this over the weekend and found several other bugs - so bigger refactor in progress - closing for now.

core/src/main/scala/org/ensime/core/DebugManager.scala
```diff
@@ -237,67 +239,75 @@ class DebugManager(
...
-      handleRPCWithVM() { vm =>
-        sender ! TrueResponse
+      sender ! handleRPCWithVM() { vm =>
+        TrueResponse
       }
-
     case DebugStopReq =>
-      handleRPCWithVM() { vm =>
+      sender ! handleRPCWithVM() { vm =>
+        if (vm.mode.shouldExit) {
```
rorygraves: @edani This fix means it should only trigger the exit on a VM started by Ensime - as per original discussion.

rorygraves:
@fommil @edani @aemoncannon 
So this is a rework of the original fix - got a bit carried away, but DebugManager definately has less errors in it right now.  Fundamental change is that the ```sender !``` only occurs in the Receive block and everything else returns an RpcResponse and the type system helps to ensure this is the case.  I found a few cases where a non-response value or, in one case, no value would be returned.
So in summary, fixed the stop thing, cleaned up a bunch of other stuff...

edani:
@rorygraves   This looks good to me. I'll see if I can reproduce the appveyor failure locally.

fommil:
Should we stop all VM that we started in the post Stop ?

rorygraves:
Excellent point. Yes.

rorygraves:
Thanks @edani

PR #960, opened 2015-05-07
Rename rpcXXX methods to XXX in api

rorygraves:


rorygraves:
I think the commit method and the changes are fairly self explainatory - basically this a small step on the Ensime as an api path.

rorygraves:
@fommil  Any idea whats going on with Shippable/elpa?

fommil:
@rorygraves yeah, they screwed up again https://github.com/Shippable/support/issues/1419

rorygraves:
ok, that makes me feel a little better...

fommil:
oh, but I think the emacs tests actually did time out this time.

PR #962, opened 2015-05-09
JSON Jerk

fommil:
The beginnings of the WebSockets / JSON protocol for ENSIME :grin:

This has shapeless magic that is more generally useful than just ENSIME, so is factored into a separate `spray-json-shapeless` module. @sirthias and @milessabin may be interested in perusing it further and I'd be greatly obliged if you could review. I've factored it into a separate module and commit, completely independent of the ENSIME codebase (and hence others can depend on our snapshot jars when this is merged to master).

But don't get too excited just yet:

1. ~~I'm depending on shapless/pimpathon SNAPSHOTs so I need to wait for upstream releases before this can be merged.~~ upstream made releases
2. there are major performance problems when deriving marshallers for large families. The workaround will require modifying the API case class hierarchy, which is undesirable (I'll do this in a follow-up PR) https://github.com/milessabin/shapeless/issues/381
3. this isn't active on an endpoint yet. I have some code that can be easily used for a WebSockets / REST server in spray, but it relies on a single family of input messages and a single family of output messages... so fixing shapeless and refactoring our API is a prerequisite.

I'll be turning my attention to the Scala Days talk over the coming weeks, so probably won't get to the follow-ups mentioned above for a while.


spray-json-shapeless/src/main/scala/org/ensime/json/FamilyFormats.scala
```diff
@@ -0,0 +1,391 @@
+package org.ensime.json
```
rorygraves: Is the goal for this package to break out into into shapeless contrib?

rorygraves:
This looks good to me - would be nice if the build passed ;)

spray-json-shapeless/src/main/scala/org/ensime/json/FamilyFormats.scala
milessabin: shapeless-contrib is already too much of a grab-bag of random stuff ... it'd be better if it was broken out into a standalone project.

fommil:
I just pushed a couple of commits I had sitting on my home computer from last week.

fommil:
Cool thanks! Now I have to rewrite that slide where I said it was in a PR :-P

PR #966, opened 2015-05-13
Return rpc abort rather than reader error if possible for parsing err…

rorygraves:
…ors.

Parse failures of RPC messages result in protocol errors rather than rpc abort which updates Emacs.
So if we can work out the rpc call id return an rpc-abort rather than a reader error (otherwise fall
back to a reader error:


rorygraves:
@fommil Shippable appears to be borked - the jobs timeout without ever actually starting.

edani:
@rorygraves  This looks good and I ran all tests.

PR #994, opened 2015-06-14
Shapeless performance fix

fommil:
many many thanks to @retronym for the performance fix in shapeless, and @milessabin for releasing so quickly.


rorygraves:
@fommil Cool stuff!  Is there anything outstanding on this (e.g. are we expecting appveyor to work at all?) - Otherwise I'm happy to merge.

fommil:
nope. BTW I turned off the dedicated hosts in shippable... their infrastructure seems able to handle our build. Maybe I was being harsh. Setting up drone is an independent thing we can do and probably worth it.

rorygraves:
Nice.

rorygraves:
I can't believe how fast a major perf issue, a shapeless fix and a ensime update happened after our talk.

fommil:
on a weekend, no less! It's almost like its some form of agile development. It must be wrong, let's burn it all down and create a plan.

rorygraves:
Better add a couple of layers of management - don't want this kind of stuff catching on.

PR #1004, opened 2015-06-20
Implicit conversion highlighting

edani:
I don't know if this should be merged before or after the giant Akka refactor but I thought I'd create thsi PR for people to try out the UI  (look for the matching one in ensime/ensime-emacs#140).

This change adds two new types of semantic highlighting.  It also adds an API call `implicitInfo` that takes a range and returns all implicit conversions within the range. Among the information it returns is `SymbolInfo` objects that can be used to visit the location where the implicits are declared.

The thing that took me so long was trying to return the implicts' names as they would appear in source code so the editor could expand them. But it's harder than I thought so I'm giving up for now.

### DIFF
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -25,6 +25,20 @@ case object ConnectionInfoReq extends RpcStartupRequest
 sealed trait RpcAnalyserRequest extends RpcRequest
 
 /**
+ * Request details about implicit conversions applied inside the given
+ * range.
+ *
+ * Responds with `List[ImplicitInfo]`.
+ *
+ * @param file source.
+ * @param range in the file to inspect.
+ */
+case class ImplicitInfoReq(
+  file: File,
+  range: OffsetRange
+) extends RpcAnalyserRequest
+
+/**
  * Responds with a `VoidResponse`.
  */
 case class RemoveFileReq(file: File) extends RpcAnalyserRequest
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -28,7 +28,7 @@ sealed trait RpcAnalyserRequest extends RpcRequest
  * Request details about implicit conversions applied inside the given
  * range.
  *
- * Responds with `List[ImplicitInfo]`.
+ * Responds with `ImplicitInfos`.
  *
  * @param file source.
  * @param range in the file to inspect.
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -203,7 +203,8 @@ sealed trait EntityInfo {
 object SourceSymbol {
   val allSymbols: List[SourceSymbol] = List(
     ObjectSymbol, ClassSymbol, TraitSymbol, PackageSymbol, ConstructorSymbol, ImportedNameSymbol, TypeParamSymbol,
-    ParamSymbol, VarFieldSymbol, ValFieldSymbol, OperatorFieldSymbol, VarSymbol, ValSymbol, FunctionCallSymbol
+    ParamSymbol, VarFieldSymbol, ValFieldSymbol, OperatorFieldSymbol, VarSymbol, ValSymbol, FunctionCallSymbol,
+    ImplicitConversionSymbol, ImplicitParamsSymbol
   )
 }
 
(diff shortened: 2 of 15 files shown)


fommil:
Great! I'll try this out and merge it into the akka refactoring.

fommil:
awesome stuff. I've had a quick look through from a compatibility point of view. There will be a lot of clashes, but it'll be trivial for me to go through and copy/paste the relevant lines onto my branch... not so easy the other way round. I'll try to do that today/tomorrow. So **don't merge** this branch, I'll pick it as a separate commit after mine (my PR is very nearly ready, so this won't delay things too much)

rorygraves:
This is really cool

aemoncannon:
Very cool. People have been asking for this for *years.*

fommil:
Great stuff! I've had a look through and I'll migrate tonight.

Some thoughts:

* the response type is a collection type instead of something solid, which means it can't be part of a single family response type. If you could create a container, that would be good.
* would it be possible to list the type parameters of implicit conversions and implicit parameters on a `C-c C-v e`? e.g. when over `dir / ".ensime_cache"` in `EnsimeConfigSpec.scala` it would be nice to see message "Implicit conversion dir: File => FilePimps" instead of "Implicit conversion of dir using FilePimps" and the same when listing implicit parameters (show the type as well).
* opening up `EnsimeConfigSpec` shows off-by-one underlines for the right hand side parameters to some `===` lines. e.g. `assert(module1.targetDirs.size === 1)` has the final `)` underlined instead of the `1`.
* In the darcula mode I've opted for white underline rather than reusing the warning face. Maybe a light gray underline for the default theme? We don't want people thinking they have warnings in their code (although @rorygraves will probably say any use of an implicit **should be** a warning! :smile:)

This is **super** awesome btw :grinning:

fommil:
aaah, I've just worked out why there is an "off by one"... it's parameters :smile: Hmm, I wonder if there is another way to do that. Overlay on the method name instead?

fommil:
@edani I've rebased for you on top of the Akka refactor. Just `git fetch fommil ; git reset --hard fommil/implicits ; git sith`

edani:
@fommil  Thanks for the feedback, this is exactly what I needed. I'll add commits to address your comments.

Regarding where to put the highlight on implicit parameters, I hesitated between putting it at the end (where the explicit params would be) or on the function name (because it looks better). I think it comes down to what would be best for multi-line applications. What does everyone think?

fommil:
The more I'm using this, the more I'm getting used to the trailing underline for implicit parameters. @rorygraves had an idea that it would be great if we inserted virtual parenthesis characters and a way to expand those virtual parenthesis into virtual parameter lists. i.e. a big button that de-magics the source code.

hedefalk:
That idea sounds great.

I think Scala-IDE (Eclipse) did this better than IDEA. I remember having troubles with implicits when moving from Eclipse. Maybe worth looking into for inspiration on GUI.

fommil:
hmm, I'm not so keen on the little question marks in the gutter though. I keep thinking I have code warnings.

edani:
@fommil I'm changing the color of the question mark to the "compiler info" (some kind of gray in my theme). Maybe I can also switch the symbol to a large dot or something.  I wanted an indicator that meant "there's something on this line worth pressing `C-c C-v e`".

About adding virtual parens to mean implicit parameters: that seems possible with an overlay. I could experiment to see how it looks. It may be hard to deal with text terminals.

About expanding the parameters: that's something I thought about but it's harder than it looks :-)
The first problem is to find the parameters' name in scala (not the same as the symbol name) and generating imports. The second is to deal with operator-style function applications: we would have to rewrite things like `myList  map  { ...}`   into `myList.map({...})(...)`.  I'll see if scala-refactoring can be of help.

fommil:
Great ideas and fantastic that it might be possible to improve the UI. But as is, this in awesome already so I'm happy for all the improvements to come later.

fommil:
@edani if you could please update this PR with the rebase I sent you, I'd be happy to merge. If you have the time, it would be good to change the response type to be solid rather than a list but it's not absolutely essential.

edani:
Yes, I'l do that in the next day or so.

fommil:
I'm tempted to raise my own PR with your changes and merge :stuck_out_tongue: We've been showing this off to all our IntelliJ coworkers and making them jealous.

edani:
Be my guest, I can change the response type as a follow-up.

edani:
I pushed a change that makes the response a `ImplicitInfos`

hedefalk:
Aaawesome!

fommil:
:smile:

PR #1008, opened 2015-06-23
hotfix for Jerk RpcResponse marshalling

fommil:
@hedefalk can you please check that this works with a `publishLocal` before we merge?


hedefalk:
:+1:

Issue #1009, opened 2015-06-23
Completion of scala keywords


tbje:
Scala keywords like implicit and override does not show up in the completion list, would be nice if it did.

fommil:
For emacs, I recommend using the `company-code` backend to company mode.

FWIW, I'm not keen on keyword completions because they get in the way, so I do this

https://github.com/fommil/dotfiles/blob/master/.emacs.d/init.el#L489

```elisp
            (set (make-local-variable 'company-backends)
                 '(ensime-company (company-keywords company-etags)))
```

which means "use `ensime-company` when ENSIME is up, otherwise fallback to `keywords` and `etags`).

You'd want something like this

```elisp
            (set (make-local-variable 'company-backends)
                 '((ensime-company company-keywords) (company-keywords company-etags)))
```

fommil:
hmm, that setup does have the weird side effect that when ENSIME is down, and you type something that looks like a keyword, then you won't get etags completions (because company-mode only uses the first completer group that says it can do a completion)

fommil:
Hmm, I suppose that doesn't help if you're not on emacs. Maybe the server should add keyword completions to its list.

rorygraves:
Fixed in #1011

PR #1011, opened 2015-06-24
Add scala keywords to completion results.

tbje:
Simple potential fix for #1009. For discussion.

### DIFF
--- a/core/src/main/scala/org/ensime/core/Completion.scala
+++ b/core/src/main/scala/org/ensime/core/Completion.scala
@@ -232,61 +232,61 @@ trait CompletionControl {
 
     val typeSearchResults = typeSearch.flatMap(Await.result(_, Duration.Inf))
 
-    val keywords = Seq(
-      "abstract",
-      "case",
-      "catch",
-      "class",
-      "def",
-      "do",
-      "else",
-      "extends",
-      "false",
-      "final",
-      "finally",
-      "for",
-      "forSome",
-      "if",
-      "implicit",
-      "import",
-      "lazy",
-      "match",
-      "new",
-      "null",
-      "object",
-      "override",
-      "package",
-      "private",
-      "protected",
-      "return",
-      "sealed",
-      "super",
-      "this",
-      "throw",
-      "trait",
-      "try",
-      "true",
-      "type",
-      "val",
-      "var",
-      "while",
-      "with",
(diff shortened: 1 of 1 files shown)


fommil:
for reference https://github.com/company-mode/company-mode/commit/9c8364d51f542c0ee9d47806d43bcc13fd7f60a1

Cool, thanks. I think the CI machines are running particularly slowly so I've restarted.

The only concern I would raise is that the completions are not semantic. For example, typing `for` will suggest `forSome` even though you might be in a block. It's not a major deal but it would be really cool if we only suggested keywords that are valid for that point.

hedefalk:
Could we put these things under client-flags? In Atom, I also use a simple language-scala package which provides code-snippets and keywords and stuff like that inside the same autocomplete feature. Basically you register yourself as an "autocomplete provider" and both ensime-atom and language-scala plugins does this. So far I haven't really had any duplication problems, I just put the Ensime result (of max 5) on top, but with this change I think I will get duplicates.

I guess this is a deeper question: should Ensime be a full goto IDE by itselfe or should it play nice with other editors+plugins? I think we can do both if we just put some features under flags, right?

fommil:
yup, all good points and emacs has a similar thing. I'm tempted to say that maybe this should be a client side feature, unless we can do it contextually on the server. @rorygraves ?

aemoncannon:
We have the AST on the server but I'm not sure how expensive it would be to
'localize' the point and figure out what keywords are applicable.

rorygraves:
@hedefalk @fommil This has to be a server side thing for me - two reasons:
1. Everybody gets the same
2. logic improvements (e.g. as @aemoncannon using the ast for region identification) need to AST

Whilst it would be very quick to get working directly pushed into the editor completer to get really good support I think it has to be server side.

fommil:
can you please rebase? I pushed a small change to stabilise the build

tbje:
Providing the "correct" keyword would be great. I also tend to prefer the server side keyword completion as it will give you the out of the box experience. Setting up the right back-end for company-mode etc at least needs to be well documented, also it seems ensime will always provide some completions so the keywords won't show up before you've written a large part of the word (or never for implicit / throw ) if used as a fallback. Rebased and improved.

fommil:
thanks!

Issue #1006, opened 2015-06-22
sealed family for RpcResponses


fommil:
This should

- ~~speed up the JerkFormats compile~~
- ~~automate the serialisation of JerkFormats~~
- ~~use shapeless for File canonicalizing, move out of protocol~~
- websockets and REST

I **really** want to move us towards the noble goal that adding a new feature should be as easy as adding a new request to `incoming.scala`, corresponding response to `outgoing.scala`, ideally some tests to the `JerkFormats` and `SwankFormats`, and then just coding up the implementation of the feature in the relevant actor.

fommil:
This is all related to the ScalaX talk I'll give in December, so I'll work on it sooner rather than later. This might involve using shapeless magic in S-Express to be more like the JSON stuff.

PR #1019, opened 2015-06-27
Happy family

fommil:
@hedefalk you might want to check the changes/additions in the `JerkFormatSpec` assertions as this will break JERK.

Should not impact SWANK.

First step towards #1006


fommil:
Note to self: don't forget to update the docker image base scala version after merging.

fommil:
looks like I've introduced a regression in the emacs tests, its hanging forever in the format source test. I'm investigating.

fommil:
@edani @aemoncannon when I run the emacs tests locally, I'm seeing this kind of transient failure a lot now:

```
ENSIME ready. M-x be_cool
Received test event: :compiler-ready
Received :compiler-ready expecting (:compiler-ready :full-typecheck-finished). Waiting for the rest
Received test event: :full-typecheck-finished
Handling test event: :full-typecheck-finished/(:events (:full-typecheck-finished) :func (lambda (G63988) (ensime-test-run-with-handlers Test file with package name (this broke when -sourcepath param was used). (ensime-test-with-proj (proj src-files) (ensime-typecheck-all)))) :is-last nil)
Checking entire project...
Received test event: :region-sem-highlighted
Got :region-sem-highlighted, expecting (:full-typecheck-finished). Ignoring event.
Received test event: :return-value
Got :return-value, expecting (:full-typecheck-finished). Ignoring event.
```

Note there is a `Received test event: :full-typecheck-finished` up there... could we perhaps still have a timing issue when the server starts up "too quickly"?

aemoncannon:
Appears to be waiting for a full type check event that never arrives?

fommil:
@aemoncannon I think it has arrived already, but is being ignored.

rorygraves:
LGTM

fommil:
@hedefalk note that this will impact your JSON protocol

edani:
@fommil  I'll try to look into the transient test errors. Does this happen on more than one unit test? Also what's your configuration? I'm assuming Linux and a not-too-old CPU?

fommil:
Linux and about 4 cores

hedefalk:
Ooh, this isn't published yet, is it?

I went ahead and published the client change https://github.com/hedefalk/ensime-atom/commit/21bc836be23750da106431f482919bd4e7308c29 to atom repo already…

fommil:
@hedefalk should be, unless the ever reliable shippable has let us down again.

hedefalk:
great!

Issue #363, opened 2014-06-13
use of @deprecated methods/apis to use strikethrough face


fommil:
as a visual indicator that it is bad.

Ideally, this would also be in the method completion dropdown.

PR #1021, opened 2015-06-27
@deprecated methods/apis use strikethrough face #363

ale64bit:
For the moment, highlights deprecated classes, traits, objects and function calls. I would appreciate some hints to integrate it to the method completion dropdown.

### DIFF
--- a/core/src/main/scala/org/ensime/core/SemanticHighlighting.scala
+++ b/core/src/main/scala/org/ensime/core/SemanticHighlighting.scala
@@ -50,36 +50,41 @@ class SemanticHighlighting(val global: RichPresentationCompiler) extends Compile
           add(TypeParamSymbol)
         } else if (sym.hasFlag(PARAM)) {
           add(ParamSymbol)
-        } else if (sym.hasFlag(ACCESSOR)) {
-          val under = sym.accessed
-          if (under.isVariable) {
-            add(VarFieldSymbol)
-          } else if (under.isValue) {
-            add(ValFieldSymbol)
+        } else {
+          if (sym.isDeprecated) {
+            add(DeprecatedSymbol)
+          }
+          if (sym.hasFlag(ACCESSOR)) {
+            val under = sym.accessed
+            if (under.isVariable) {
+              add(VarFieldSymbol)
+            } else if (under.isValue) {
+              add(ValFieldSymbol)
+            } else {
+              false
+            }
+          } else if (sym.isMethod) {
+            if (sym.nameString == "apply" || sym.nameString == "update") { true }
+            else if (sym.name.isOperatorName) {
+              add(OperatorFieldSymbol)
+            } else {
+              add(FunctionCallSymbol)
+            }
+          } else if (sym.isVariable && sym.isLocalToBlock) {
+            add(VarSymbol)
+          } else if (sym.isValue && sym.isLocalToBlock) {
+            add(ValSymbol)
+          } else if (sym.hasPackageFlag) {
+            add(PackageSymbol)
+          } else if (sym.isTrait) {
+            add(TraitSymbol)
+          } else if (sym.isClass) {
+            add(ClassSymbol)
+          } else if (sym.isModule) {
+            add(ObjectSymbol)
           } else {
             false
           }
-        } else if (sym.isMethod) {
-          if (sym.nameString == "apply" || sym.nameString == "update") { true }
-          else if (sym.name.isOperatorName) {
-            add(OperatorFieldSymbol)
-          } else {
-            add(FunctionCallSymbol)
-          }
-        } else if (sym.isVariable && sym.isLocalToBlock) {
-          add(VarSymbol)
-        } else if (sym.isValue && sym.isLocalToBlock) {
-          add(ValSymbol)
-        } else if (sym.hasPackageFlag) {
-          add(PackageSymbol)
-        } else if (sym.isTrait) {
-          add(TraitSymbol)
-        } else if (sym.isClass) {
-          add(ClassSymbol)
-        } else if (sym.isModule) {
-          add(ObjectSymbol)
-        } else {
-          false
         }
       }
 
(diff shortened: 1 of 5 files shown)


rorygraves:
@ale64bit Very cool stuff - a couple of items:
1) You build is failing on formatting - normally running ```sbt compile``` will reformat the code to meet the standard and you just need to commit it - it actually looks slightly confused about the comments within the list.

2) This is more one for @edani @fommil - is it possible to layer designations - i.e. for a depreated val can I simply add 'ValSymbol' and ```DeprecatedSymbol``` rather than having to have a suite of ```DeprecatedXXXSymbol```?  There does not appear to be anything in the code preventing it, and it would make the server code much nicer 
```scala 
if(sym.isDeprecated)
  add(DeprecatedSymbol)
```

rather than a lot of almost duplicated code.  I guess this comes down to whether the client side can layer the designations.

ale64bit:
@rorygraves Thanks for the feedback!

1) I don't know what could cause the formatting failure. Anywhere I can check that before commit the changes?

2) Definitely I agree to DRY the code. Initially, I had a single `DeprecatedSymbol` object, but the problem with it, is that the semantic highlighting doesn't differ in the editor between different kinds of deprecated objects (e.g. a deprecated `trait` looks the same as a deprecated function call) and that wasn't acceptable (IntelliJ differs between those). Any suggestions would be nice.

rorygraves:
@ale64bit  No problem - thanks for taking the time to contribute :smile: 

1) Did you do an ```sbt it:compile it:test``` ?

2) I agree, they should be highlighted + strikethrough, what I was hoping was it was possible to fix a strikethrough definition on top of the colour etc from the other designation (i.e. if they could be stacked) - I don't know enough about the emacs side rendering to answer the question.  If it doesn't it would be a really nice feature to have.

rorygraves:
I normally do ```sbt clean test it:test```

rorygraves:
If you look at the shippable.yml script section you can see what the build runs.

ale64bit:
That's weird, I'm usually doing `sbt clean test it:test` also. I will just take out the comment from the `List` just in case.

ale64bit:
About 2), the problem is that each `SourceSymbol` is mapped to a `String` in the symbol designations request, and that's what you get in the editor side. What I could do, is to unify everything with a single `DeprecatedSymbol` and then in the mapping, differ between each kind of symbol somehow so that the client can handle each deprecated variant.

Any suggestions are well appreciated! :D  I got familiar with the code base just yesterday, so I don't want to mess it up.

rorygraves:
The symbol designations are just a list of text region and tag
```scala
case class SymbolDesignations(
  file: File,
  syms: List[SymbolDesignation]
)

case class SymbolDesignation(
  start: Int,
  end: Int,
  symType: SourceSymbol
)
```

There is nothing preventing two SymbolDesignations overlapping, but I'm not sure how its handled on the emacs side.  So (1, 3, "var") and (1,3,"deprecated") could co-exist if the emacs rendering is smart enough to apply both (thats my hope).

fommil:
Cool! If you base on top of my PR instead of master it should make the marshalling a lot easier for you. I'll have a proper look today or tomorrow.

fommil:
You realise this is going to make my/Rory's work project look like a train track :stuck_out_tongue:

fommil:
Re layers I'm sure we handle it because this is what the implicit stuff does.

ale64bit:
@fommil oops, sorry. Skipped your rebase message. About the warning, I suppose that if `-deprecation` is enabled you will get them for free.

As suggested, layering works. Substituted all deprecated symbols by one and stack it on top of others. There's still something that is missing: how can I differ between a val/var declaration and its usage? Because otherwise it will strike-through the declaration of val/var and that's not intended.

@fommil @rorygraves btw, amazing work! ENSIME looks like the kind of thing that people will get addicted to.

fommil:
You mean if you declare a val and assign it to a deprecated value? I'd expect that to be strikethrough actually.

ale64bit:
No, I mean that in this code
```scala
@deprecated("stinks", "1.0")
val foo = 42
val bar = foo + 1
```
only the second foo must be strike-through.

fommil:
I'd be ok with the first `foo` being strikethrough as well.

rorygraves:
Me too

ale64bit:
@fommil @rorygraves Done. Works as intended, highlights only the usage, not the declaration.

core/src/main/scala/org/ensime/core/SemanticHighlighting.scala
```diff
@@ -50,36 +50,41 @@ class SemanticHighlighting(val global: RichPresentationCompiler) extends Compile
...
-        } else if (sym.hasFlag(ACCESSOR)) {
-          val under = sym.accessed
-          if (under.isVariable) {
-            add(VarFieldSymbol)
-          } else if (under.isValue) {
-            add(ValFieldSymbol)
-          } else {
-            false
+        } else {
+          if (sym.isDeprecated) {
```
fommil: can you do a standard `else if` here please? It'll avoid the indentation of everything else I think.

fommil: or does that break the flow-through? hmm.

fommil:
actually ignore my comment about rebasing to my branch, it doesn't affect you because you haven't introduced any new incoming or outgoing messages.

fommil:
but if you could please squash your commits, that would be great.

core/src/main/scala/org/ensime/core/SemanticHighlighting.scala
ale64bit: I nested that `else`, because only at that point I'm sure that I can add the symbol as deprecated and I don't have to ask about deprecation inside every remaining `if`. Otherwise, I will have to ask about deprecation of particular symbols and the code gets verbose. WDYT?

fommil: ok, I think what you've done is sensible.

ale64bit:
@fommil done squashing commits.

fommil:
cool! One last very small thing, can you add a line note in the `swank/CHANGELOG.md`? It's a pain to maintain this, but it helps people.

ale64bit:
@fommil I missed that one. Done :smile:

fommil:
awesome, thanks! I've always wanted this.

fommil:
we'll wait until this dists for the server then reruns the corresponding emacs CI tests. Updates like this to both server and client need to be well timed.

PR #1024, opened 2015-06-29
canonising File with shapeless

fommil:


fommil:
@edani this should reduce the problems with windows tests. Canonising is no longer performed in the protocol layer, it is performed on the domain objects just before/after marshalling.

edani:
OK I'll test this as soon as I can.

fommil:
I need to pull out the `spray-json-shapless` stuff for some people, not least my work project.

core/src/test/scala/org/ensime/config/EnsimeConfigSpec.scala
```diff
@@ -26,27 +21,26 @@ class EnsimeConfigSpec extends FunSpec with Matchers {
...
+        abc.mkdirs()
+        cache.mkdirs()
 
         test(dir, s"""
 (:name "project"
  :scala-version "2.10.4"
- :java-home $javaHome
- :root-dir $dirStr
- :cache-dir $cacheStr
+ :java-home "$javaHome"
```
edani: This won't work, $javaHome etc. contain backslashes which break the lisp parser. They really must be escaped.

swank/src/test/scala/org/ensime/server/protocol/swank/SwankFormatsSpec.scala
```diff
@@ -58,24 +57,24 @@ class SwankFormatsSpec extends FlatSpec with Matchers with EnsimeTestData {
...
-      s"""(swank:import-suggestions $file1_str 1 ("foo" "bar") 10)""",
+      s"""(swank:import-suggestions "$file1" 1 ("foo" "bar") 10)""",
       ImportSuggestionsReq(file1, 1, List("foo", "bar"), 10): RpcRequest
     )
   }
 
   it should "unmarshal RpcAnalyserRequests" in {
     unmarshal(
-      s"""(swank:remove-file $file1_str)""",
+      s"""(swank:remove-file "$file1")""",
```
edani: The "unmarshal" tests fail: $file1 etc  contain unescaped backslashes that break string parsing

swank/src/test/scala/org/ensime/server/protocol/swank/SwankFormatsSpec.scala
```diff
@@ -368,7 +367,7 @@ class SwankFormatsSpec extends FlatSpec with Matchers with EnsimeTestData {
     )
     marshal(
       DebugExceptionEvent(33L, dtid, "threadNameStr", Some(sourcePos1.file), Some(sourcePos1.line)): EnsimeEvent,
-      s"""(:debug-event (:type exception :exception 33 :thread-id "13" :thread-name "threadNameStr" :file $file1_str :line 57))"""
+      s"""(:debug-event (:type exception :exception 33 :thread-id "13" :thread-name "threadNameStr" :file "$file1" :line 57))"""
```
edani: The "marshal" tests fail: marshalling escapes backslashes which means that the expected string isn't equal to "$file1"

jerk/src/test/scala/org/ensime/jerk/JerkFormatsSpec.scala
```diff
@@ -71,24 +59,24 @@ class JerkFormatsSpec extends FlatSpec with Matchers
 
     roundtrip(
       ImportSuggestionsReq(file1, 1, List("foo", "bar"), 10),
-      s"""{"point":1,"maxResults":10,"names":["foo","bar"],"typehint":"ImportSuggestionsReq","file":$file1_str}"""
+      s"""{"point":1,"maxResults":10,"names":["foo","bar"],"typehint":"ImportSuggestionsReq","file":"$file1"}"""
```
edani: The Jerk tests fail for the same reasons as the Sank tests.

edani:
@fommil  Can you try to update this PR with all (?) the test changes reverted? I don't think we can get working protocol tests without escaping backslashes.

fommil:
Ah, it was just backslashes. OK in that case a good place to backslash is in the marshal/roundtrip comparison. I'll try to get AppVeyor working so I can see this myself.

fommil:
Looks like we can do this:

https://github.com/scala/scala/blob/2.11.x/test/junit/scala/StringContextTest.scala#L63

which allows us to automatically escape anything of a certain type. That means we don't even need to think about it. I'll try and get AppVeyor working and then I can cycle this.

fommil:
@FeodorFitsner ever since we got an `appveyor.yml` into `master`, our Github Pull Request build status appears to have been hijacked by AppVeyor. Are you doing anything special that would be hiding our other builds? (i.e. Drone, Shippable, Coveralls, possibly Travis.)

fommil:
yay!

FeodorFitsner:
I'm just wondering why I can't see Shippable build details. Is it like a private/hidden build there?

image

fommil:
@FeodorFitsner sorry I pulled you in earlier because it looked like AppVeyor hijacked the build status, but they all came back when I force pushed. A glitch probably in github. Yeah, shippable is so very nearly useful... We're swapping to drone for our docker builds. Linux builds are running the integration tests, we'd love to have them in windows too, just I'd ncase you're eyeballing the timings ;-)

fommil:
I've updated this. It is using a SNAPSHOT of shapeless, but miles says there will be a release soon.

PR #1028, opened 2015-07-01
support multiple connections close #1027

### LINKED ISSUE #1027: allow reconnections and multiple clients
rory says I busted this, so I'll fix it. I have a branch already, but I need to write some testing around a new class I introduced.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23469703-allow-reconnections-and-multiple-clients?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github).
</bountysource-plugin>

fommil:


rorygraves:
@fommil - https://github.com/rorygraves/ensime-server/tree/sam_reconnecting

rorygraves:
Code looks good to me - if you can pull my change on the top it fixes my issues.

fommil:
@rorygraves ! ping

rorygraves:
@fommil  Is this a real build error or a transient emacs side issue?

fommil:
Hmm, since its failing so epically I'll look into it. Could just be shippable being its usual crap self.

```
(error Invalid version string: 'nil') error in process filter: version-to-list
```

fommil:
now using drone as the CI, so hopefully we'll get some answers about the stability of this build.

fommil:
```
Starting test: 'Test inspect type at point.'
Copying /tmp/ensime_test_proj_2169OAl/cache to /tmp/ensime_test_proj_2169bKr/cache
ENSIME server starting...
Connecting to Swank on port 54862..
Received test event: :indexer-ready
Received :indexer-ready expecting (:connected :compiler-ready :indexer-ready :full-typecheck-finished). Waiting for the rest
ENSIME ready. Hack and be merry!
error in process filter: Invalid version string: 'nil'
```

so looks like a regression. Weird. I'll have a look tomorrow.

fommil:
I actually have no idea how that error could possibly have happened. The only thing that could trigger an `Invalid version string` is a call to `ensime--test-scala-version` and it is set to read `SCALA_VERSION` and by this point it was already called and gave an answer back. Somehow emacs-lisp managed to set the var to `nil`... might be a `volatile` memory bug in emacs, you never know.

fommil:
hmm, no... I think this is a timing issue. Worth noting that I don't see the problem on my local tests. The `connection-info` handler must be getting something that doesn't have the version string in it.

fommil:
wow, I think I found the problem... it's a client bug! This is what was happening:

* async listen for connection-info
* wait for indexer/analyser
* send requests

but in this new world, the indexer/analyser are basically ready immediately, so we got our `connection-info` **after** we started sending requests, and we didn't know the protocol version... hence the error.

This took a lot longer to work out than I am going to admit. But in my defence, I've been writing C and Java most of the day.

fommil:
ok, I workaround it in the server with an evil hack... send the "ready" messages immediately after the connection-info. *shudder*

fommil:
self merging, because rory has already reviewed and this is blocking dev in atom and sublime.

rorygraves:
Very weird bug - thanks @fommil

PR #1030, opened 2015-07-02
drone might support emacs tests

fommil:


fommil:
ha! Screw you shippable.

PR #1033, opened 2015-07-03
improve drone build

fommil:


fommil:
I'm pretty sure this isn't good

```
/usr/local/bin/drone: line 57:  2177 Segmentation fault      ensime-emacs/test/run_emacs_tests.sh
```

fommil:
ok, looks like emacs-24.1 has a bug in it so we might not be able to matrix build against it in `-batch` mode. This is probably why the tests weren't doing it this way in the first place.

PR #1038, opened 2015-07-08
use shapeless instead of the reflection API

fommil:
to help @koshelev


fommil:
Transient on drone. We still have some timing issues in the Emacs code.

fommil:
another transient, ok we need to fix this

PR #1039, opened 2015-07-08
Run it:test in appveyor

edani:
Trying to diagnose test failure... "it works on my box"!
~~Do not merge, I removed .drone.yml to save time~~

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,19 +8,27 @@ install:
         )
         [System.IO.Compression.ZipFile]::ExtractToDirectory("C:\sbt-bin.zip", "C:\sbt")
       }
+  - ps: |
+      if (!(Test-Path -Path "C:\Program Files\Java\jdk1.7.0\src.zip")) {
+        (new-object System.Net.WebClient).DownloadFile(
+          'http://downloads.sourceforge.net/project/jdk7src/jdk7u10/2013-01-04/src-jdk.zip',
+          'C:\Program Files\Java\jdk1.7.0\src.zip'
+        )
+      }
   - cmd: SET JDK_HOME=C:\Program Files\Java\jdk1.7.0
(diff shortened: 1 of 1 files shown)


fommil:
You sure your machine isn't just a Linux box with a picture of some rolling green hills on it? :-P

edani:
@fommil Do you know if there's a way to inspect appveyor's machine, especially to look at log files?

fommil:
I'm not sure. Maybe as a quick hack you could change the run script to just run the failing test and change the logging definition to go to stdout?

edani:
@fommil  I got past the first problem which was just the indexer taking too long. Now we're missing "src.zip: in the jdk directory ( `C:\Program Files\Java\jdk1.7.0"` ) . Should we raise an issue with the appveyor people?

fommil:
@FeodorFitsner is there a way for us to request that the Java sources be bundled with the image that we're using?

(Failing that, I suppose we could download them ourselves)

edani:
The latest commit I pushed should work. I needed up downloading  src.zip from sourceforge. I couldn't find a way to get jdk 6 from Oracle without creating an account.

Also, I'm having problems running the emacs test suite in Windows: it seems that in emacs `--batch` mode, `start-process` doesn't work as expected. The process starts and appears to run, but it never exits nor sends any output to emacs. It works fine in interactive mode, obviously. I can come up with a small example if anyone wants to take a look.

edani:
Boo, the drone CI is hanging during the emacs test suite. I'll do a new push with more logging enabled.

fommil:
ha, just when we thought we had it fixed.

edani:
The CI worked after all, I just assumed it was hanging...

fommil:
I hit build again :-) sounds like we still have Emacs transients

PR #1043, opened 2015-07-13
Replace VoidResponse with TrueResponse for DebugManager

rorygraves:
+ a few random cleanups along the way.

This addresses issues with the Ensime-sublime plugin as the return value changed from true to nil at some point during the refactors, breaking the sublime debugging support.


rorygraves:
whoops - broke formatting too - will fix later

rorygraves:
@fommil - Back to you

PR #1053, opened 2015-08-04
HACK: Avoid keeping stale symbols when reloading a single file

### LINKED ISSUE #1052: Spurious type errors are reported when trying to use dotty with Ensime
I'd like to hack on [dotty](http://github.com/lampepfl/dotty) using Ensime, but it keeps hitting type errors even though the project compiles fine with `scalac`, here's how to reproduce what I'm seeing:
- Open `src/dotty/tools/dotc/core/Symbols.scala`
- Launch ENSIME with `M-x ensime` (after having followed the [Quick Start Guide](https://github.com/ensime/ensime-emacs/wiki/Quick-Start-Guide))
- Enter a single space, in the file, remove it, save the file
- Press `M-n` to find the first spurious type error, which will look like:
```scala
type mismatch;
found   : dotty.tools.dotc.core.Symbols.(some other)Symbol(in object Symbols){type ThisName = N}
required: dotty.tools.dotc.core.Symbols.Symbol(in object Symbols)
```
Notice that the types should match here, they don't because the found type is `(some other)Foo` instead of `Foo`, which makes me suspect that Ensime is keeping around stale symbols which confuses the presentation compiler.

Also note that this does not happen with scala-ide, so this does not seem to be a bug in the presentation compiler itself.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25235508-spurious-type-errors-are-reported-when-trying-to-use-dotty-with-ensime?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github).
</bountysource-plugin>

smarter:
This hack fixes #1052 for me, but I don't pretend to understand why since this is my first time looking at this code and at the presentation compiler. Basically, I noticed that [`Global#reload`](https://github.com/scala/scala/blob/2.11.x/src/interactive/scala/tools/nsc/interactive/Global.scala#L763-L769) (which is indirectly called in Ensime using `askReload`) does not remove the symbols corresponding to a previous loading of the current coding units, this is only done by [`Global#syncTopLevelSyms`](https://github.com/scala/scala/blob/2.11.x/src/interactive/scala/tools/nsc/interactive/Global.scala#L673-L689) which is called indirectly by [`Global#typedTree`](https://github.com/scala/scala/blob/2.11.x/src/interactive/scala/tools/nsc/interactive/Global.scala#L664-L671) which is itself called by [`Global#waitLoadedTyped`](https://github.com/scala/scala/blob/2.11.x/src/interactive/scala/tools/nsc/interactive/Global.scala#L1150-L1174) which happens to be called by Ensime when `async = false`(I have no idea what is supposed to be asynchronous or why that matters): https://github.com/ensime/ensime-server/blob/master/core/src/main/scala/org/ensime/core/Analyzer.scala#L300-L301

Anyway, this is surely not the right solution but I hope it can lead to someone figuring out what should be done exactly.

### DIFF
--- a/core/src/main/scala/org/ensime/core/Analyzer.scala
+++ b/core/src/main/scala/org/ensime/core/Analyzer.scala
@@ -197,7 +197,7 @@ class Analyzer(
       }
       sender ! VoidResponse
     case TypecheckFileReq(fileInfo) =>
-      handleReloadFiles(List(fileInfo), async = true)
+      handleReloadFiles(List(fileInfo), async = false)
       sender ! VoidResponse
     case TypecheckFilesReq(files) =>
       handleReloadFiles(files.map(SourceFileInfo(_)), async = false)


sschaef:
I would make `handleReloadFiles` atomic (by wrapping it in a `Response`), otherwise you can't be sure that no race condition is involved. See how scala-ide is doing it: https://github.com/scala-ide/scala-ide/blob/429c32a54d47570ff09e5551bdce82a534f087e3/org.scala-ide.sdt.core/src/org/scalaide/core/internal/compiler/ScalaPresentationCompiler.scala#L280 We don't even ask for typechecking afterwards because it is done by need. But I don't know how this works out for ensime.

fommil:
cool! Good that you've found something that works for you. We'd still like to do this async though, so perhaps we're not waiting for the right message to come back before changing the state of the analyzer actor

fommil:
@sschaef I was just talking with @rorygraves about how we'd love to have this common library between ensime and scala-ide again. How much work would it be to create an eclipse-independent library from the scala-ide code? I think that's the right way to start... we're happy to refactor to use it (and would definitely contribute!)

fommil:
I'll have a look later to see if our actors are not managing the state correctly

fommil:
@smarter can you please try #1054

sschaef:
We depend on quite some powerful Eclipse APIs, we first have to do deeper analysis on how much we depend on Eclipse, I can't say how much work is required to extract an Eclipse independent scala-ide core. But I don't think we have the resources to create such a core before the end of the year. Our current goal is to first ship sbt-server support, after that we wanted to address dotc support and therefore larger refactorings in our code base.

fommil:
that would be awesome! Beneficial to both of us.

PR #1054, opened 2015-08-04
wait for typechecks to complete

fommil:
I guess the thing we're losing with this is that while we're typechecking anything, the whole analyzer will be blocked up. But at least we know where the race condition is now, thanks @smarter

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,0 +1,4 @@
+-XX:MaxPermSize=256m
+-Xss2m
+-Xmx2g
+
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,8 +18,8 @@ install:
   - cmd: SET JDK_HOME=C:\Program Files\Java\jdk1.7.0
   - cmd: SET JAVA_HOME=C:\Program Files\Java\jdk1.7.0
   - cmd: SET PATH=C:\sbt\sbt\bin;%JDK_HOME%\bin;%PATH%
-  - cmd: SET SBT_OPTS=-XX:MaxPermSize=2g -Xmx4g -Xss2m
-  - cmd: SET AKKA_TEST_TIMEFACTOR=3
+  - cmd: SET SBT_OPTS=-XX:MaxPermSize=256m -Xss2m -Xmx2g
+  - cmd: SET AKKA_TEST_TIMEFACTOR=10
   - cmd: mkdir .ensime_cache
 build_script:
   - sbt clean compile test:compile it:compile
--- a/core/src/it/scala/org/ensime/fixture/ProjectFixture.scala
+++ b/core/src/it/scala/org/ensime/fixture/ProjectFixture.scala
@@ -22,16 +22,16 @@ object ProjectFixture extends Matchers {
     import testkit._
 
     val probe = TestProbe()
-    val project = TestActorRef[Project](Project(probe.ref), "project")
-
-    project ! ConnectionInfoReq
-    expectMsg(ConnectionInfo())
-
     probe.ignoreMsg {
       // these are too noisy for tests
       case e: SendBackgroundMessageEvent => true
       case e: DebugOutputEvent => true
     }
+
+    val project = TestActorRef[Project](Project(probe.ref), "project")
+
+    project ! ConnectionInfoReq
+    expectMsg(ConnectionInfo())
 
     probe.receiveN(3) should contain only (
       Broadcaster.Persist(AnalyzerReadyEvent),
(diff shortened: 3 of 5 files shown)


smarter:
This seems to work, thanks a lot!

fommil:
@smarter you're the one that found the problem, thank you! I still want to investigate why the `async` flag was ever there. I want to make sure we're not losing functionality by dropping it.

fommil:
https://github.com/ensime/ensime-server/commit/a0f9ad6be19844f00b831e1033d27570d62cab52#diff-b0dd9d7afb0d2dca2292a4efbe58fe5e introduced the sync wait on typechecking... but as we've seen, it didn't really work. @edani are you ok with this change?

edani:
Let me look at this and try to remember why I did things this way.

edani:
OK, I remember. The `swank:typecheck-file` call was supposed to be synchronous, to avoid race conditions (for instance, when `swank:symbol-designations` follows right after). `swank:typecheck-files` stayed asynchronous.  It looks like the async-ness got reversed in a0a5de7be1aa46605d5d1212aee256d29955d74b, so @smarter 's hack was going in the right direction. I would also make TypecheckFilesReq asynchronous again.

There must be something else going on though, we should be able to avoid the race condition whether `async` is on or off. But for now I'd go with #1053 as a stop gap.

fommil:
Note that this PR has the effect of making both synchronous by using actor state to manage the access to the PC. Do we really want to allow queries when the PC is reloading some files?

fommil:
Appveyor is timing out. They obviously downgraded the hardware. I'll push an update to to akka timeout later.

edani:
The thing is, the pres compiler is supposed to handle all this... On the ensime side, operations that rely on a file to be typechecked call askLoadedTyped on it, and it doesn't block for longer than it should.  Also, sending another "reload" command before the previous one has finished shouldn't block at all, it should interrupt the pres compiler.

fommil:
OK I'll make a small change to deal with that case and update the PR.

fommil:
ok, updated. Although I forgot to bump the appveyor timeout so we might still see problems there.

edani:
@fommil  Thanks, I'll take this for a spin...

fommil:
no functional changes in the latest push, just moved everything into `preStart` in `Analyzer` and bumped the build machine requirements.

fommil:
AppVeyor failed with

```
Vector(SendBackgroundMessageEvent(Initializing Analyzer. Please wait...,105), Persist(AnalyzerReadyEvent), Persist(FullTypeCheckCompleteEvent)) did not contain only (Persist(AnalyzerReadyEvent), Persist(FullTypeCheckCompleteEvent), Persist(IndexerReadyEvent))
```

which looks like indexing took too long.

**sigh** after all @edani 's hard work to get the integration tests working on Windows, it looks like we've been bumped to slow hardware again :cry: @FeodorFitsner is that a fair assessment?

fommil:
interesting:

```
java.lang.AssertionError: assertion failed: expected:  but got unexpected message SymbolSearchResults(List(TypeSearchResult(java.util.Random,Random,Class,Some(LineSourcePosition(C:\Users\appveyor\AppData\Local\Temp\1\ensime-it-BasicWorkflow-3981050149899545939.project\.ensime_cache\dep-src\source-jars\java\util\Random.java,91))), TypeSearchResult(scala.util.Random,Random,Class,Some(LineSourcePosition(C:\Users\appveyor\AppData\Local\Temp\1\ensime-it-BasicWorkflow-3981050149899545939.project\.ensime_cache\dep-src\source-jars\scala\util\Random.scala,21)))))
```

fommil:
oh, interesting. On Windows the actual class is being picked up, not the companion (as it is in the Linux build). I wonder if the version of Java has anything to do with this. In any case, we have a fragile test.

fommil:
damn, hit that sexp transient that shows up about once every 20 runs.

fommil:
the `Random` thing was a transient by the looks of it. Something to be aware of, and make the test less brittle.

fommil:
ok, we're green... if @edani and @smarter are happy with this, we can merge. Sorry it ended up turning into a cleanup of the `Analyzer` Actor as well as addressing the async issue.

edani:
@fommil  I found the root cause of the race condition, I should have a fix shortly. I think it should be fixable without a "stashing" state, so the analyzer can be more responsive when many files are loaded. I should have a PR on top of yours within a few hours.

fommil:
Oh OK, that'd be very good!

edani:
See #1058

PR #1055, opened 2015-08-05
DO NOT MERGE log network messages

edani:
Also pulls ensime-emacs from an alternate branch to do more client-side logging.


fommil:
You don't need to edit the code to do this BTW :-)

Just change the logback file in the Emacs side to have trace level for this class.

edani:
Ahh OK.

fommil:
Users can even provide a custom log .

Issue #472, opened 2014-07-01
natural search for type


fommil:
Imagine having https://github.com/ornicar/scalex in your editor and project-specific Scaladoc browser :-)

Not just using its web-interface, but having it work for everything on your classpath, with completion of types.

If you'd like this feature, you can help out by giving us some example queries and the expected responses that you'd want to see (if they are not obvious already). We want a simple, natural, query language not a strict DSL.

ShaneDelmore:
I think this is an incredibly useful idea, if it could find functions by signature available in the current project or libraries, instead of just open buffers.  On my team we are starting to see a lot of duplicated methods rewritten by multiple developers because it is easier to write yet another max function than it is find out if there is already one written by your coworkers you can re-use.

fommil:
:smile: I know, it'd be awesome. We're currently working towards stabilising for 1.0 but if you were able to help out by coming up with a way of indexing (in lucene) and persisting (in SQL/H2) the domain objects that we produce during classfile/depickle parsing, then we'd love to hear/chat more about this.

ShaneDelmore:
I just discovered the project, and am pretty new to Scala in general but if I am capable I would love to help out as I would like to do a little more open source work.  Baby steps, first I have to get ensime actually working (I am currently an intellij user and find it doesn't "just work" on multi-project sbt projects that intellij works with.  I would just end up adding noise right now I think but if I can get myself working in some fashion then I was planning on picking through your low hanging fruit to see if I could be useful.

fommil:
hmm, that's a shame. I thought it did "just work". Are you using the sbt plugin to generate the .ensime file?

We could modularise the task of working on this ticket if that helps:

1. create some standalone tests (also define user input/expected response)
2. think about how to index/persist the project's information for searching
3. implement it in our existing lucene/h2 persistence
4. add protocol support
5. add emacs GUI support

rorygraves:
@ShaneDelmore Dive in - any help you give (low hanging or otherwise) is all appreciated :smile:

aemoncannon:
+1 : )

adelbertc:
Hey folks,

Just had a brief chat with @fommil - I have entertained the idea of (attempting to) write a Hoogle-ish tool for Scala. I'm currently playing around scraping what I need from the compiler via a compiler plugin, but it looks like there's some similar discussion/work being done here.

[...]

PR #1057, opened 2015-08-06
Docserver

fommil:
backed up behind #1054

refactor at this stage with no extra functionality (but there is a lot of extra test coverage), but expect one final commit tomorrow to add a splash screen listing all the javadocs available for the project.

This is crying out for an HTML5 dev to help by adding a search bar to find documentation and is perhaps the first step towards #472 

BTW, we could also serve up source code using the same mechanism.

@aemoncannon do you mind having a look?


fommil:
oh bugger, I just realised I am returning relative paths instead of absolute paths. I had a hack I was going to add in. @aemoncannon would it be hard to change the emacs side to accept relative or absolute paths? If relative, read the port number from the `.ensime_cache/http` file and append the relative path.

aemoncannon:
@fommil 
So the idea here is to disentangle the docs stuff from the web serving stuff, so the http can be general to many things, correct?

I think absolute uris are preferable from an ease-of-implementation point of view. Especially now that we have 3 active client implementations!

Looks great in general.

core/src/main/scala/org/ensime/core/DocResolver.scala
```diff
@@ -109,16 +94,16 @@ class DocServer(
...
-        "#" + URLEncoder.encode(if (docType == Javadoc8) toJava8Anchor(s) else s, "UTF-8")
+        "#" + { if (docType == Javadoc8) toJava8Anchor(s) else s }
       }.getOrElse("")
-      s"http://localhost:${port.getOrElse(0)}/$jarName/$path$anchor"
+      s"$jarName/$path$anchor"
     } else {
       val scalaSig = maybeReplaceWithUsecase(jar, sig.scala)
       val anchor = scalaSig.fqn.mkString +
         scalaSig.member.map {
-          "@" + URLEncoder.encode(_, "UTF-8")
```
aemoncannon: it turned out we didn't need this?

fommil:
yup, exactly. Cool, I'll ping when I think it's ready for merging, but should only be small changes from here on.

core/src/main/scala/org/ensime/core/DocResolver.scala
fommil: yeah, I don't think we need to encode the links because all browsers do this automatically. Plus it makes it a lot easier to test.

fommil:
this will fail the emacs tests I think because it is waiting on https://github.com/ensime/ensime-emacs/pull/208

fommil:
these are transients in the appveyor build :cry: over to you @rorygraves

rorygraves:
I think address #1052, is a really bad commit message.

fommil:
yeah, the message should have the summary first

rorygraves:
@fommil Happy with the changes - fix the commit message and feel free to self-merge.

PR #1058, opened 2015-08-08
Fix prescomp race from #1052

### LINKED ISSUE #1052: Spurious type errors are reported when trying to use dotty with Ensime

edani:
Fixes #1052

I think I got it right. I removed the Analyzer state that stashes all requests, it's simpler to rely on the locking provided by the pres compiler. This should keep operations from waiting for longer than they have to.

I checked the test case that @smarter mentioned and it's looking good.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,0 +1,4 @@
+-XX:MaxPermSize=256m
+-Xss2m
+-Xmx2g
+
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,8 +18,8 @@ install:
   - cmd: SET JDK_HOME=C:\Program Files\Java\jdk1.7.0
   - cmd: SET JAVA_HOME=C:\Program Files\Java\jdk1.7.0
   - cmd: SET PATH=C:\sbt\sbt\bin;%JDK_HOME%\bin;%PATH%
-  - cmd: SET SBT_OPTS=-XX:MaxPermSize=2g -Xmx4g -Xss2m
-  - cmd: SET AKKA_TEST_TIMEFACTOR=3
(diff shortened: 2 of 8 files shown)


fommil:
cool! A lot of hard work to go and find out what needed to ask a reload of the prescomp there. I'll close my PR.

PR #1061, opened 2015-08-11
do not recreate InputStreamReaders as they can potentially buffer the

fommil:
underlying stream, as discussed in
https://github.com/ensime/ensime-emacs/pull/206

### DIFF
--- a/core/src/it/scala/org/ensime/fixture/ProjectFixture.scala
+++ b/core/src/it/scala/org/ensime/fixture/ProjectFixture.scala
@@ -26,6 +26,7 @@ object ProjectFixture extends Matchers {
       // these are too noisy for tests
       case e: SendBackgroundMessageEvent => true
       case e: DebugOutputEvent => true
+      case ClearAllScalaNotesEvent => true
     }
 
     val project = TestActorRef[Project](Project(probe.ref), "project")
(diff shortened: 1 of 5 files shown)


aemoncannon:
lgtm

rorygraves:
@fommil  Has this broken the integration tests - slightly confused because AFAIK we don't do message encoding and decoding.

fommil:
@rorygraves looks like a weird transient **sigh**

rorygraves:
Its your search one about Random vs Random$

fommil:
oh, and appveyor seems to depend on the hardware they give us. Maybe the timeouts need bumped again.

fommil:
@rorygraves that plus debugtest timeout, and linux failed for some unknown reason.

rorygraves:
More transient errors:
```
[info] - should open the test project *** FAILED ***
[info]   Vector(ClearAllScalaNotesEvent, ClearAllScalaNotesEvent) did not contain only (ClearAllScalaNotesEvent, FullTypeCheckCompleteEvent) (BasicWorkflow.scala:148)
[info]   org.scalatest.exceptions.TestFailedException:
```
It takes two messages expecting 1x ClearAllScalaNotes and 1x FullTypeCheckCompleteEvent - I wonder what causes it to see 2x ClearAllScalaNotes (PC restart??)

fommil:
ok, now linux tests pass.... but we think there is still a problem with the protocol. For outbound messages, the prefix number means the number of bytes, but for inbound it means the number of UTF-8 chars. The correct fix sounds like it should be number of bytes in both directions.

aemoncannon:
Nice catch on the bytes vs chars.

Issue #930, opened 2015-04-11
Requirement failed: FileSourceFileInfo does not exist


greenrd:
Whenever I visit a new Scala file in emacs, I get an RPC abort and the following exception:

```
16:55:38.893 [default-akka.actor.default-dispatcher-155] ERROR org.ensime.core.Analyzer - Error handling RPC: java.lang.IllegalArgumentException: requirement failed: FileSourceFileInfo(/home/user/foo/Bar.scala) does not exist
java.lang.IllegalArgumentException: requirement failed: FileSourceFileInfo(/home/user/foo/Bar.scala) does not exist
	at scala.Predef$.require(Predef.scala:233) ~[scala-library.jar:na]
	at org.ensime.core.Analyzer$$anonfun$handleReloadFiles$1.apply(Analyzer.scala:257) ~[ensime_2.10-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.Analyzer$$anonfun$handleReloadFiles$1.apply(Analyzer.scala:256) ~[ensime_2.10-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at scala.collection.immutable.List.foreach(List.scala:318) ~[scala-library.jar:na]
	at org.ensime.core.Analyzer.handleReloadFiles(Analyzer.scala:256) ~[ensime_2.10-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.Analyzer.process(Analyzer.scala:161) ~[ensime_2.10-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.Analyzer$$anonfun$receive$1.applyOrElse(Analyzer.scala:83) ~[ensime_2.10-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at akka.actor.Actor$class.aroundReceive(Actor.scala:465) ~[akka-actor_2.10-2.3.8.jar:na]
	at org.ensime.core.Analyzer.aroundReceive(Analyzer.scala:20) ~[ensime_2.10-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516) [akka-actor_2.10-2.3.8.jar:na]
[...]

PR #1062, opened 2015-08-14
Move to akka.io and improve error handling

rorygraves:
Fundamentally there are two changes here:

1. Move from raw sockets to akka.io for socket protocol handling.
2. Create a top level actor that is responsible for handling core actor failures and shutdown gracefully.


fommil:
this is looking good, I've been using it all day, but we need the fix for the missing file as that will impact a lot of people since we are more likely to see the server bail on errors because we're not restarting the analyser on exceptions, so that's something we should keep in mind over the coming weeks.

rorygraves:
@fommil  @greenrd  I've pushed an extra change to this PR to address #930 - the call should now return an error rather than causing the Analyser to fail.    I have not had time to test it in anger - I plan to add a test, but realistically that will not occur until tomorrow evening.  Feel free to have a play.

rorygraves:
@fommil  Added a test as well, missing file on typecheck reports an error rather than failing server - back to you.

rorygraves:
@rorygraves   - Aide memoir
```
Received test event: :indexer-ready
Received :indexer-ready/t expecting (:connected :compiler-ready :indexer-ready :full-typecheck-finished) with guard func nil. Waiting for the rest
06:29:48.068 DEBUG akka://ENSIME/user/ensime-main/tcp-server/con1 a.a.LocalActorRef - unhandled message from Actor[akka://ENSIME/user/ensime-main/project#1319839364]: RpcResponseEnvelope(Some(1),ConnectionInfo(None,EnsimeImplementation(ENSIME),0.8.17))
```

PR #1070, opened 2015-08-21
Encode symbols when looking them up in indexer

### LINKED ISSUE #835: support symbolic methods (those that are renamed in bytecode)
which is a bit annoying for scalaz and akka tell/ask

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8712706-support-symbolic-methods-those-that-are-renamed-in-bytecode?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github).
</bountysource-plugin>

edani:
Fix #835

This seems to work in most cases :-)   The exception is operators on strings etc. which try to look up things like `java.lang.String.$plus`. But that looks like a separate issue.

### DIFF
--- a/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
@@ -76,6 +76,25 @@ class RichPresentationCompilerSpec extends WordSpec with Matchers
         }
     }
 
+    "find source declaration of standard operator" in withPresCompiler { (config, cc) =>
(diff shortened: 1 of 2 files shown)


fommil:
Nice!

hedefalk:
Awesome!

rorygraves:
Good stuff!

PR #1078, opened 2015-08-30
Fix #1056: locate source of imported Java classes

### LINKED ISSUE #1056: go to definition fail: import java.io.File
When hacking in WebServerSpec in my docserver branch, I'm not able to get typo info on `File` when its in the import list.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25449667-go-to-definition-fail-import-java-io-file?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github).
</bountysource-plugin>

edani:
The problem was that there is  both a "java.io.File" class and "java.io.File" object. But there is obviously no "/java/io/File$.class" in the JRE jars.

Another way to fix this could be to have the indexer associate java "objects" to the appropriate .class files.


fommil:
:-)

PR #1090, opened 2015-09-12
single jar deploys

fommil:
sbt-scoverage now works, ~~but I've not enabled it because the debug tests fail when run with coverage (getting debug exceptions instead of values, things like that... probably interacting in a strange way with the profiled code)~~ but I've not made it fail the build if coverage is too low. We need #1091 for that.


fommil:
@sksamuel any ideas why we might be seeing exceptions when debugging code that has been profiled with scoverage? Do the instrumented classes throw any exceptions for tracing purposes?

fommil:
actually, I just remembered how to turn off the instrumentation of our test packages. I'll do that, it should fix it.

sksamuel:
They only write out to disk, so maybe there's an issue with that and you're
seeing an IOException ?

fommil:
@sksamuel I remembered that we needed to disable instrumentation of the code that we were running under debug mode.

fommil:
Great, a typo in the drone file

fommil:
sorry, CI is so slow... we're waiting on drone adding back in the multiple build hosts support. It's currently only using one node on one box (we have at least 3 idle)

fommil:
Damn you drone

fommil:
thank the maker!

PR #1094, opened 2015-09-14
Fix #997 : typeAt works on Import trees

### LINKED ISSUE #997: Errors hovering on some imports…
_From @hedefalk on June 16, 2015 9:7_

```
10:45:38.121 [Thread-10] INFO  o.e.s.protocol.swank.SwankProtocol - Received msg: (:swank-rpc (swank:type-at-point "/Users/viktor/dev/projects/ensime-server/server/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala" 46) 60)
10:45:38.122 [ENSIME-akka.actor.default-dispatcher-12] INFO  o.e.s.protocol.swank.SwankProtocol - Writing: (:return (:ok nil) 60)
```

error in callback: TypeError: Cannot read property ':arrow-type' of null

I get just (:ok nil) on java.util.File

_Copied from original issue: ensime/ensime-atom#18_

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/22212391-errors-hovering-on-some-imports?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github).
</bountysource-plugin>

edani:
Fix #997   Hovering on package names or imported classes isn't super useful but at least we don't return `nil` now. Hovering over an imported class member is what's important.

### DIFF
--- a/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
@@ -363,6 +363,34 @@ class RichPresentationCompilerSpec extends WordSpec with Matchers
         assert(supersNames.toSet === Set("pipo", "bidon", "Object", "Product", "Serializable", "Any"))
       }
 
+    "get type info for imports" in withPresCompiler { (config, cc) =>
+      val expected = Map(
+        "1.0" -> Some("java$"),
+        "1.1" -> Some("java$"),
+        "1.2" -> Some("java.io$"),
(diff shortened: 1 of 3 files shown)


fommil:
Great! BTW I'm starting a new contract today, which is windows based. So very glad that you've been keeping it in good shape :-)

PR #1108, opened 2015-10-03
First pass on java type checking, type at point, doc finding

aemoncannon:
Still rough. Please give it a high level look.

Note:  I had a hell of a time trying to go full netbeans. Just typechecking a file seems to require buying into their indexing framework. A walk through the debugger revealed all kinds of ide code (org.netbeans....editor??) being loaded, singletons talking to each other all over the place, etc. In the end I just used/ripped the minimum that I needed to make these tests pass (TreeUtilities is a copy paste job).

### DIFF
--- a/api/src/main/scala/org/ensime/api/config.scala
+++ b/api/src/main/scala/org/ensime/api/config.scala
@@ -25,6 +25,8 @@ case class EnsimeConfig(
   def root = rootDir
   def debugVMArgs = debugArgs
   def referenceSourceJars = referenceSourceRoots
+  def disableSourceMonitoring = true
+  def disableClassMonitoring = true
 
   // some marshalling libs (e.g. spray-json) might not like extra vals
   val modules = subprojects.map { module => (module.name, module) }.toMap
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -1,6 +1,7 @@
 package org.ensime.api
 
 import java.io.File
+import java.nio.charset.Charset
 
 /**
  * There should be exactly one `RpcResponseEnvelope` in response to an
@@ -67,6 +68,9 @@ case object CompilerRestartedEvent extends GeneralSwankEvent
 /** The presentation compiler has invalidated all existing notes.  */
 case object ClearAllScalaNotesEvent extends GeneralSwankEvent
 
+/** The presentation compiler has invalidated all existing notes.  */
+case object ClearAllJavaNotesEvent extends GeneralSwankEvent
+
 case class Note(
   file: String,
   msg: String,
@@ -79,6 +83,12 @@ case class Note(
 
 /** The presentation compiler is providing notes: e.g. errors, warnings. */
 case class NewScalaNotesEvent(
+  isFull: Boolean,
+  notes: List[Note]
+) extends GeneralSwankEvent
+
+/** The presentation compiler is providing notes: e.g. errors, warnings. */
+case class NewJavaNotesEvent(
   isFull: Boolean,
   notes: List[Note]
 ) extends GeneralSwankEvent
@@ -177,6 +187,8 @@ case class SourceFileInfo(
 ) {
   // keep the log file sane for unsaved files
   override def toString = s"SourceFileInfo($file,${contents.map(_ => "...")},$contentsIn)"
+  def isJava = file.getName().endsWith(".java")
+  def isScala = file.getName().endsWith(".scala")
 }
 
 sealed trait PatchOp {
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -187,8 +187,6 @@ case class SourceFileInfo(
 ) {
   // keep the log file sane for unsaved files
   override def toString = s"SourceFileInfo($file,${contents.map(_ => "...")},$contentsIn)"
-  def isJava = file.getName().endsWith(".java")
-  def isScala = file.getName().endsWith(".scala")
 }
 
 sealed trait PatchOp {
--- a/core/src/main/java/org/ensime/core/javac/CompilationInfo.java
+++ b/core/src/main/java/org/ensime/core/javac/CompilationInfo.java
@@ -1,0 +1,77 @@
+package org.ensime.core.javac;
+
+import com.sun.source.tree.*;
+import com.sun.source.tree.Tree.Kind;
+import com.sun.source.util.SourcePositions;
+import com.sun.source.util.TreePath;
+import com.sun.source.util.TreePathScanner;
+import com.sun.source.util.TreeScanner;
+import com.sun.tools.javac.api.JavacScope;
+import com.sun.tools.javac.code.Flags;
+import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.comp.AttrContext;
+import com.sun.tools.javac.comp.Enter;
+import com.sun.tools.javac.comp.Env;
+import com.sun.tools.javac.comp.Resolve;
+import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.tree.JCTree.JCClassDecl;
+import com.sun.tools.javac.tree.JCTree.JCMethodDecl;
+import com.sun.tools.javac.util.Context;
+import java.util.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.lang.model.SourceVersion;
+import javax.lang.model.element.*;
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.type.UnionType;
+import javax.lang.model.util.Types;
+import org.netbeans.api.java.lexer.JavaTokenId;
+import org.netbeans.api.java.source.JavaSource.Phase;
+import org.netbeans.api.lexer.TokenSequence;
+import org.netbeans.modules.java.source.builder.CommentHandlerService;
+import org.netbeans.modules.java.source.builder.CommentSetImpl;
+import org.netbeans.modules.java.source.pretty.ImportAnalysis2;
+import org.netbeans.modules.java.source.transform.ImmutableTreeTranslator;
+import org.netbeans.api.lexer.TokenHierarchy;
+import java.util.HashSet;
+import java.io.Reader;
+import java.io.IOException;
+import com.sun.source.util.Trees;
+import javax.lang.model.util.Elements;
+import com.sun.source.tree.CompilationUnitTree;
+import com.sun.source.util.JavacTask;
+import org.netbeans.api.lexer.InputAttributes;
+
+public final class CompilationInfo {
+    private final JavacTask javacTask;
+    private final Trees trees;
+    private final TokenHierarchy<Reader> tokenHierarchy;
+    private final CompilationUnitTree compilationUnit;
+
+    public CompilationInfo(final JavacTask javacTask,
+                           final CompilationUnitTree compilationUnit) throws IOException {
+        boolean ignoreEncodingErrors = true;
+        this.javacTask = javacTask;
+        this.trees = Trees.instance(javacTask);
+        this.tokenHierarchy =
+            TokenHierarchy.create(compilationUnit.getSourceFile().openReader(ignoreEncodingErrors),
+                                  JavaTokenId.language(),
+                                  new HashSet<JavaTokenId>(),
+                                  new InputAttributes());
+        this.compilationUnit = compilationUnit;
+    }
+
+    public CompilationInfo(final JavacTask javacTask,
+                           final Element element) throws IOException {
+        this(javacTask, (CompilationUnitTree)(Trees.instance(javacTask).getTree(element.getEnclosingElement())));
+    }
+
+    public JavacTask getJavacTask() { return javacTask; }
+    public Trees getTrees() { return trees; }
+    public Types getTypes() { return javacTask.getTypes(); }
+    public Elements getElements() { return javacTask.getElements(); }
+    public TokenHierarchy<Reader> getTokenHierarchy() { return tokenHierarchy; }
+    public CompilationUnitTree getCompilationUnit() { return compilationUnit; }
+}
--- a/core/src/main/java/org/ensime/core/javac/TreeUtilities.java
+++ b/core/src/main/java/org/ensime/core/javac/TreeUtilities.java
@@ -1,0 +1,853 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright 1997-2010 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Oracle and Java are registered trademarks of Oracle and/or its affiliates.
+ * Other names may be trademarks of their respective owners.
+ *
+ * The contents of this file are subject to the terms of either the GNU
+ * General Public License Version 2 only ("GPL") or the Common
+ * Development and Distribution License("CDDL") (collectively, the
+ * "License"). You may not use this file except in compliance with the
+ * License. You can obtain a copy of the License at
+ * http://www.netbeans.org/cddl-gplv2.html
+ * or nbbuild/licenses/CDDL-GPL-2-CP. See the License for the
+ * specific language governing permissions and limitations under the
+ * License.  When distributing the software, include this License Header
+ * Notice in each file and include the License file at
+ * nbbuild/licenses/CDDL-GPL-2-CP.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the GPL Version 2 section of the License file that
+ * accompanied this code. If applicable, add the following below the
+ * License Header, with the fields enclosed by brackets [] replaced by
+ * your own identifying information:
+ * "Portions Copyrighted [year] [name of copyright owner]"
+ *
+ * Contributor(s):
+ *
+ * The Original Software is NetBeans. The Initial Developer of the Original
+ * Software is Sun Microsystems, Inc. Portions Copyright 1997-2010 Sun
+ * Microsystems, Inc. All Rights Reserved.
+ *
+ * If you wish your version of this file to be governed by only the CDDL
+ * or only the GPL Version 2, indicate your decision by adding
+ * "[Contributor] elects to include this software in this distribution
+ * under the [CDDL or GPL Version 2] license." If you do not indicate a
+ * single choice of license, a recipient has the option to distribute
+ * your version of this file under either the CDDL, the GPL Version 2 or
+ * to extend the choice of license to its licensees as provided above.
+ * However, if you add GPL Version 2 code and therefore, elected the GPL
+ * Version 2 license, then the option applies only if the new code is
+ * made subject to such option by the copyright holder.
+ */
+package org.ensime.core.javac;
+
+import com.sun.source.tree.*;
+import com.sun.source.tree.Tree.Kind;
+import com.sun.source.util.SourcePositions;
+import com.sun.source.util.TreePath;
+import com.sun.source.util.TreePathScanner;
+import com.sun.source.util.TreeScanner;
+import com.sun.tools.javac.api.JavacScope;
+import com.sun.tools.javac.code.Flags;
+import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.comp.AttrContext;
+import com.sun.tools.javac.comp.Enter;
+import com.sun.tools.javac.comp.Env;
+import com.sun.tools.javac.comp.Resolve;
+import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.tree.JCTree.JCClassDecl;
+import com.sun.tools.javac.tree.JCTree.JCMethodDecl;
+import com.sun.tools.javac.util.Context;
+import java.util.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.lang.model.SourceVersion;
+import javax.lang.model.element.*;
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.type.UnionType;
+import javax.lang.model.util.Types;
+import org.netbeans.api.java.lexer.JavaTokenId;
+import org.netbeans.api.java.source.JavaSource.Phase;
+import org.netbeans.api.lexer.TokenSequence;
+
+/**
+ *
+ * @author Jan Lahoda, Dusan Balek, Tomas Zezula
+ */
+public final class TreeUtilities {
+    
+    /**{@link Kind}s that are represented by {@link ClassTree}.
+     * 
+     * @since 0.67
+     */
+    public static final Set<Kind> CLASS_TREE_KINDS = EnumSet.of(Kind.ANNOTATION_TYPE, Kind.CLASS, Kind.ENUM, Kind.INTERFACE);
+    
+    private final CompilationInfo info;
+    
+    /** Creates a new instance of CommentUtilities */
+    public TreeUtilities(final CompilationInfo info) {
+        assert info != null;
+        this.info = info;
+    }
+    
+    /**Checks whether the given tree represents a class.
+     * @deprecated since 0.67, <code>Tree.getKind() == Kind.CLASS</code> should be used instead.
+     */
+    @Deprecated
+    public boolean isClass(ClassTree tree) {
+        return (((JCTree.JCModifiers)tree.getModifiers()).flags & (Flags.INTERFACE | Flags.ENUM | Flags.ANNOTATION)) == 0;
+    }
+    
+    /**Checks whether the given tree represents an interface.
+     * @deprecated since 0.67, <code>Tree.getKind() == Kind.INTERFACE</code> should be used instead.
+     */
+    @Deprecated
+    public boolean isInterface(ClassTree tree) {
+        final long flags = ((JCTree.JCModifiers) tree.getModifiers()).flags;
+        return (flags & Flags.INTERFACE) != 0 && (flags & Flags.ANNOTATION) == 0;
+    }
+    
+    /**Checks whether the given tree represents an enum.
+     * @deprecated since 0.67, <code>Tree.getKind() == Kind.ENUM</code> should be used instead.
+     */
+    @Deprecated
+    public boolean isEnum(ClassTree tree) {
+        return (((JCTree.JCModifiers)tree.getModifiers()).flags & Flags.ENUM) != 0;
+    }
+
+    /**
+     * Checks wheteher given variable tree represents an enum constant.
+     */
+    public boolean isEnumConstant(VariableTree tree) {
+        return (((JCTree.JCModifiers) tree.getModifiers()).flags & Flags.ENUM) != 0;
+    }
+    
+    /**Checks whether the given tree represents an annotation.
+     * @deprecated since 0.67, <code>Tree.getKind() == Kind.ANNOTATION_TYPE</code> should be used instead.
+     */
+    @Deprecated
+    public boolean isAnnotation(ClassTree tree) {
+        return (((JCTree.JCModifiers)tree.getModifiers()).flags & Flags.ANNOTATION) != 0;
+    }
+    
+    /**Returns whether or not the given tree is synthetic - generated by the parser.
+     * Please note that this method does not check trees transitively - a child of a syntetic tree
+     * may be considered non-syntetic.
+     * 
+     * @return true if the given tree is synthetic, false otherwise
+     * @throws NullPointerException if the given tree is null
+     */
+    public boolean isSynthetic(TreePath path) throws NullPointerException {
+        if (path == null)
+            throw new NullPointerException();
+        
+        while (path != null) {
+            if (isSynthetic(path.getCompilationUnit(), path.getLeaf()))
+                return true;
+
+            path = path.getParentPath();
+        }
+        
+        return false;
+    }
+    
+    boolean isSynthetic(CompilationUnitTree cut, Tree leaf) throws NullPointerException {
+        JCTree tree = (JCTree) leaf;
+        
+        if (tree.pos == (-1))
+            return true;
+
+        if (leaf.getKind() == Kind.METHOD) {
+            //check for synthetic constructor:
+            return (((JCMethodDecl)leaf).mods.flags & Flags.GENERATEDCONSTR) != 0L;
+        }
+
+        //check for synthetic superconstructor call:
+        if (leaf.getKind() == Kind.EXPRESSION_STATEMENT) {
+            ExpressionStatementTree est = (ExpressionStatementTree) leaf;
+            
+            if (est.getExpression().getKind() == Kind.METHOD_INVOCATION) {
+                MethodInvocationTree mit = (MethodInvocationTree) est.getExpression();
+                
+                if (mit.getMethodSelect().getKind() == Kind.IDENTIFIER) {
+                    IdentifierTree it = (IdentifierTree) mit.getMethodSelect();
+                    
+                    if ("super".equals(it.getName().toString())) {
+                        SourcePositions sp = info.getTrees().getSourcePositions();
+                        return sp.getEndPosition(cut, leaf) == (-1);
+                    }
+                }
+            }
+        }
+        
+        return false;
+    }
+    
+    public TreePath pathFor(int pos) {
+        return pathFor(new TreePath(info.getCompilationUnit()), pos);
+    }
+
+    /*XXX: dbalek
+     */
+    public TreePath pathFor(TreePath path, int pos) {
+        return pathFor(path, pos, info.getTrees().getSourcePositions());
+    }
+
+    /*XXX: dbalek
+     */
+    @SuppressWarnings("fallthrough")
+    public TreePath pathFor(TreePath path, int pos, SourcePositions sourcePositions) {
+        if (info == null || path == null || sourcePositions == null)
+            throw new IllegalArgumentException();
+        
+        class Result extends Error {
+            private static final long serialVersionUID = 42L;
+            TreePath path;
+            Result(TreePath path) {
+                this.path = path;
+            }
+        }
+        
+        class PathFinder extends TreePathScanner<Void,Void> {
+            private int pos;
+            private SourcePositions sourcePositions;
+            
+            private PathFinder(int pos, SourcePositions sourcePositions) {
+                this.pos = pos;
+                this.sourcePositions = sourcePositions;
+            }
+            
+            public Void scan(Tree tree, Void p) {
+                if (tree != null) {
+                    if (sourcePositions.getStartPosition(getCurrentPath().getCompilationUnit(), tree) < pos && sourcePositions.getEndPosition(getCurrentPath().getCompilationUnit(), tree) >= pos) {
+                        if (tree.getKind() == Tree.Kind.ERRONEOUS) {
+                            tree.accept(this, p);
+                            throw new Result(getCurrentPath());
+                        }
+                        super.scan(tree, p);
+                        throw new Result(new TreePath(getCurrentPath(), tree));
+                    }
+                }
+                return null;
+            }
+
+            @Override
+            public Void visitVariable(VariableTree node, Void p) {
+                int[] span = findNameSpan(node);
+                if (span != null && span[0] <= pos && pos < span[1]) {
+                    throw new Result(getCurrentPath());
+                }
+                return super.visitVariable(node, p);
+            }
+
+            @Override
+            public Void visitMethod(MethodTree node, Void p) {
+                int[] span = findNameSpan(node);
+                
+                if (span != null && span[0] <= pos && pos < span[1]) {
+                    throw new Result(getCurrentPath());
+                }
+                
+                return super.visitMethod(node, p);
+            }
+        }
+        
+        try {
+            new PathFinder(pos, sourcePositions).scan(path, null);
+        } catch (Result result) {
+            path = result.path;
+        }
+        
+        if (path.getLeaf() == path.getCompilationUnit())
+            return path;
+        
+        TokenSequence<JavaTokenId> tokenList = tokensFor(path.getLeaf(), sourcePositions);
+        tokenList.moveEnd();
+        if (tokenList.movePrevious() && tokenList.offset() < pos) {
+            switch (tokenList.token().id()) {
+                case GTGTGT:
+                case GTGT:
+                case GT:
+                    if (path.getLeaf().getKind() == Tree.Kind.MEMBER_SELECT || TreeUtilities.CLASS_TREE_KINDS.contains(path.getLeaf().getKind()) || path.getLeaf().getKind() == Tree.Kind.GREATER_THAN)
+                        break;
+                case RPAREN:
+                    if (path.getLeaf().getKind() == Tree.Kind.ENHANCED_FOR_LOOP || path.getLeaf().getKind() == Tree.Kind.FOR_LOOP ||
+                            path.getLeaf().getKind() == Tree.Kind.IF || path.getLeaf().getKind() == Tree.Kind.WHILE_LOOP ||
+                            path.getLeaf().getKind() == Tree.Kind.DO_WHILE_LOOP || path.getLeaf().getKind() == Tree.Kind.TYPE_CAST)
+                        break;
+                case SEMICOLON:
+                    if (path.getLeaf().getKind() == Tree.Kind.FOR_LOOP &&
+                            tokenList.offset() <= sourcePositions.getStartPosition(path.getCompilationUnit(), ((ForLoopTree)path.getLeaf()).getUpdate().get(0)))
+                        break;
+                case RBRACE:
+                    path = path.getParentPath();
+                    switch (path.getLeaf().getKind()) {
+                        case CATCH:
+                            path = path.getParentPath();
+                        case METHOD:
+                        case FOR_LOOP:
+                        case ENHANCED_FOR_LOOP:
+                        case SYNCHRONIZED:
+                        case WHILE_LOOP:
+                        case TRY:
+                            path = path.getParentPath();
+                            break;
+                        case IF:
+                            do {
+                                path = path.getParentPath();
+                            } while (path != null && path.getLeaf().getKind() == Tree.Kind.IF);
+                            break;
+                    }
+                    break;
+            }
+        }
+        return path;
+    }
+
+    /**Computes {@link Scope} for the given position.
+     */
+    @SuppressWarnings("fallthrough")
+    public Scope scopeFor(int pos) {
+        List<? extends StatementTree> stmts = null;
+        SourcePositions sourcePositions = info.getTrees().getSourcePositions();
+        TreePath path = pathFor(pos);
+        CompilationUnitTree root = path.getCompilationUnit();
+        switch (path.getLeaf().getKind()) {
+            case BLOCK:
+                stmts = ((BlockTree)path.getLeaf()).getStatements();
+                break;
+            case FOR_LOOP:
+                stmts = ((ForLoopTree)path.getLeaf()).getInitializer();
+                break;
+            case ENHANCED_FOR_LOOP:
+                stmts = Collections.singletonList(((EnhancedForLoopTree)path.getLeaf()).getStatement());
+                break;
+            case METHOD:
+                stmts = ((MethodTree)path.getLeaf()).getParameters();
+                break;
+        }
+        if (stmts != null) {
+            Tree tree = null;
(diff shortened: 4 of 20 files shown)


rorygraves:
Awesome!

aemoncannon:
Squashed and rebased.

aemoncannon:
Added support for finding method docs (in addition to reference types).b

core/src/main/java/org/ensime/core/javac/TreeUtilities.java
```diff
@@ -0,0 +1,1077 @@
...
+                            } while (path != null && path.getLeaf().getKind() == Tree.Kind.IF);
+                            break;
+                    }
+                    break;
+            }
+        }
+        return path;
+    }
+    
+//    /**Parses given type in given context.
```
rorygraves: Did you comment this out, or was it already like this?  If you took it out I would add a comment at the top with where the code came from (after the 'DO NOT REMOVE' notice, and then just delete the commented code.

core/src/main/scala/org/ensime/util/FileUtil.scala
```diff
@@ -33,7 +33,11 @@ object RichFileObject {
 object FileUtils {
 
   def isScalaSourceFile(f: File): Boolean = {
-    f.exists && f.getName.endsWith(".scala")
+    f.getName.endsWith(".scala")
```
rorygraves: Is there any reason to remove the f.exists call here?

rorygraves:
Overall this looks good, from the AppVeyor test looks like you broke the BasicWorkFlowTest
One question, does this support mixed mode stuff, I notice we have a scala compiler and a java compiler interface - if a scala file depends on a java file or vice-versa does that work?

aemoncannon:
Added fixes for browsing method overloads.

aemoncannon:
Fixed BasicWorkFlow..

aemoncannon:
@rorygraves 
re: mixed mode

In binary mode, things should just work. 

In source mode, uncompiled changes to java will not be visible for scala and vice versa. The latter is readily fixable by also sending java changes to the scala pres compiler. The latter...  I don't know.

My motivation is to make this work for a Giant Java Project - so I'm going to focus my efforts on binary mode.

edani:
@aemoncannon We used to see problems sending Java sources to the scala pres.compiler (especially involving scala aliases to Java types, such as String). Back then , I asked about it on the scala-ide list and they said "don't do it". My guess is that there isn't a lot of interest in supporting this feature.

core/src/main/scala/org/ensime/util/FileUtil.scala
aemoncannon: IMO it's surprising that this would hit the disk. Probably a (my) mistake in the first place. Only one client of this method and it looks safe to change.

rorygraves: Cool, fair enough

aemoncannon:
What's the difference between drone and appveyor and do we need to wait for both? Drone looks to be crashing jvm maybe..

rorygraves:
Drone does the publishing and the matrix build, I thought appveyor was for a windows build.
We tend to be on the cutting edge of Drone and they made a breaking change, it appears to be good now, but it can't publish because it lost the sonotype keys

fommil:
Bugger. I have the keys. You should actually have them somewhere but the formatting is important. Sorry we'll have to wait until I'm back on 23rd for publishing unless you can extract them from the sqllite file.

+1 for concentrating on binary mode for Scala files seeing Java files. Source mode just doesn't scale at all.

Really looking forward to this one!

I realised after I sent my message about the lucene/h2 database that we probably don't persist everything that you need. But we definitely scan it, so it shouldn't be too hard to add a "hierarchy" table for inheritance info and a "members" table for method completion. Argh, I'd actually really like to write that but I don't have my laptop.

fommil:
Sucks that netbeans is designed that way. At least we can rip out the bits we need and ultimately that'll still save us a load of dev time. It'll come into its own for refactoring I reckon.

aemoncannon:
I know it should really be in another PR, but couldn't resist adding first cut of completion. Just members now, and only works in simple contexts, but it's nice and responsive. Woo!

fommil:
I'll definitely have a good look on the 23rd if you can wait that long before merging.

fommil:
Is there a way to get Java formatting working as part of the SBT build? I think we might have to use the eclipse formatter for that unless there is a nice netbeans one. Might be worth putting our Java formatting logic (when we have one) in a shared module for that very reason.

aemoncannon:
More completion stuff, with tests.

aemoncannon:
Pushed some Analyzer changes that allow the actor to continue living if scala compiler is not created (in this case, when it explodes on construction due to scala lib missing from classpath)

aemoncannon:
So far I'm very encouraged by performance on my big project. The JavaCompiler is stupid simple -- it just forks a new javac for every request -- but that might be good enough!

fommil:
I'm setting myself the goal of indexing the type/member hierarchy and reverse lookups by Sunday. That should open up some options for further features.

aemoncannon:
Cool. Not sure if you saw my previous message, but don't need the member lookups (at least not for java completion). The javac model is enough.

aemoncannon:
Btw, we probably explored this already, but what about some kind of hybrid source/binary mode, where we generate class files for edited source files and stash them into a temp dir that we preppend to the classpath?

fommil:
For Scala or Java? I thought we'd just rely on the build tool to do that. It's there a complication in Java?

edani:
I don't know about Java, but if Scala is involved, this trick won't work for transitive dependencies (e.g, we edit "A.java" ; "B.scala" depends on "A.java" but isn't loaded in the editor ;  "C.scala" depends on "B.scala" and is loaded in the editor). In that case, I don't know of any solution other than recompiling with the build tool.

aemoncannon:
Ah. I see. We just rely on sbt to do its thing. Sadly this won't help me on my java project because the builtdtool is dumb and CPU intense.

fommil:
Ah, well for Java you could hack it and invoke javac from Emacs I think (basically write a minimal build tool in elisp) and then invoke it on buffer save.

You might need a richer project description in the .ensime file though. I certainly would not be against updating the format to be a truer representation of the build.

fommil:
Re Scala Lib I think we might have a special mode for that. It's needed to hack on Scala itself. I'll have a look when I review.

aemoncannon:
I'm seeing a hang on startup that seems to be caused by spinning up lots and lots of FileMonitors. Makes sense as I'm throwing 700 source roots and 300 target directories at it. Is there a way I can make the monitoring more scalable? what if we identified the common parent of all those source roots?

fommil:
Uuugh. I can imagine. We have an evil hack in that area to work around a bug in Apache vfs. I think it needs an upstream fix, and if we move to Java 7+ we could write a backend using NIO2. I'm not sure this is fixable in is current state. How many source directories are we talking about, and which OS? I bet this will affect me as well.

fommil:
It probably wouldn't be too hard to fix this upstream. I've just tried to avoid Java since returning to Emacs. I've heard we have a fix for that :smile:

fommil:
I missed the sizes. Yeah, maybe monitoring the common parent might work, but if there is anything else in those directories it'll scan them too and possibly be slower (think .git). But maybe we could workaround that too. I'd like to just fix it upstream first before going too far down the hacking route. I have 2, possibly, 3, open tickets there. If we could use nio2 in Java 7 that'd make the biggest difference and I think somebody had a half working solution.

fommil:
One other thing we could do to fix this (possibly upstream) is to have a way of merging listeners. That way there should only be one thread looking through all dirs.

aemoncannon:
Actually it looks like that's already what we're doing (adding all dirs to a single watcher, with one thread). Maybe it's just that one thread is doing a ton of work now, and that's why it's showing up as hotspot.

fommil:
Is this on windows by any chance?

aemoncannon:
Linux

api/src/main/scala/org/ensime/api/config.scala
```diff
@@ -25,6 +25,8 @@ case class EnsimeConfig(
   def root = rootDir
   def debugVMArgs = debugArgs
   def referenceSourceJars = referenceSourceRoots
+  def disableSourceMonitoring = true
```
fommil: it's a shame we have to do this

api/src/main/scala/org/ensime/api/outgoing.scala
```diff
@@ -177,6 +187,8 @@ case class SourceFileInfo(
 ) {
   // keep the log file sane for unsaved files
   override def toString = s"SourceFileInfo($file,${contents.map(_ => "...")},$contentsIn)"
+  def isJava = file.getName().endsWith(".java")
```
fommil: I'd prefer if we moved any code from the API module into somewhere else (the typeclass approach vs OOP) as it keeps the API schema definition really clean. (it's not possible to use typeclasses for toString, equals, hashCode unfortunately)

core/src/main/java/org/ensime/core/javac/CompilationInfo.java
```diff
@@ -0,0 +1,77 @@
...
+                                  new InputAttributes());
+        this.compilationUnit = compilationUnit;
+    }
+
+    public CompilationInfo(final JavacTask javacTask,
+                           final Element element) throws IOException {
+        this(javacTask, (CompilationUnitTree)(Trees.instance(javacTask).getTree(element.getEnclosingElement())));
+    }
+
+    public JavacTask getJavacTask() { return javacTask; }
```
fommil: what are your thoughts on using https://projectlombok.org/ to autogenerate all this boilerplate?

core/src/main/java/org/ensime/core/javac/TreeUtilities.java
```diff
@@ -0,0 +1,853 @@
...
+ * "[Contributor] elects to include this software in this distribution
+ * under the [CDDL or GPL Version 2] license." If you do not indicate a
+ * single choice of license, a recipient has the option to distribute
+ * your version of this file under either the CDDL, the GPL Version 2 or
+ * to extend the choice of license to its licensees as provided above.
+ * However, if you add GPL Version 2 code and therefore, elected the GPL
+ * Version 2 license, then the option applies only if the new code is
+ * made subject to such option by the copyright holder.
+ */
+package org.ensime.core.javac;
```
fommil: perhaps we should create a `netbeans` module for everything that we copy from them.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -51,7 +54,9 @@ class Analyzer(
   private var settings: Settings = _
   private var reporter: PresentationReporter = _
 
-  protected var scalaCompiler: RichCompilerControl = _
+  protected var scalaCompiler: Option[RichCompilerControl] = _
```
fommil: yeah, this is a bit of a hack. We shouldn't need to do this, because the scala compiler should be on ENSIME's path not necessarily the project's. I'll try to send you a fix for this today.

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -0,0 +1,178 @@
...
+
+  import scala.concurrent.ExecutionContext.Implicits.{ global => exe }
+
+  val compiler = ToolProvider.getSystemJavaCompiler()
+  private val listener = new JavaDiagnosticListener()
+  private val silencer = new SilencedDiagnosticListener()
+  val fileManager = compiler.getStandardFileManager(listener, null, charset)
+  val cp = config.compileClasspath.mkString(File.pathSeparator)
+
+  def askTypecheckFiles(files: Iterable[SourceFileInfo]): Future[Boolean] = Future {
```
fommil: this is probably better as a `Future[Unit]`

fommil: actually, I'm not sure it needs to be a `Future` at all

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -0,0 +1,178 @@
...
+    reportHandler.clearAllJavaNotes()
+    typecheckAll(files.map(getJavaFileObject))
+    true
+  }
+
+  def askTypeAtPoint(file: SourceFileInfo, offset: Int): Future[Option[TypeInfo]] = Future {
```
fommil: similarly for all the other methods. Do they really need to be `Future`s? It would be better if they were just normal methods and if we need to call them asynchronously, we'll do it at invocation rather than baked into the API/impl.

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  private def typecheckAll(files: Iterable[JavaFileObject]): Unit = {
+    val task = compiler.getTask(null, fileManager, listener, List(
+      "-cp", cp, "-Xlint:all", "-proc:none"
+    ), null, asJavaIterable(files)).asInstanceOf[JavacTask]
+    val t = System.currentTimeMillis()
+    task.parse()
+    task.analyze()
+    System.err.println("Parsed and analyzed: " + (System.currentTimeMillis() - t) + "ms")
```
fommil: looks like some println debugging has been left in... move to SLF4J or remove?

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  private def typecheckForUnits(files: Iterable[JavaFileObject]): Vector[CompilationInfo] = {
+    val task = compiler.getTask(null, fileManager, silencer, List(
+      "-cp", cp, "-Xlint:none", "-proc:none"
+    ), null, asJavaIterable(files)).asInstanceOf[JavacTask]
+    val t = System.currentTimeMillis()
+    val units = task.parse().map(new CompilationInfo(task, _)).toVector
+    task.analyze()
+    System.err.println("Parsed and analyzed for trees: " + (System.currentTimeMillis() - t) + "ms")
```
fommil: again

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -0,0 +1,178 @@
...
+    task.parse()
+    task.analyze()
+    System.err.println("Parsed and analyzed: " + (System.currentTimeMillis() - t) + "ms")
+  }
+
+  private def typecheckForUnits(files: Iterable[JavaFileObject]): Vector[CompilationInfo] = {
+    val task = compiler.getTask(null, fileManager, silencer, List(
+      "-cp", cp, "-Xlint:none", "-proc:none"
+    ), null, asJavaIterable(files)).asInstanceOf[JavacTask]
+    val t = System.currentTimeMillis()
```
fommil: btw, we should have Guava on the classpath, so the `Stopwatch` is available for this sort of thing. It prints the timings out really cleanly

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -0,0 +1,178 @@
...
+
+  private class JavaObjectWithContents(val f: File, val contents: String)
+      extends SimpleJavaFileObject(f.toURI, JavaFileObject.Kind.SOURCE) {
+    override def getCharContent(ignoreEncodingErrors: Boolean): CharSequence = contents
+  }
+
+  private class JavaObjectFromFile(val f: File)
+      extends SimpleJavaFileObject(f.toURI, JavaFileObject.Kind.SOURCE) {
+    override def getCharContent(ignoreEncodingErrors: Boolean): CharSequence = {
+      FileUtils.readFile(f, charset) match {
```
fommil: I think we have a cleaner util for this now by importing the `org.ensime.util.file._` syntax. I'm not a big fan of the `Either[Failure, Success]` approach, especially when its just wrapping an exception anyway (there is merit if errors are common and can be detected without an exception because they can be a lot faster).

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -0,0 +1,178 @@
...
+    }
+  }
+
+  private class JavaDiagnosticListener extends DiagnosticListener[Object] with ReportHandler {
+    def report(diag: Diagnostic[_ <: Object]): Unit = {
+      reportHandler.reportJavaNotes(List(
+        Note(
+          diag.getSource() match {
+            case jfo: JavaFileObject => jfo.getName()
+            case _ => "NA"
```
fommil: would `Unknown` perhaps be better, or is this exceedingly rare?

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
```diff
@@ -0,0 +1,239 @@
...
+trait JavaCompletion { self: JavaCompiler =>
+
+  import CompletionUtil._
+
+  def completionsAt(info: SourceFileInfo, offset: Int, maxResultsArg: Int, caseSens: Boolean): CompletionInfoList = {
+    val s = contentsAsString(info, charset)
+
+    val preceding = s.slice(Math.max(0, offset - 100), offset)
+
+    System.err.println("PRECEDING: " + preceding)
```
fommil: more :smile:

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
```diff
@@ -0,0 +1,239 @@
...
+    } else if (ImportRegexp.findFirstMatchIn(preceding).isDefined) {
+      (pathToPoint(info, indexAfterTarget) flatMap {
+        case (info: CompilationInfo, path: TreePath) => {
+          getEnclosingMemberSelectTree(path).map { m =>
+            packageMemberCandidates(info, m, defaultPrefix, caseSens)
+          }
+        }
+      })
+    } else if (isMemberAccess) {
+      // TODO how to avoid allocating a new string? buffer of immutable string slices?
```
fommil: unless we can show this is a performance bottleneck, I wouldn't give it another thought

fommil: btw, don't forget the `.substring` memory leak http://stackoverflow.com/questions/15612157 (I think it's fixed Java 7+)

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
```diff
@@ -0,0 +1,239 @@
...
+        }
+      })
+    }).getOrElse(CompletionInfoList(defaultPrefix, List()))
+  }
+
+  private def getEnclosingMemberSelectTree(path: TreePath): Option[MemberSelectTree] = {
+    var p = path
+    while (p != null) {
+      p.getLeaf match {
+        case m: MemberSelectTree => return Some(m)
```
fommil: it would be good to rewrite this without the `return` statement. It has far too many weird behaviours for me to be comfortable with it.

core/src/main/scala/org/ensime/core/javac/JavaDocFinding.scala
```diff
@@ -0,0 +1,129 @@
...
+  // For each (argument, parameter) pair compute an integer distance between them.
+  // Int.MaxValue denotes an incompatible pairing (e.g. reference argument supplied for
+  // primitive param). 0 denotes type equality. 
+  private def typeDeltas(info: CompilationInfo, m: MethodInvocationTree, c: ExecutableElement): (ExecutableElement, Array[Int]) = {
+    val types = info.getTypes()
+    val args = m.getArguments
+    val params = c.getParameters
+    var i = 0
+    val result = new Array[Int](args.length)
+    while (i < args.length) {
```
fommil: I don't think we need a mutable `i` here, `for { i <- 0 to args.length}` should suffice. I assume `Array` is necessary to meet a Java API.

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -114,18 +114,20 @@ class ClassfileWatcher(
...
-      if ancestor.getName isAncestor root.getName
-      _ = workaround.removeFile(ancestor)
-      _ = workaround.addFile(ancestor)
-    } {
-      // side effects in the for comprehension
+    if (!config.disableClassMonitoring) {
+      // must remove then add to avoid leaks
+      for {
+        d <- config.targetClasspath
+        dir = vfs.vfile(d)
```
fommil: it would be more canonical to put the `if` statement in the for comprehension instead of outside it (will also result in a much smaller diff)

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -165,8 +167,10 @@ class SourceWatcher(
   fm.setRecursive(true)
   fm.start()
 
-  config.modules.values.foreach { m =>
```
fommil: this should probably have been a for comprehension to start with

core/src/main/scala/org/ensime/util/FileUtil.scala
fommil: aah, ok, this functionality is duplicated with the API.

project/EnsimeBuild.scala
```diff
@@ -58,6 +58,7 @@ object EnsimeBuild extends Build with JdkResolver {
 
   lazy val commonSettings = scalariformSettings ++ basicSettings ++ Seq(
     //resolvers += Resolver.sonatypeRepo("snapshots"),
+    resolvers += "Spring IO" at "http://repo.spring.io/libs-release-remote",
```
fommil: ouch, this is going to need to be added to the `build.sbt` in emacs, vim, etc etc.

fommil:
Looks great! I'll make some changes today and send to you in a PR for you to squash

fommil:
oh noes!

```
[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn] 	* org.apache.lucene:lucene-core:3.5.0 -> 4.7.2
```

that only works for Java 7+. Can we get away with using intransitive netbeans dependencies to try to limit the footprint? I'm sure we're not actually making use of their lucene indexing.

Or maybe we need to drop back to an older version of Netbeans.

fommil:
Or we could make the decision now that we only officially support Java 7+

I'm uncomfortable doing that without getting feedback from users, because Java 6 is still the official JDK for Scala 2.11 and I know a lot of corporates are still using Java 6.

fommil:
oh, hold on, they want an old version of lucene and we need the latest Java 6 one. Chances are their calls to lucene will fail if they try to make them, but I think we're ok. **loads** of jars from netbeans, I guess that's not surprising.

fommil:
`StandardCharsets` was introduced in Java 7, https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html but we're still supporting Java 6.

BTW, if you want to make SBT / ENSIME use a specific JDK, just create one of these

```
cat .sbtopts 
-java-home
/usr/lib/jvm/java-6-oracle

```

(must end with a newline)

fommil:
a good alternative is `com.google.common.base.Charsets`

fommil:
or our `org.ensime.util.file.DefaultCharset`

fommil:
hmm, something is using java 7... might have to downgrade netbeans

```
[warn] /home/fommil/Projects/ensime-server/core/src/main/java/org/ensime/core/javac/CompilationInfo.java:  major version 51 is newer than 50, the highest major version supported by this compiler.
```

fommil:
I've checked the netbeans jars, 7.3.1 targets JDK6 but 7.4+ targets JDK7. I'll try to roll back the version we're depending on.

fommil:
argh, TreeUtilities is using a lot of the JDK7 API.

api/src/main/scala/org/ensime/api/config.scala
aemoncannon: Btw I'm thinking undocumented flags for use in the short term for very big projects.
It's a shame, but class monitoring is useless in my situation anyway, because I don't have continuous build. Source monitoring less so, but I can live without it.

Obviously looking forward to getting rid of these : )

api/src/main/scala/org/ensime/api/outgoing.scala
aemoncannon: Fair. I'll move this.

core/src/main/java/org/ensime/core/javac/CompilationInfo.java
aemoncannon: Sounds cool. Wait for 3 or 4 more repetitions?

core/src/main/scala/org/ensime/core/Analyzer.scala
aemoncannon: That would be nice. I seem to recall scala compiler *requiring* scala lib on project's path, but I hope I'm wrong : )

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
aemoncannon: done

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
aemoncannon: done

fommil:
@aemoncannon do you have a diff of the TreeUtilities from the original? I'm trying to go back to Netbeans 7.3.1 for Java 6 to continue working. We can upgrade to Java 7+ later if we all agree, but it will require checking with our users and also updating the build image (which is geared towards Java 6 and I'd have to change the base image to a more recent version of debian). I'm really keen that we stay on Java 6 until we cut 1.0.

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
aemoncannon: was trying to express, with types, the fact that the public api is non-blocking. this is a departure from scala presentation compiler which blocks awaiting serial processing of the requests. still think i should change?

fommil:
nevermind, found it

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
rorygraves: I'm out and about right now but there is a whole other discussion to be had here about compiler interfaces. Looking a t raising the level of the interface so that we can provide support for dotty

Cheers

Rory

fommil:
hmm, I'm not sure your `TreeUtilities.java` is based on the version 8 source

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
aemoncannon: done.

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
aemoncannon: revised.

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
aemoncannon: done

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
aemoncannon: ack.

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
aemoncannon: example?

core/src/main/scala/org/ensime/core/javac/JavaDocFinding.scala
aemoncannon: done

core/src/main/scala/org/ensime/core/Analyzer.scala
fommil: yeah, I remember fixing that so it would work for scala compiler / library hackers. I hope I didn't break it again.

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
aemoncannon: hrm. it feels more clear/readable the way it is.

project/EnsimeBuild.scala
aemoncannon: wha? why would they duplicate ensime build config?

fommil: they don't, they use `sbt` to download ensime-server (and all its dependencies). If we have dependencies that aren't on maven central, that repo needs to be added to the bootstrapper

fommil:
wow! I didn't expect you to actually make any changes, I was going to do that. But as it turns out, I've been fighting the Java 6 / 7 thing anyway so I've not gotten very far. btw, @aemoncannon do you have access to gitter?

aemoncannon:
I pushed several fixes for your comments. PTAL. I didnt touch the charset or java 7vs6 debacle.

aemoncannon:
p.s. Thanks for the review

fommil:
do you have an unsquashed version of the changes? :-)

fommil:
also, do you remember which version of TreeUtilities you forked from?

aemoncannon:
I can probably dig it up - just found it by googling...

fommil:
awesome, that'd really help me to create the version based off of netbeans 7.3.1

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
fommil: this is the first thing that comes up, but not sure it does a great job of motivating why **not** to use it https://tpolecat.github.io/2014/05/09/return.html

The problem I've seen in the wild is when somebody `return`s from a closure and it actually returns from the closure (whereas a natural reading would expect it to return the whole method) and continues through. Given that it's not always obvious when you're typing in a closure, or the method body, it results in all manner of weird stuff happening.

fommil:
If I had to guess, I'd say you used the one from version 7.1 and removed everything you didn't need http://grepcode.com/file_/bits.netbeans.org/maven2/org.netbeans.api/org-netbeans-modules-java-source/RELEASE71/org/netbeans/api/java/source/TreeUtilities.java/?v=source

fommil:
aah, which allowed you to create the trivial impl of CompilationInfo. OK, now I get it.

aemoncannon:
:-)  I'm not proud of it, but the tests pass!

fommil:
ok, I've rebased to 7.3.1. Next I'll try to get it to work on Java 6. Might mean a loss of some functionality. NetBeans must compile under Java 7 with target of Java 6 so I'm guessing Netbeans 7.3 would probably blow up if it was actually run under JDK6.

fommil:
@aemoncannon you forgot to check the integration tests :smile: I'll fix as its all part of the scala analyzer for pure java project thing.

fommil:
fyi, this may be relevant from here on http://stackoverflow.com/questions/3656645

fommil:
@aemoncannon you might want to run `git config user.email YOUR_GMAIL_ADDRESS` in your ensime-server repo :wink: (been there)

fommil:
I've got it compiling for Java 6! https://github.com/fommil/ensime-server/commits/java-type-checking I'll raise a PR once I have done the things I said I'd do (scala-library, etc)

fommil:
excellent, despite all the extra netbeans jars, the assembled jar doesn't change much in size!

fommil:
damn, my JDK6 fix doesn't work when run for a JDK7 project. grr... back to the drawing board.

fommil:
btw, Netbeans say these are the offical nexus jars https://twitter.com/netbeans/status/657632494565634048

aemoncannon:
Neat!

fommil:
nice, they have a more complete set of jars, so we could even go back to 6.9.1 from here :smile:

fommil:
but I am not 100% sure that those old jars are compatible with Java 7 or Java 8 at runtime.

fommil:
hmm, interestingly enough, some versions of netbeans are also on bintray. e.g.  https://jcenter.bintray.com/org/netbeans

fommil:
hmm, binary change detection seems to be completely disabled now.

fommil:
interesting, I just got this error when running on a Java 7 project with my Java 6 compiled ensime-server.

It would suggest that the internal javac API has changed (which isn't too surprising) but I checked and my ensime-server branch does compile under Java 7. Perhaps the API changed in a subtle way, e.g. one version takes `String`, another takes `CharSequence`.

```
com.sun.tools.javac.util.Name$Table.fromString(Ljava/lang/CharSequence;)Lcom/sun/tools/javac/util/
```

```
21:37:13.268 DEBUG akka://ENSIME/user/ensime-main/project/analyzer o.e.c.Analyzer - received handled message TypeAtPointReq(/home/fommil/Projects/matrix-toolkits-java/src/main/java/no/uib/cipr/matrix/io/MatrixVectorReader.java,OffsetRange(16340,16340)) in state ready
21:37:13.437 INFO  None o.e.c.j.JavaCompiler - Parsed and analyzed for trees: 168ms
java.lang.NoSuchMethodError: com.sun.tools.javac.util.Name$Table.fromString(Ljava/lang/CharSequence;)Lcom/sun/tools/javac/util/Name;
	at org.ensime.core.javac.TreeUtilities.findNameSpan(TreeUtilities.java:391)
	at org.ensime.core.javac.TreeUtilities$1PathFinder.visitMethod(TreeUtilities.java:220)
```

fommil:
and when I do it with the correct version of Java, I get

```
21:48:16.005 DEBUG akka://ENSIME/user/ensime-main/tcp-server/con1/8 a.a.LocalActorRef - unhandled message from Actor[akka://ENSIME/user/ensime-main/project/analyzer#-700367121]: Failure(java.lang.IllegalStateException: Compilation in progress)
```

ok, new approach. I'll stop testing on my own projects and go back to good old fashioned unit / integration tests tomorrow. Bed now (sorry for the spam).

fommil:
@aemoncannon are you able to use https://gitter.im/ensime/ensime-server ? (it uses github to log in, no passwords needed). There is a really very interesting conversation in there that I think you might like to know about :smile:

api/src/main/scala/org/ensime/api/config.scala
fommil: I'm going to make these `Option`. I missed that they were hardwired. That makes the for comprehension approach suggested earlier make more sense (although admittedly its a bit weird to have an `Option[Boolean]` that would never be `false`).

For a while now I've wanted to create a user-wide config file, which might work well here. But I think this is still project specific.

core/src/main/scala/org/ensime/core/Refactoring.scala
```diff
@@ -113,9 +113,7 @@ trait RefactoringHandler { self: Analyzer =>
   }
 
   def handleFormatFile(fileInfo: SourceFileInfo): String = {
-    val sourceFile = createSourceFile(fileInfo)
-    val contents = sourceFile.content.mkString
-    ScalaFormatter.format(contents, config.formattingPrefs)
+    ScalaFormatter.format(fileInfo.file.readString, config.formattingPrefs)
```
fommil: this breaks the test because the file might not exist on disk

fommil: I'm on it, fixed in my branch.

fommil:
currently creating a test project and integration test for when there is no scala-library on the classpath.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -174,90 +223,130 @@ class Analyzer(
...
-      sender ! handleRefactorPrepareRequest(req)
+
+      sender ! scalaCompiler.map(handleRefactorPrepareRequest(_, req)).getOrElse(VoidResponse)
     case req: ExecRefactorReq =>
-      sender ! handleRefactorExec(req)
+      sender ! scalaCompiler.map(handleRefactorExec(_, req)).getOrElse(VoidResponse)
     case req: CancelRefactorReq =>
       sender ! handleRefactorCancel(req)
+    case CompletionsReq(f, point, maxResults, caseSens, reload) if FileUtils.isJava(f.file) =>
+      javaCompiler.askCompletionsAtPoint(f, point, maxResults, caseSens) pipeTo sender
```
fommil: these uses of `pipeTo` actually end up returning the wrong type. They forward the `Option` whereas we have been returning the value or `FalseResponse` until now. I wonder if perhaps we should just leave `Analyzer` alone and then separate out the Java requests in `Project` and send to a Java actor. We'd have to be careful when a request has multiple files in the parameter, which are mixed language, but that is probably never valid anyway.

fommil: (I've pushed a broken test to my branch to show this)

fommil:
ok, I'm now making some biggish changes to move all the Java Compiler related stuff into a separate Java Actor AND make sure the Scala Actor doesn't bail when we don't have scala-library on the classpath. It should be ok to merge my branch (I'll not force push to it) but let's just make sure we don't step on each others toes around that area.

aemoncannon:
Cool. Scala actor java actor siblings beneath analyzer parent?

fommil:
@aemoncannon nope, even simpler... JavaAnalyzer as a peer of (Scala)Analyzer

fommil:
cool, I've done the refactor and I have failing tests on the pure java project so now I can look into it in detail. Pushing (broken) code to my branch now if you're curious.

fommil:
ok, pushed a fix for pure java projects (and "Odersky mode" scala projects)

aemoncannon:
Radical

fommil:
@aemoncannon feel free to merge (and even squash) my changes next time you're hacking on this. If you squash, just let me know which commit you got up to.

fommil:
I'm looking at the Java 6 / 7 debacle next. My instinct tells me that the internal JDK API used by Netbeans has binary changes between major (possibly minor) Java releases. That means we're probably going to be restricted to one version of Java. I'm seriously tempted to suggest that we just drop Java 6 support with this PR... but to feel comfortable I'd really want to get 1.0 out the door first so that people on Java 6 can have a fallback version.

Which version of Java do you use at work?

fommil:
ok, this is really weird. I'm using Java 7 to compile ensime and I'm using Java 7 to run ENSIME on itself. If I go into the `CompilationInfo.java` file and enable `ensime-mode` (should this maybe be `enjime-mode`?) then do a completion on `this.` at the end of the first constructor I get the same exception I was seeing earlier:

```
17:41:14.288 INFO  akka://ENSIME/user/ensime-main/tcp-server/con1/32 o.e.c.j.JavaCompiler - Parsed and analyzed for trees: 86ms
java.lang.NoSuchMethodError: com.sun.tools.javac.util.Name$Table.fromString(Ljava/lang/CharSequence;)Lcom/sun/tools/javac/util/Name;
	at org.ensime.core.javac.TreeUtilities.findNameSpan(TreeUtilities.java:391)
	at org.ensime.core.javac.TreeUtilities$1PathFinder.visitMethod(TreeUtilities.java:220)
	at org.ensime.core.javac.TreeUtilities$1PathFinder.visitMethod(TreeUtilities.java:184)
	at com.sun.tools.javac.tree.JCTree$JCMethodDecl.accept(JCTree.java:690)
	at com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:67)
	at org.ensime.core.javac.TreeUtilities$1PathFinder.scan(TreeUtilities.java:200)
	at org.ensime.core.javac.TreeUtilities$1PathFinder.scan(TreeUtilities.java:184)
	at com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:80)
	at com.sun.source.util.TreeScanner.scan(TreeScanner.java:90)
	at com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:98)
	at com.sun.source.util.TreeScanner.visitClass(TreeScanner.java:132)
	at com.sun.tools.javac.tree.JCTree$JCClassDecl.accept(JCTree.java:618)
	at com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:67)
	at org.ensime.core.javac.TreeUtilities$1PathFinder.scan(TreeUtilities.java:200)
	at org.ensime.core.javac.TreeUtilities$1PathFinder.scan(TreeUtilities.java:184)
	at com.sun.source.util.TreeScanner.scan(TreeScanner.java:90)
	at com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:98)
	at com.sun.source.util.TreeScanner.visitCompilationUnit(TreeScanner.java:119)
	at com.sun.tools.javac.tree.JCTree$JCCompilationUnit.accept(JCTree.java:492)
	at com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:49)
	at org.ensime.core.javac.TreeUtilities.pathFor(TreeUtilities.java:231)
	at org.ensime.core.javac.TreeUtilities.pathFor(TreeUtilities.java:166)
	at org.ensime.core.javac.TreeUtilities.pathFor(TreeUtilities.java:160)
	at org.ensime.core.javac.JavaCompiler$$anonfun$pathToPoint$1.apply(JavaCompiler.scala:73)
	at org.ensime.core.javac.JavaCompiler$$anonfun$pathToPoint$1.apply(JavaCompiler.scala:72)
	at scala.Option.flatMap(Option.scala:171)
	at org.ensime.core.javac.JavaCompiler.pathToPoint(JavaCompiler.scala:72)
	at org.ensime.core.javac.JavaCompletion$class.completionsAt(JavaCompletion.scala:65)
	at org.ensime.core.javac.JavaCompiler.completionsAt(JavaCompiler.scala:29)
	at org.ensime.core.javac.JavaCompiler$$anonfun$askCompletionsAtPoint$1.apply(JavaCompiler.scala:67)
	at org.ensime.core.javac.JavaCompiler$$anonfun$askCompletionsAtPoint$1.apply(JavaCompiler.scala:67)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
	at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

fommil:
(ignore my last deleted message if you got it by email, I'm an idiot that's all)

fommil:
wow, I think some of the JDK java API source is not available for the older version of OpenJDK (e.g. try to find `com.sun.tools.javac.util.Table` in the openjdk6 source)

fommil:
oh, no, I'm still an idiot. Inner class...doh!

fommil:
ha! There's the :monkey: 

Java 6, `Table`

```
        public Name fromString(CharSequence s) {
            return Name.fromString(this, s);
        }
```

Java 7

```
        public Name fromString(String s) {
            char[] cs = s.toCharArray();
            return fromChars(cs, 0, cs.length);
        }
```

looks like the hack is to call the `fromUtf8` method which is available in both versions. I strongly suspect we're going to end up being able to compile this with Java 6 but it'll still fail at runtime unless the project uses Java 7.

aemoncannon:
I'm OK with that as long as we die with an informative message.

fommil:
@aemoncannon which version of Java do you need for work? We should probably target that exact version for stability.

aemoncannon:
7

fommil:
interesting, my branch now compiles under JDK6 and JDK7 and passes the tests under JDK7. But when run under JDK6 I get failures in the JavaCompiler tests like this:

```
[info] - should find type at point *** FAILED ***
[info]   java.lang.IllegalStateException: Compilation in progress
[info]   at com.sun.tools.javac.api.JavacTool.beginContext(JavacTool.java:148)
[info]   at com.sun.tools.javac.api.JavacTaskImpl.beginContext(JavacTaskImpl.java:172)
[info]   at com.sun.tools.javac.api.JavacTaskImpl.prepareCompiler(JavacTaskImpl.java:143)
[info]   at com.sun.tools.javac.api.JavacTaskImpl.parse(JavacTaskImpl.java:226)
[info]   at org.ensime.core.javac.JavaCompiler.typecheckForUnits(JavaCompiler.scala:108)
```

there must be something subtle about the way the compiler works in JDK6. I recall javac 7 got a lot faster, so maybe we just need to block (as a hack) or correctly transition the actor state between "ready" and "compiling".

aemoncannon:
Oh maybe it's because I'm reusing a compiler tool?

fommil:
that was it! Passing now in JDK6. We need to make sure that the clients can download all the jars before we can merge this.

fommil:
I'm closing this one as I've raised another PR with all your changes and my ammendments are merged into one. I'm hoping that one is going to go green and then I'll merge and there will be much :tada:

PR #1110, opened 2015-10-04
Remove ".package" when indexing symbol names, for good

### LINKED ISSUE #959: Package object members not imported correctly
When I do `C-c C-r t` and select a completion that is inside a package object, it gets imported as

    import com.example.foo.package.Bar

instead of

    import com.example.foo.Bar

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13450723-package-object-members-not-imported-correctly?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F239449&utm_medium=issues&utm_source=github).
</bountysource-plugin>

edani:
Fir #959
- Remove ".package" from type-alias symbols
- Fix symbol position lookup

### DIFF
--- a/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
@@ -405,9 +405,16 @@ class RichPresentationCompilerSpec extends WordSpec with Matchers
         "}",
         "trait  /*3*/C { ",
         "   val /*3.1*/z: Int = 1",
-        "   class /*3.2*/Z {} ",
+        "   class /*3.2*/Z {}",
         "}",
-        "class /*4*/D extends C { }"
+        "class /*4*/D extends C { }",
+        "package object /*5.0*/pkg {",
+        "   type /*5.1*/A  = java.lang.Error",
+        "   def /*5.2*/B = 1",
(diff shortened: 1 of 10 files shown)


rorygraves:
@edani  Changes look good, can you do me a favour and rebase (its not actually needed) and repush - I'm trying to check that Drone is now correctly building changes again.

rorygraves:
@edani The 2.10 failure looks real
(from the drone build)

edani:
Ah yes, it's a failure in the new test I wrote. I'll take a look.

fommil:
Cool. We should probably be indexing the Scala name as well as the raw Java name, but I suspect the effort to benefit ratio is pretty high.

edani:
The last appveyor failure looks unrelated to my chnage

rorygraves:
@edani Agreed, it looks like ScalaCheck has found an edge case in the sexp parser.
I will merge this later today, once I've had a stab at sorting out the drone publishing keys.

rorygraves:
@edani Sorry, meant to pull this earlier
Won't be deployed until @fommil  returns as drone was rebuilt and I don't have the deployment keys.

Issue #425, opened 2014-06-23
find usages (reverse lookups)


fommil:
there is no real reason why we can't use lucene to do the "who references me" style questions... no need to re-type check anything. Every time a file is edited, we rebuild the index for that file.

fommil:
this might be dropped from 1.0 unless it turns out to be easy to implement

tpolecat:
(Adding myself to this since issue since I'm looking into the db stuff.)

fommil:
w00t! I should imagine it will dramatically increase the DB size so any thoughts on optimising the storage would be well received (although I suspect having a good filter on which symbols we care about... i.e. not `java.lang.*` or any anonymous classes) would be good enough.

FYI when we call this method https://github.com/ensime/ensime-server/blob/master/core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala#L17 we get handed the FQNs of everything in that file.

cvogt:
+1 :)

fommil:
@cvogt you should know that @tpolecat is putting together a patch to move us to Doobie and he has threatened to look into this :smile: It's also on my personal hitlist soonish.

cvogt:
@fommil @tpolecat no problem. I like what @tpolecat is doing. I am trying to excite people about type-safe composable queries :), but I also like doobie's free monad. Seems a bite more elaborate than slick's.

tpolecat:
I should have more time to work on this next month.

fommil:
I don't really like the scalaz functional style because I find it confuses more than helps, plus is a massive source of object creation / GC overhead. However, in this instance, the benefit is moving away from Slick (which I've now labeled as dirty until FreeSlick is ready) and getting back to raw SQL queries, which were causing compiler warnings in Slick. Hopefully @tpolecat can show us why free monads are useful, but I suspect we'll want all that to be contained to the DAO as implementation details.

fommil:
although it is pretty awesome that we now have two of the Scala community's top DB contributors able to help on this :grinning: so whatever happens I'm sure we'll have a rock solid implementation.

Is there any chance of an NIO implementation of the JDBC driver for H2?

t3chnoboy:
Was it implemented in https://github.com/ensime/ensime-server/pull/1602 ?

fommil:
@t3chnoboy no, that's something else, but the graph branch now stores the necessary information.

olkinn:
is this still WIP?

[...]

PR #1119, opened 2015-10-24
Upgrade to Slick 3.1 and non-blocking Database IO

cvogt:
will add another commit or two looking to migrate to non-blocking and refactorings

### DIFF
--- a/core/src/it/scala/org/ensime/fixture/SearchServiceFixture.scala
+++ b/core/src/it/scala/org/ensime/fixture/SearchServiceFixture.scala
@@ -3,6 +3,7 @@ package org.ensime.fixture
 import akka.actor.ActorSystem
 import org.ensime.api._
 import org.ensime.indexer.{ EnsimeVFS, SearchService }
+import scala.concurrent._
 import scala.concurrent.duration._
 
 trait IsolatedSearchServiceFixture extends IsolatedSourceResolverFixture {
@@ -12,7 +13,7 @@ trait IsolatedSearchServiceFixture extends IsolatedSourceResolverFixture {
     try {
       testCode(config, searchService)
     } finally {
-      searchService.shutdown()
+      Await.ready(searchService.shutdown(), 10.seconds)
       actorSystem.shutdown()
       actorSystem.awaitTermination(10.seconds)
     }
@@ -33,7 +34,7 @@ trait SharedSearchServiceFixture
   }
 
   override def afterAll(): Unit = {
-    _search.shutdown()
+    Await.ready(_search.shutdown(), 10.seconds)
     super.afterAll()
   }
 
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -7,6 +7,7 @@ import org.ensime.api._
 import org.ensime.core.debug.DebugManager
 import org.ensime.indexer._
 
+import scala.concurrent._
 import scala.concurrent.duration._
 import scala.util.Try
 
@@ -18,6 +19,8 @@ class Project(
     broadcaster: ActorRef,
     implicit val config: EnsimeConfig
 ) extends Actor with ActorLogging with Stash {
+  val DB_SHUTDOWN_TIMEOUT = 30 seconds
+
   import context.{ dispatcher, system }
 
   /* The main components of the ENSIME server */
@@ -62,7 +65,7 @@ class Project(
     // make sure the "reliable" dependencies are cleaned up
     Try(classfileWatcher.shutdown())
     Try(sourceWatcher.shutdown())
-    Try(searchService.shutdown())
+    Try(Await.ready(searchService.shutdown(), DB_SHUTDOWN_TIMEOUT))
     Try(vfs.close())
   }
 
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -19,7 +19,7 @@ class Project(
     broadcaster: ActorRef,
     implicit val config: EnsimeConfig
 ) extends Actor with ActorLogging with Stash {
-  val DB_SHUTDOWN_TIMEOUT = 30 seconds
+  private val DB_SHUTDOWN_TIMEOUT = 30 seconds
 
   import context.{ dispatcher, system }
 
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -57,9 +57,9 @@ class DatabaseService(dir: File) extends SLF4JLogging {
   // TODO reverse lookup table
 
   // file with last modified time
-  def knownFiles(): List[FileCheck] = Await.result(db.run(fileChecks.to[List].result), Duration.Inf)
+  def knownFiles(): Future[List[FileCheck]] = db.run(fileChecks.to[List].result)
 
-  def removeFiles(files: List[FileObject])(implicit ec: ExecutionContext): Int = await(
+  def removeFiles(files: List[FileObject])(implicit ec: ExecutionContext): Future[Int] =
     db.run {
       val restrict = files.map(_.getName.getURI)
 
@@ -69,7 +69,6 @@ class DatabaseService(dir: File) extends SLF4JLogging {
         fileChecks.filter(_.filename inSet restrict).delete
       )
     }
-  )
 
   private val timestampsQuery = Compiled {
     filename: Rep[String] => fileChecks.filter(_.filename === filename).map(_.timestamp).take(1)
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -16,7 +16,7 @@ import scala.concurrent.duration._
 import slick.driver.H2Driver.api._
 
 class DatabaseService(dir: File) extends SLF4JLogging {
-  val DB_CREATE_TIMEOUT = 30 seconds
+  private val DB_CREATE_TIMEOUT = 30 seconds
 
   lazy val (datasource, db) = {
     // MVCC plus connection pooling speeds up the tests ~10%
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -74,14 +74,12 @@ class DatabaseService(dir: File) extends SLF4JLogging {
     filename: Rep[String] => fileChecks.filter(_.filename === filename).map(_.timestamp).take(1)
   }
 
-  def outOfDate(f: FileObject)(implicit ec: ExecutionContext): Boolean = {
+  def outOfDate(f: FileObject)(implicit ec: ExecutionContext): Future[Boolean] = {
     val uri = f.getName.getURI
     val modified = f.getContent.getLastModifiedTime
 
-    await(
-      db.run(
-        timestampsQuery(uri).result.headOption.map(_.map(_.getTime < modified).getOrElse(true))
-      )
+    db.run(
+      timestampsQuery(uri).result.headOption.map(_.map(_.getTime < modified).getOrElse(true))
     )
   }
 
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -33,14 +33,13 @@ class DatabaseService(dir: File) extends SLF4JLogging {
     (ds, Database.forDataSource(ds))
   }
 
-  def shutdown()(implicit ec: ExecutionContext): Unit = {
+  def shutdown()(implicit ec: ExecutionContext): Future[Unit] = for {
     // call directly - using slick withSession barfs as it runs a how many rows were updated
     // after shutdown is executed.
-    await(
-      db.run(sqlu"shutdown") flatMap (_ => db.shutdown)
-    )
-    datasource.close()
-  }
+    _ <- db.run(sqlu"shutdown")
+    _ <- db.shutdown
+    _ = datasource.close()
+  } yield ()
 
   if (!dir.exists) {
     log.info("creating the search database")
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -16,6 +16,7 @@ import scala.concurrent.duration._
 import slick.driver.H2Driver.api._
 
 class DatabaseService(dir: File) extends SLF4JLogging {
+  val DB_CREATE_TIMEOUT = 30 seconds
 
   private def await[T](f: Future[T]): T = Await.result(f, Duration.Inf)
 
@@ -44,10 +45,11 @@ class DatabaseService(dir: File) extends SLF4JLogging {
   if (!dir.exists) {
     log.info("creating the search database")
     dir.mkdirs()
-    await(
+    Await.result(
       db.run(
         (fileChecks.schema ++ fqnSymbols.schema).create
-      )
+      ),
+      DB_CREATE_TIMEOUT
     )
   }
 
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -83,11 +83,12 @@ class DatabaseService(dir: File) extends SLF4JLogging {
 
   def persist(check: FileCheck, symbols: Seq[FqnSymbol])(implicit ec: ExecutionContext): Future[Option[Int]] =
     db.run(
-      (fileChecks += check) andThen (fqnSymbols ++= symbols)
+      (fileChecksCompiled += check) andThen (fqnSymbolsCompiled ++= symbols)
     )
 
-  private val findCompiled = Compiled((fqn: Rep[String]) =>
-    fqnSymbols.filter(_.fqn === fqn).take(1))
+  private val findCompiled = Compiled {
+    fqn: Rep[String] => fqnSymbols.filter(_.fqn === fqn).take(1)
+  }
 
   def find(fqn: String): Future[Option[FqnSymbol]] = db.run(
     findCompiled(fqn).result.headOption
@@ -141,6 +142,7 @@ object DatabaseService {
     def idx = index("idx_filename", filename, unique = true)
   }
   private val fileChecks = TableQuery[FileChecks]
+  private val fileChecksCompiled = Compiled(TableQuery[FileChecks])
 
   case class FqnSymbol(
       id: Option[Int],
@@ -179,4 +181,5 @@ object DatabaseService {
     def uniq = index("idx_uniq", (fqn, descriptor, internal), unique = true)
   }
   private val fqnSymbols = TableQuery[FqnSymbols]
+  private val fqnSymbolsCompiled = Compiled { TableQuery[FqnSymbols] }
 }
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -95,9 +95,9 @@ class DatabaseService(dir: File) extends SLF4JLogging {
   private val findCompiled = Compiled((fqn: Rep[String]) =>
     fqnSymbols.filter(_.fqn === fqn).take(1))
 
-  def find(fqn: String): Option[FqnSymbol] = await(db.run(
+  def find(fqn: String): Future[Option[FqnSymbol]] = db.run(
     findCompiled(fqn).result.headOption
-  ))
+  )
 
   import org.ensime.indexer.IndexService._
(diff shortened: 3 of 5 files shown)


fommil:
wow, that was quick! :smile:

cvogt:
yeah, I used slick before ;)

fommil:
how hard would it be to move to the SQL variant of the API?

cvogt:
@fommil takes some manual labor, fragile, remembering SQL dialect specifics. You are happy to do it when I am done, but I am cleaning up some verbosity, so maybe I can convince you to stick with the type-safe one after I am done :).

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -220,8 +220,8 @@ class SearchService(
     backlogActor ! FileUpdate(f, symbols)
   }(workerEC)
 
-  def shutdown(): Unit = {
-    db.shutdown()
+  def shutdown()(implicit ec: ExecutionContext): Unit = {
+    db.shutdown
```
cvogt: should this use Projects ec (like implemented here) or workerEC?

cvogt: also why is workerEC passed in explicitly everywhere, shouldn't it be taken implicitly just right?

fommil: there are a couple of ECs in that code and we wanted to make sure we were using the right one at the right time. It needs a rewrite.

fommil: oh, looks like someone changed it to only use one execution context. Yeah, no need for explicit passing.

cvogt: so Project's actor EC or SearchService's workerEC for the shutdown?

fommil:
looks like its failing the CI. Don't forget the integration test phase. type `sbt it:test`

fommil:
when you're happy, please squash the changelog and I'll merge. Great stuff!

cvogt:
will still do a few things

cvogt:
Are you sure you want the commits to be squashed? I'd write them in small self-contains packages, that are easy to understand individually. I'd also amend the earlier commits to make CI pass. missed the it:

fommil:
yeah, squash is good. We prefer the commits to be user-focussed so that users can just skim the log and see what new features have been added in the last few weeks. In this case "upgraded slick" is the main point.

cvogt:
I mean, whatever you want, but isn't that what merge commits are for? Is the problem only that github doesn't show only merge commits into master by default, but everything? Squashing means throwing away valuable info to devs.

fommil:
That'd be really nice if github presented the merge commits with their PR name, and hid the individual commits, wouldn't it?

Either is fine, just ping when its good. TBH I'll probably be the only person poking around this code for the foreseeable future as Aemon is focussed on Java support, Eric is dropping in random awesome features and Rory is looking at the compiler APIs.

cvogt:
@fommil Either way works for me, too. Just a healthy questioning of the reasons brought up :).

fommil:
sorry our CI is running a bit slow tonight because the other PR is doing a lot of downloading of jars that haven't been cached by `master`.

cvogt:
no worries. I'd ask you for the thorough review in the end as well, because all this changing side-effecty stuff to async is fragile. Forget to await a Future[Unit] where you previously returned a Unit and the side-effect may never happen and the type-system doesn't complain, because hej, a Future[Unit] is good enough when Unit is expected, no ;)?

fommil:
yes, I know... I was shuddering when I seen you touch it because I know its like a house of cards! :smile: I'll check out your branch when you're happy and do a review. It's very nice timing actually, because I'd decided to dedicate tomorrow to closing the highly requested #425

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -95,54 +92,61 @@ class SearchService(
...
-      index.commit()
-      log.debug("...done committing index")
-      count
-    }
-
-    for {
-      r <- removed
-      i <- indexedAndCommitted
-    } yield (r, i)
+    } yield (removed.size, persisted.size)
```
cvogt: a single comprehension encompassing all of the above seemed cleaner here and since I had to touch it anyways

cvogt:
This is ready for a thorough review now. DatabaseService is fully async. SearchService blocks right away in some cases, which means there is room to push that out even further, but at least DatabaseService won't need to be touched for this.

I'll give this another review myself to try to see if I forgot any blocking required for side-effects to happen. After the review I'll fix up what's required so we get this ready to merge.

I still want to do some refactoring inside of DatabaseService, but no need to do this in the same PR.

cvogt:
actually the changes were so few, I did add them here after all

cvogt:
Alright, I used Scala's deprecation warnings to track down the usages of all methods I changed and make sure they properly block when necessary. Found a few spots, and committed fixes above. I'd probably rebase and integrate them into the original commits before merging. (Or squash it all if desired).

fommil:
you'll have a very small conflict to merge with upstream, or just grab it from the version of the PR in my branch.

core/src/it/scala/org/ensime/fixture/SearchServiceFixture.scala
```diff
@@ -12,7 +13,7 @@ trait IsolatedSearchServiceFixture extends IsolatedSourceResolverFixture {
     try {
       testCode(config, searchService)
     } finally {
-      searchService.shutdown()
+      Await.ready(searchService.shutdown(), 10.seconds)
```
fommil: In the tests we can use the akka test utils timefactor. Although I'm not sure what to do about the timeouts in the `main` code, they might be prone to random test failures on slow CI hardware.

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -62,7 +65,7 @@ class Project(
     // make sure the "reliable" dependencies are cleaned up
     Try(classfileWatcher.shutdown())
     Try(sourceWatcher.shutdown())
-    Try(searchService.shutdown())
+    Try(Await.ready(searchService.shutdown(), DB_SHUTDOWN_TIMEOUT))
```
fommil: I'm tempted to just not wait on this. The reason for all the `Try`s is that we don't care about the outcome, we just want to take best endeavours to shut down cleanly before it annoys the user and they force the kill.

core/src/main/scala/org/ensime/indexer/DatabaseService.scala
```diff
@@ -28,83 +32,76 @@ class DatabaseService(dir: File) extends SLF4JLogging {
...
     log.info("creating the search database")
     dir.mkdirs()
-    db withSession { implicit s =>
-      (fileChecks.ddl ++ fqnSymbols.ddl).create
-    }
+    Await.result(
+      db.run(
+        (fileChecks.schema ++ fqnSymbols.schema).create
+      ),
+      DB_CREATE_TIMEOUT
```
fommil: Given the way that ENSIME works, the end effect to the user is going to be the same whether the creation succeeds or fails. Therefore I'm tempted to remove the timeout. If ENSIME doesn't come up, the user will see that the latest log message is "creating the search database" and then no more processing. Perhaps we should have the response log message "... created the search database".

core/src/main/scala/org/ensime/indexer/DatabaseService.scala
```diff
@@ -28,83 +32,76 @@ class DatabaseService(dir: File) extends SLF4JLogging {
...
-      u <- fileChecks if u.filename === filename
-    } yield u.timestamp
-
-    query(uri).list.map(_.getTime).headOption match {
-      case Some(timestamp) if timestamp < modified => true
-      case Some(_) => false
-      case _ => true
-    }
+    db.run(
+      timestampsQuery(uri).result.headOption.map(_.map(_.getTime < modified).getOrElse(true))
```
fommil: I find the nested maps to be confusing, I might bring back the pattern match above. More verbose, yes, but it clearly shows the three conditions, which are easy to reason about.

cvogt: You probably know that: But the nested maps are because of the added Future Monad, now Future[Option[...]]. I find nested Monads are generally confusing. For the handling of the inner one here, I don't find the match helping my understanding. Happy to bring it back if you want. Just tell me what's needed to get this merged.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -30,6 +31,8 @@ class SearchService(
 ) extends ClassfileIndexer
     with ClassfileListener
     with SLF4JLogging {
+
+  private val QUERY_TIMEOUT = 30 seconds
```
fommil: hmm, timeouts here make me nervous. They will fail when the CI machines are running slowly, and sometimes the queries can be really long on a user machine. It all happens in the background so users don't typically notice (e.g. if H2 decides to recreate its memory space on disk)

core/src/main/scala/org/ensime/indexer/DatabaseService.scala
cvogt: you mean Duration.Inf?

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -64,85 +67,85 @@ class SearchService(
...
-    // individual DELETEs in H2 are really slow
-    val removing = stale.grouped(1000).map { files =>
-      Future {
-        index.remove(files)
-        db.removeFiles(files)
-      }(workerEC)
-    }
-
-    val removed = Future.sequence(removing).map(_ => stale.size)
+    for {
```
fommil: ok, this is as far as I can get with a visual review. I'm going to need to check this out and look in detail because the original code here is extremely fragile. I'll push the changes suggested above to my branch to save you the effort.

core/src/main/scala/org/ensime/indexer/DatabaseService.scala
fommil: I'll try a couple of variations and see what seems easiest to understand and you can have a look and let me know what you think.

core/src/main/scala/org/ensime/core/Project.scala
cvogt: What do you suggest concretely? What does not wait mean? 0 seconds? That's just like skipping the whole call and not shut Slick down. Not sure what the implications are. How about reducing it to 5 seconds or something? Not sure if this ever takes longer than a split second anyways, but better to have a timeout instead of Duration.Inf just in case.

core/src/it/scala/org/ensime/fixture/SearchServiceFixture.scala
cvogt: I have no idea, I just picked 10 seconds because of the 10 seconds below and though that Slick should reasonably shut down much faster than that. Tell me if a change is needed.

core/src/main/scala/org/ensime/indexer/SearchService.scala
cvogt: So what do you suggest? Increase? Block infinitely? We need one for Await.result/Await.ready

core/src/main/scala/org/ensime/indexer/DatabaseService.scala
fommil: yeah. The end effect to the users is the same because they'll still be waiting forever in their ENSIME client even if this throws.

core/src/main/scala/org/ensime/indexer/SearchService.scala
cvogt: cool. How do you suggest handling potential history re-writes here and integration of those changes?

cvogt:
Ok, awaiting your changes on top of this then.

core/src/main/scala/org/ensime/indexer/SearchService.scala
fommil: We can leave these ones as they are all QUERY related, I thought perhaps they were also for the writes. Although I suspect we're not expecting an exception in the downstream usecase, which might result in blowing up the actors, but that's a different problem.

fommil:
on a fresh version of `ensime-server` I see

```
11:57:45.891 DEBUG akka://ENSIME/user/ensime-main/project o.e.c.Project - indexed 1 and removed 0
```

so something subtle has changed (I would have expected more like 1,000 things indexed on startup)

fommil:
weird, we have tests around that and they seem to be passing.

fommil:
above comments are in https://github.com/fommil/ensime-server/commits/slick-3.1

fommil:
hold on, I think that lasts commit broke everything

fommil:
fixed now. Brackets matter around there.

fommil:
hmm, and now the indexing job doesn't complete. OK, definitely something funky going on... I'll delve in. Back later.

core/src/main/scala/org/ensime/indexer/SearchService.scala
fommil: just `git --reset hard fommil/slick-3.0` and that'll handle everything.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -64,85 +67,85 @@ class SearchService(
...
 
   def refreshResolver(): Unit = resolver.update()
 
-  def persist(check: FileCheck, symbols: List[FqnSymbol]): Unit = try {
+  def persist(check: FileCheck, symbols: List[FqnSymbol]): Future[Unit] = try {
     index.persist(check, symbols)
-    db.persist(check, symbols)
+    db.persist(check, symbols).map(_ => ())
   } catch {
     case e: SQLException =>
```
fommil: I'm assuming that if the underlying slick layer throws an exception, it'll only be visible when we try to read the result of the Future. In which case, the below exception logic needs to move, or be attached as an `onFailure`

core/src/main/scala/org/ensime/indexer/DatabaseService.scala
fommil: I'm an idiot, we have a convenience on `FileCheck` to do exactly this.

core/src/main/scala/org/ensime/indexer/SearchService.scala
cvogt: You are totally right! Good catch!

cvogt: .recover in case of try-catch should do it

cvogt: http://www.scala-lang.org/api/current/index.html#scala.concurrent.Future@recover[U>:T](pf:PartialFunction[Throwable,U])(implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[U]

fommil:
this is turning into a long needed rewrite of `SearchService`

cvogt:
I'd suggest manually bisecting through the commits here to see where the problem introduced. Or maybe linear searching form the start. Do you mind locating the problem @fommil ? Most likely source is def refresh() I'd guess.

fommil:
yeah, I'm rewriting `refresh`... it was a mess before you even got here :wink:

cvogt:
@fommil don't you think we should locate the problem first, fix it, merge and then do the re-write :)?

fommil:
I'm almost done. It's more of a refactor than a rewrite, just putting everything into little blocks of work instead of a big mish-mash. It's still not pretty, but it's better. Almost done.

fommil:
I'm seeing these exceptions

```
java.util.concurrent.RejectedExecutionException: null
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:1774) ~[na:1.6.0_45]
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:768) ~[na:1.6.0_45]
	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:656) ~[na:1.6.0_45]
	at scala.concurrent.impl.ExecutionContextImpl$$anon$1.execute(ExecutionContextImpl.scala:136) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.backend.DatabaseComponent$DatabaseDef$class.runSynchronousDatabaseAction(DatabaseComponent.scala:230) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.jdbc.JdbcBackend$DatabaseDef.runSynchronousDatabaseAction(JdbcBackend.scala:38) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.backend.DatabaseComponent$DatabaseDef$class.runInContext(DatabaseComponent.scala:207) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.jdbc.JdbcBackend$DatabaseDef.runInContext(JdbcBackend.scala:38) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.backend.DatabaseComponent$DatabaseDef$class.runInContext(DatabaseComponent.scala:146) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.jdbc.JdbcBackend$DatabaseDef.runInContext(JdbcBackend.scala:38) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.backend.DatabaseComponent$DatabaseDef$class.runInternal(DatabaseComponent.scala:75) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.jdbc.JdbcBackend$DatabaseDef.runInternal(JdbcBackend.scala:38) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.backend.DatabaseComponent$DatabaseDef$class.run(DatabaseComponent.scala:72) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at slick.jdbc.JdbcBackend$DatabaseDef.run(JdbcBackend.scala:38) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
	at org.ensime.indexer.DatabaseService.outOfDate(DatabaseService.scala:81) ~[ensime_2.11.7-0.9.10-SNAPSHOT-assembly.jar:0.9.10-SNAPSHOT]
```

which would imply that the search parameter to `timestampsQuery` is `null`, but I added some logging and it is not. Anything else that can cause this?

fommil:
the `null` could be irrelevant. Are we submitting too many jobs to the queue?

fommil:
yeah, I'm getting convinced that's what is happening here... and then the indexing job never finishes. We have one of these

```
  search-service-dispatcher {
    type = Dispatcher
    executor = "thread-pool-executor"
    thread-pool-executor {
      # minimum number of threads to cap factor-based core number to
      core-pool-size-min = 2
      # No of core threads ... ceil(available processors * factor)
      core-pool-size-factor = 2.0
      # maximum number of threads to cap factor-based number to
      core-pool-size-max = 10
    }
  }
```

I'll customise it further to have an infinite queue.

cvogt:
@fommil mh, are we trying to run queries after the slick Database has been .shutdown() somehow? (By the way, there is also a blocking .close(), if that makes things easier)

cvogt:
See this for adjusting the queue size: http://slick.typesafe.com/doc/3.1.0/database.html#database-thread-pool

Number of threads should stay small though.

cvogt:
let me know if it helped

fommil:
it looks like our custom `ExecutionContext` had a bounded limit of tasks it could queue. I've made that infinite but I'm still getting rejected execution exceptions (this time in `persist`). Data database has not been closed here.

fommil:
aah, looks like slick has its own one! Ok, this looks like it will help.

fommil:
does the thread size need to match the size of our BoneCP thread pool?

fommil:
:tada: that was it :smile: needed to set the queue size to `-1` for infinite. Startup on a big project is extremely demanding on the CPU, memory and database, so I'm not surprised about this.

fommil:
ok, I'm happy with it now. It looks solid, thank you!

Feel free to grab my branch and squash/don't squash as you see fit (we don't have a rigid policy, but we do tend to prefer squashing). You can take my review commits as your own unless you don't want to have your name beside `refresh()` (I wouldn't blame you).

cvogt:
yes slick has it's own thread pool exclusively for DBIO, which does the magic for async performance. I can explain the details when we run into each other at some conference or you watch Stefan's talk.

fommil:
I raised another PR with commits squashed to get CI running. If you do the squashing here, I can just close that one.

cvogt:
ok, will rearrange things here

cvogt:
actually, I will close this to keep a github record and re-open a new one for the sqash

PR #1120, opened 2015-10-24
Add java type checking and doc finding.

fommil:
This is a rough foundation for java support. The new JavaCompiler is
based on the javac apis introduced in java6.


fommil:
@aemoncannon I squashed all my changes into your commit since it skips out the changes to the Analyzer. But this has flagged up that the jcenter repo is not in sbt by default on Windows, which is just awesome.

fommil:
oh the joys of sbt https://github.com/sbt/sbt/issues/2253

fommil:
OMG, we just got green lights on Windows! :smile:

fommil:
if this goes green on Linux, we're good to merge. I've added the necessary changes to the emacs side for the resolver to work and pinged the Vim/Atom/Sublime folk.

aemoncannon:
Awesome.Thanks for all your help. I'm going to keep chiseling away at
scala/java parity..

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  // needs to be recreated in JDK6. JDK7 seems more capable of reuse.
+  def getTask(
+    lint: String,
+    listener: DiagnosticListener[JavaFileObject],
+    files: List[JavaFileObject]
+  ): JavacTask = {
+    // TODO: take a charset for each invocation
+    val compiler = ToolProvider.getSystemJavaCompiler()
+    val fileManager = compiler.getStandardFileManager(listener, null, DefaultCharset)
```
aemoncannon: I think we'll lose out on caching within the file manager if we recreate this. I need to benchmark.

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -0,0 +1,162 @@
...
+    compiler.getTask(null, fileManager, listener, List(
+      "-cp", config.allJars.mkString(File.pathSeparator),
```
aemoncannon: Make the cp string once?

core/src/main/scala/org/ensime/core/JavaAnalyzer.scala
```diff
@@ -0,0 +1,72 @@
...
+    case CompletionsReq(f, point, maxResults, caseSens, _) =>
+      sender() ! javaCompiler.askCompletionsAtPoint(f, point, maxResults, caseSens)
+
+    case DocUriAtPointReq(file, range) =>
+      sender() ! javaCompiler.askDocSignatureAtPoint(SourceFileInfo(file, None, None), range.from)
+
+    case TypeAtPointReq(file, range) =>
+      sender() ! javaCompiler.askTypeAtPoint(SourceFileInfo(file, None, None), range.from)
+
+    case SymbolDesignationsReq(f, start, end, tpes) =>
```
aemoncannon: Previously the various javaCompiler.ask* methods could execute in parallel. Did you mean to force serial?

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
fommil: I'd eat my hat if this is a performance matter. I think we do this in two places though so maybe it should be moved to EnsimeConfig (or its typeclass)

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
fommil: Hmm, maybe we could switch based on Java version

core/src/main/scala/org/ensime/core/JavaAnalyzer.scala
fommil: Yeah, because the actor would expect that. If we want to do it in parallel, its trivial we just spawn the requests in an anonymous actor (better than future)

fommil:
ok, I'm going for it!!

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
fommil: actually, if you're benchmarking, can you also please check that the various javac tasks and file managers are correctly garbage collected. Thought just occured to me after merging.

fommil:
@aemoncannon great stuff on the java/scala parity! My vote goes on "go to definition" for next feature :smile:

BTW, how do you see the protocol evolving? I'd like to see us move to a new sealed trait of requests into the JavaAnalyzer, with a sealed trait of responses. Initially they could have the same contents as the scala variants, but we had some plans to extend the messages post-1.0 to be more explicit about JVM/Scala names and I'd hate for us to be constrained by the Java support.

aemoncannon:
Just remembered: would you look at my elisp PR? Its necessary for method
param expansion.

fommil:
sure, I'll look now. I'm battling with the CI (that I broke) but have it all working again. I'm going to self merge the niggles PR unless you want to have a quick look.

PR #1123, opened 2015-10-26
A few followup fixes for Java support

aemoncannon:
* Prefer completions with exact match + test
* Etc.


core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -39,8 +40,7 @@ class JavaCompiler(
     val compiler = ToolProvider.getSystemJavaCompiler()
     val fileManager = compiler.getStandardFileManager(listener, null, DefaultCharset)
     compiler.getTask(null, fileManager, listener, List(
-      "-cp", config.allJars.mkString(File.pathSeparator),
```
aemoncannon: (i have several hundred jars on my cp)

PR #1124, opened 2015-10-26
A couple java things that were broken in the shuffle.

aemoncannon:


fommil:
I'm actually putting together a tiny PR around the ensime-sbt langtools source files. I'll fix the obvious `EnsimeConfig` / classpath bug while I'm there. I'm not sure what to do about the messages on startup though. Maybe worth just fixing it in the client?

aemoncannon:
Re messages. I'd rather not change the clients. Can we just keep the
behavior consistent with Scala for now?

fommil:
No problem... we can always change the client later when the java stuff gets its own message set (or just drop it if we have time and it makes sense). I'm going to bed now but I think you're going to need to change the assertion in `ProjectFixture` for the integration tests to pass. Can you please merge my PR if it goes green? That'll at least fix the classpath issue.

BTW, if you download the drone client (and have docker installed) you can run the full CI tests locally by typing `drone exec` http://readme.drone.io/docs/cli/index.html

aemoncannon:
I'll rebase to yours once it goes green and you merge.

aemoncannon:
@fommil Appveyor failing in some kind of websocket flow test. My guess is it's a flaky test. Would you take a look?

fommil:
Hmm, I don't see why that test should fail. Might just be dodgy running out cheap hardware, an it often does at that time of night.

fommil:
Actually I believe the integration tests should fail with this update. You'll need to update the expectations of the project fixture as its going to be seeing a lot more messages on startup now.

aemoncannon:
Oh, duh. We'll be sending them once for Scala and once for Java. Hmmm. I think you were right in the first place and I should just fix the elisp to no longer wait for analyzer-ready.

Will this break anything server side? does server queue things correctly if client starts sending stuff before the pres compiler is ready?

aemoncannon:
On third thought, that'd be a pretty disruptive change. Also, from a UI standpoint, it's nice to know when things are ready. Either we do it explicitly with analyzer-ready, or we do it implicitly by making the user wait for the rpc queue to drain -- I like the former.

aemoncannon:
Here's an idea: we wait at Project for CompilerReady from *both* Analyzers, then we send the single CompilerReady to the client. This comes with no external behavior change, as java ready will be instantaneous.

rorygraves:
That would work (waiting for both internally and emitting only one)

fommil:
The server does the right thing. From an API perspective, as soon as you get the connection info message, you're good to go.

fommil:
Re: sending only once. Have a look at broadcaster. We could make it not send duplicate persisted messages and I think that achieves what you want.

fommil:
I'll be home in about an hour, I can put together a small PR unless you get there first. Broadcaster feels like the right place and might just be a change of a List to a Set.

fommil:
It'd be really nice if the UI didn't care though :-)

fommil:
ha! `Broadcaster` already does what I was suggesting. That means the client should only see these messages once.

fommil:
see #1129

PR #1127, opened 2015-10-30
Revert add-import refactoring

dickwall:
There seems to be an unwanted interaction with the updated refactoring
library in this code which causes a regression in how add-imports works.
I tested the original code which seems to work in all cases now, including
the case that triggered this change, so perhaps the bug in the underlying
library was fixed?


fommil:
it looks like the tests have failed because of import reordering, but the functionality is ok.

fommil:
see #1129

PR #1141, opened 2015-11-05
Finish revert of java events. Suspect they're breaking elisp tests.

aemoncannon:


aemoncannon:
Build is already broken. This should fix. Merging.

PR #1142, opened 2015-11-05
Await ConnectionInfoReq before initializing project.

aemoncannon:
All other messages are buffered until ConnectionInfoReq has been handled. The object is to alleviate a race condition in clients where the protocol version is required to send an rpc, but the connection info hasn't been received yet. NOTE: This PR does not eliminate the race, but should make it much lower probability. The next step is to make the connection info request blocking on the client.


fommil:
When we move to websockets, I think we might be able to send the protocol version back in the connection header and just lose connection info altogether.

We could just set the default protocol version to the current one and not move it forward for a while.

PR #1145, opened 2015-11-07
Go to declaration for Java (WIP)

aemoncannon:
TODO:
 * Implementation of askLinkPos to get fine-grained positions.
    ** should be pretty easy. just need to parse the file, then walk the tree looking for the element

### DIFF
--- a/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
@@ -3,20 +3,24 @@ package org.ensime.core
 import java.io.File
 
 import akka.event.slf4j.SLF4JLogging
+import org.ensime.api.OffsetSourcePosition
+import org.ensime.api.LineSourcePosition
 import org.ensime.core.javac.JavaCompiler
 import org.ensime.fixture._
 import org.ensime.indexer.EnsimeVFS
+import org.ensime.indexer.SearchServiceTestUtils
 import org.scalatest._
 import scala.concurrent.Await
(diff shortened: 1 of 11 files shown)


aemoncannon:
Btw, I love having the elisp tests running for the server. Such warm feels.

fommil:
heh, I think I'd go with "less nervous"

fommil:
interestingly it fails on windows too.

Btw, you can force a JDK in `sbt` if you're using paulp's `sbt-extras` by creating a `.sbtopts` file.

aemoncannon:
Oh good tip

aemoncannon:
@fommil Been hitting some sexp <=> string QuickCheck failures. Can you make any sense of it?

fommil:
If they happen just restart. They are extremely rare.

aemoncannon:
Woo! I had relax the test to make this pass, but I think I know how to fix things. The root of the trouble is in looking up Elements given a Tree or a Type. I've got some ideas how to make this more robust across JVMs.

fommil:
Prepare thyself for the hack day tomorrow :-|

aemoncannon:
Should I be worried? : )

fommil:
Apparently there might be 10+ people there keen to do ENSIME stuff. Although might be mostly sublime, Vim and atom.

aemoncannon:
Hey that's pretty awesome. At some point I want to evangelize at work and
the additional clients will be a great selling point :)

fommil:
I think Josh of SBT fame recently rejoined you. Maybe hunt him down?

Issue #1068, opened 2015-08-20
scalariform dies if it doesn't like the input


fommil:
we're not handling this error case

```
12:01:03.869 ERROR akka://ENSIME/user/ensime-main/project/analyzer a.a.OneForOneStrategy - Expected token EOF but got Token(RBRACE,},6470,})
scalariform.parser.ScalaParserException: Expected token EOF but got Token(RBRACE,},6470,})
	at scalariform.parser.ScalaParser.scalariform$parser$ScalaParser$$accept(ScalaParser.scala:80) ~[scalariform_2.11-0.1.6.jar:0.1.6]
	at scalariform.parser.ScalaParser.compilationUnit(ScalaParser.scala:1811) ~[scalariform_2.11-0.1.6.jar:0.1.6]
	at scalariform.parser.ScalaParser.compilationUnitOrScript(ScalaParser.scala:55) ~[scalariform_2.11-0.1.6.jar:0.1.6]
	at scalariform.formatter.ScalaFormatter$$anon$1.parse(ScalaFormatter.scala:480) ~[scalariform_2.11-0.1.6.jar:0.1.6]
...
	at org.ensime.core.RefactoringHandler$class.handleFormatFile(Refactoring.scala:118) ~[core_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.Analyzer.handleFormatFile(Analyzer.scala:41) ~[core_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.Analyzer$$anonfun$allTheThings$1.applyOrElse(Analyzer.scala:259) ~[core_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.Analyzer$$anonfun$allTheThings$1.applyOrElse(Analyzer.scala:152) ~[core_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36) ~[scala-library-2.11.7.jar:na]
	at org.ensime.core.Analyzer$$anonfun$ready$1.applyOrElse(Analyzer.scala:149) ~[core_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
```

PR #1149, opened 2015-11-14
#1068 scalariform dies if it doesn't like the input

VlachJosef:
ScalaFormatter.format throws ScalaParserException. When this happens formatting will not be applied and original file content will be returned. I've provided test for this issue.


VlachJosef:
I think there is actually way to do 'no op'. I'll give it a try.

rorygraves:
Nice!

Issue #595, opened 2014-09-16
lazy vals semantically highlighted as vars


fommil:
I just noticed that references to lazy vals are being given the `scala-font-lock:var-face`. That is wrong, they should be getting the `scala-font-lock:val-face` (or is it just `font-lock-variable-name-face`)

Not sure if this is a bug in the client or the server yet: a test case in the server shouldn't be too hard to set up.

noelmarkham:
Thanks to @fommil in helping get me started with `ensime-server` development today.

A few things about this bug:
* It's actually only `lazy val` _fields_ that are suffering from this issue:  <img width="264" alt="screen shot 2015-11-14 at 20 48 08" src="https://cloud.githubusercontent.com/assets/1492487/11165757/0c55733e-8b11-11e5-9990-becff4206e37.png">
* There's some odd behaviour in the presentation compiler which says that lazy val fields are actually variables. I can provide evidence if you want - just let me know how/where.

I have a fix for this, I'm just preparing the PR.

rorygraves:
I would be interested in the evidence (mostly out of idle curiosity I accept).  Looking forward to the PR.

PR #1150, opened 2015-11-14
lazy vals semantically highlighted as vars #595

noelmarkham:


core/src/main/scala/org/ensime/core/SemanticHighlighting.scala
```diff
@@ -56,7 +56,9 @@ class SemanticHighlighting(val global: RichPresentationCompiler) extends Compile
           }
           if (sym.hasFlag(ACCESSOR)) {
             val under = sym.accessed
-            if (under.isVariable) {
+            if (sym.hasFlag(LAZY)) {
```
rorygraves: I would be tempted to add a comment here stating that the compiler mis-reports it and that lazy can only be a val.

fommil:
great stuff!

PR #1154, opened 2015-11-15
Fix source file existence check.

aemoncannon:
No reason to fail if the SourceFileInfo points to source.


fommil:
smashing. I was hoping to move us away from `FileUtils` but there doesn't seem to be a natural place for these kinds of utils at the moment.

PR #1156, opened 2015-11-16
Fix a bug breaking purely in-memory files.

aemoncannon:
It turns out that passing a string to BatchSourceFile constructor is the magic password to force it to use a VirtualFile internally. After this PR, completion should work for in-memory files with no backing disk file.


fommil:
looks like an encoding issue with the in-memory sources

aemoncannon:
sweet. i've got an elisp PR coming soon to exercise this in a test.

PR #1164, opened 2015-11-28
Fixes for in-memory files

fommil:


fommil:
hmm, looks like the `ensime-emacs` test is broken against `master` for this in-memory stuff.

fommil:
@aemoncannon after looking into this, it is pretty clear that `scalac` is not able to handle `BatchSourceFile`s that refer to the same `path` if one (and only only) of the implementations is a `VirtualFile`. There are two ways we can go if you want to add support for truly virtual files:

1. every time we add a file to the presentation compiler, check if it is already registered as an in-memory file (i.e. backed by `VirtualFile`) and remove that version.
2. allow `VirtualFile` backed files if their `.getName` starts with a `*` (i.e. an emacs buffer) and disallow files that exist on disk if their name starts with a `*`

Probably 1 is The Right Way to do it as it supports files that have not yet been saved properly.

fommil:
this is how I feel... ackbar relaxes

aemoncannon:
Bummer. I guess #1 doesn't sound too bad. I'm still confused about how
scalac even knows the difference between virtual and non. Some api method
of SourceFile must be betraying us right?

fommil:
yeah, it seems like maybe something is doing `equals` on the `AbstractFile` and it isn't just doing path comparisons. My change to the `Completion` logic should mean that whatever we do in the `createSourceFile` will work there.

For 1, we'd probably want to have an `if f.exists()` guard on the current `createSourceFile` matchers, falling through to the `VirtualFile` variant. But I'm not sure how tricky its going to be.

BTW, I removed the emacs test for this stuff so you will need to revert that if you want the test coverage. However, it should be relatively easy to create a standalone integration test for this, just follow the `timingTest` pattern.

aemoncannon:
I think I see what's happening. The unitOfFile map in interactive/Global.scala is LinkedHashMap[AbstractFile, RichCompilationUnit], and the implementations of PlainFile, VirtualFile are mutually non-equal with each other.. so the compiler tracks separate compilation units for the virtual vs non-virtual.

fommil:
Good find!! I actually centralised the construction of all AbstractFiles.... we *could* try overriding both (this might be a terrible idea)

aemoncannon:
Yeah.. that does feel a bit cavalier. I wonder if we could subclass PlainFile to get the behavior we need for memory files.

Issue #1160, opened 2015-11-24
<blah> is defined twice conflicting both originated in...


fommil:
(fix pushed to master)

_From @gaydenko on November 19, 2015 15:38_

![1447938705](https://cloud.githubusercontent.com/assets/3740016/11276237/5248ae3a-8ef2-11e5-9f36-c7a669f1d767.png)
After today updating of ensime-emacs to ensime-20151117.2103 starting of editing of any file results in the error, where artifact can be a method, a val or even internal method as $init$. The issue doesn't depend on the code under editing. Say, this simple project https://github.com/gaydenko/temp01 is also affected by the issue. ensime-sbt 0.2.1 is in use as a global plugin. Java is Oracle one build 1.8.0_66-b17. Scala is 2.11.7. Up to date Arch Linux is in use, kernel 4.2.5.

Have tried without success:

 -   to clean ensime_cache
 -   to redownload the world by deleting ~/.ivy2

Few last commits to server are related to files manipulations, and it can be a reason of accounting the same file twice. Crazy idea is if developers use, say, Mac treating files manipulations different in comparison with, say, Linux.

fommil:
Thank you for using ENSIME!

We are getting a lot of bug reports recently that are missing key information, so we're using https://github.com/Astrac/github-autoresponder to ask that everything is captured in the ticket.

Before somebody responds to this ticket, can you please confirm that you have checked the following:

* are you using the latest release of ensime-emacs? Make sure to use MELPA to update your packages regularly.
* try nuking your ivy cache (specifically `~/.ivy2/cache/org.ensime`), then type `ensime-update` before trying again (upstream bug in `sbt`).
* check the [tickets flagged as FAQ on the server](https://github.com/ensime/ensime-server/issues?labels=FAQ) and [tickets flagged as FAQ for Emacs](https://github.com/ensime/ensime-emacs/issues?labels=FAQ).
* check the [most recently updated tickets](http://github.com/ensime/ensime-emacs/issues?direction=desc&sort=updated) and do a quick search to see if anybody has reported this already.
* if its about a non-scala specific question, check that the thing you're experiencing isn't from another Emacs mode. If you don't have an Emacs guru to hand, try asking on our gitter channel https://gitter.im/ensime/ensime-emacs

If that solved your problem, great! Please close the ticket. If not, please read on...

[...]

PR #1165, opened 2015-12-02
Restore unsaved file support.

aemoncannon:
Turns out explicitly passing plain-old PlainFiles does the trick. Virtual file not necessary, and we avoid the PlainFile vs VirtualFile nonsense of #1160. Plus a test.


fommil:
This is extremely weird.

fommil:
Will you add the Emacs test again now?

PR #1167, opened 2015-12-04
Bump protocol for changes in a71407d3.

aemoncannon:
Need this so we can check version in client and send sourcefileinfo.

### DIFF
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -546,7 +546,7 @@ case class EnsimeImplementation(
 case class ConnectionInfo(
   pid: Option[Int] = None,
   implementation: EnsimeImplementation = EnsimeImplementation("ENSIME"),
-  version: String = "0.8.17"
+  version: String = "0.8.18"
 ) extends RpcResponse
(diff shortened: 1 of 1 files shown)


fommil:
hmm, that merge was a little premature, I should have asked for some documentation in `swank/CHANGELOG.md`

aemoncannon:
Ah. Damn. I will follow up.

fommil:
nah, don't bother... just stick it in a commit with your next PR :smile:

Sorry, I've not had time to look at the Java one, I've been swamped with work and talk prep. I'll get to it over the coming weeks.

aemoncannon:
No worries. I actually want to look at our CI w.r.t jvm version coverage.
Do you think it's feasible to run the integrations vs 1.6,1.7,1.8? I think
it's the version of tools.jar that is causing pain.

fommil:
Yup! You'd need to install the jdks in the docker image (1.X branch in ensime-docker). Comments there, because I thought we'd do this post 1.0.

Tests will probably fail on non-JDK6 runs because of bytecode changes resulting in search priority changes.

You *could* just disable some tests on jdk6 and promise me they pass :-)

PR #1169, opened 2015-12-06
Fix a couple bugs in Java unsaved file flow.

aemoncannon:


fommil:
You really love these unsaved files!

fommil:
Looks like a real failure on `Starting test: 'Test misc operations on unsaved source file.'`

Now that the emacs tests have really stabilised, we should probably look to reduce the logging levels to just the test name and the failure message.

aemoncannon:
Looks like a dumb mistake I put in ensime-emacs. Odd that tests didn't catch it..
PR'ed over there, and then will poke this branch to rebuild.

aemoncannon:
Hah, yeah I'm obsessed with unsaved files! I feel like it's something new users would hit a lot. Or I'm just putting off doing important java stuff ;-)

aemoncannon:
Is there any way to poke drone to re-check without pushing change?

fommil:
@aemoncannon yeah, just log in and click the restart button. It ignores you if you don't go to fommil.com and login. The authors know about the bug.

PR #1170, opened 2015-12-08
Add mutable working set of source files to JavaCompiler.

aemoncannon:
This gives visibility on inter-unit api changes as you edit, without requiring recompile. Note we haven't wired up compiler restart for java projects yet, so it won't respond to recompiles, and will just get slower and slower as you work, but so far is not a problem.


aemoncannon:
p.s. I'm actually using this at work today(!!!) and it's awesome.

rorygraves:
Sounds awesome!

fommil:
I'm ridiculously close to being allowed to use ENSIME at work, so expect super large Scala project support over coming year

PR #1171, opened 2015-12-09
Add a fallback doc handler for android online reference.

aemoncannon:


fommil:
Do they have versions of their API?

PR #1172, opened 2015-12-12
Move into it's own file. Fix test and related code.

aemoncannon:


aemoncannon:
@fommil 
Incidentally, how bad do you think it is that we're using com.sun.* apis? Specifically the Compiler Tree API. Netbeans depends on it too, but I'm guessing they ship a JDK. I've already witnessed some instability between java versions (nothing we can't hack around).

aemoncannon:
See
http://blog.codefx.org/java/dev/how-java-9-and-project-jigsaw-may-break-your-code/

"Internal JARs like lib/rt.jar and lib/tools.jar will no longer be accessible. Their content will be stored in implementation-specific files with a deliberately unspecified and possibly changing format.

Code which assumes the existence of these files, will stop working correctly. This might also lead to some transitional pains in IDEs or similar tools as they heavily rely on these files."

aemoncannon:
Ah, but at least the Compiler Tree API is marked @jdk.exported, so it will at least be accessible in Oracle Java 9. https://docs.oracle.com/javase/8/docs/jdk/api/javac/tree/jdk/class-use/Exported.html

fommil:
Cool. We'd probably just do whatever NetBeans does.

PR #1174, opened 2015-12-12
Idea

rumoku:


fommil:
Not sure about this. It's the old user vs project config options problem.

rumoku:
https://github.com/Astrac/consierge/blob/master/.gitignore - as a comparison.

fommil:
Just put this stuff in your user config and you'll never have to set it again

Issue #996, opened 2015-06-15
prepare/exec refactor should be one call...


fommil:
... that simply returns the diff of what should happen.

This places the responsibility of the edit (and undos) in the hands of the editor.

rorygraves:
@fommil and I discussed this idea yesterday.  The current refactor undo is very fragile if, between doing refactor and undo you make other non-refactor edits.  The idea is by supplying the information to the client 'these 4 files need these changes', the client can insert them into its undo stack allowing IntelliJ style 'are you sure you want to undo 'refactor name' operation?'

2.0 feature as it requires a non-trivial amount of client side work otherwise we break the existing refactoring support. 

@fommil Maybe we should just remove the 'undo stack' within the server for 1.0?

fommil:
Yeah, it's complicated and I wouldn't trust it over my VCS anyway.

aemoncannon:
+1

fommil:
this has been done!

PR #1180, opened 2015-12-17
Add RefactorReq and RefactorDiffEffect (#996)

lxol:
Extend the ensime protocol to request a refactor change as a diff file in a unified output format.

### DIFF
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -236,6 +236,15 @@ case class PrepareRefactorReq(
 ) extends RpcAnalyserRequest
 
 /**
+ * Responds with a `RefactorFailure` or a `RefactorDiffEffect`.
+ */
+case class RefactorReq(
+  procId: Int,
+  params: RefactorDesc,
+  interactive: Boolean
+) extends RpcAnalyserRequest
+
+/**
  * Responds with a `RefactorFailure` or a `RefactorEffect`.
  */
 case class ExecRefactorReq(procId: Int, tpe: RefactorType) extends RpcAnalyserRequest
--- a/core/src/main/scala/org/ensime/util/FileUtil.scala
+++ b/core/src/main/scala/org/ensime/util/FileUtil.scala
@@ -82,4 +82,28 @@ object FileUtils {
     }
   }
 
+  def writeDiffChanges(changes: List[FileEdit], cs: Charset): Either[Exception, File] = {
+    //TODO: add support for NewFile and DeleteFile
+    val editsByFile = changes.collect { case ed: TextEdit => ed }.groupBy(_.file)
+    try {
+      val diffContents =
+        editsByFile.map {
+          case (file, fileChanges) =>
+            readFile(file, cs) match {
+              case Right(contents) =>
+                FileEditHelper.diffFromTextEdits(fileChanges, contents, file, file)
+              case Left(e) => throw e
+            }
+        }.mkString("\n")
+
+      Right({
+        val diffFile = java.io.File.createTempFile("ensime-diff-", ".tmp").canon
+        diffFile.writeString(diffContents)
+        diffFile
+      })
+    } catch {
+      case e: Exception => Left(e)
+    }
+  }
+
 }
--- a/project/EnsimeBuild.scala
+++ b/project/EnsimeBuild.scala
@@ -312,7 +312,8 @@ object EnsimeBuild extends Build with JdkResolver {
       "com.typesafe.akka" %% "akka-slf4j" % akkaVersion,
       "org.scala-refactoring" %% "org.scala-refactoring.library" % "0.8.0",
       "commons-lang" % "commons-lang" % "2.6",
-      "commons-io" % "commons-io" % "2.4" % "test,it"
+      "commons-io" % "commons-io" % "2.4" % "test,it",
+      "com.googlecode.java-diff-utils" % "diffutils" % "1.3.0"
     ) ++ logback ++ testLibs(scalaVersion.value, "it,test")
   )
 
--- a/swank/src/main/scala/org/ensime/server/protocol/swank/SwankFormats.scala
+++ b/swank/src/main/scala/org/ensime/server/protocol/swank/SwankFormats.scala
@@ -442,6 +442,7 @@ object SwankProtocolResponse {
   }
   // must be after FileEditFormat
   implicit val RefactorEffectFormat = SexpFormat[RefactorEffect]
+  implicit val RefactorDiffEffectFormat = SexpFormat[RefactorDiffEffect]
 
   // must be after SourcePosition
   implicit val TypeSearchResultFormat = SexpFormat[TypeSearchResult]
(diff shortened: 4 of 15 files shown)


fommil:
Oh wow, awesome! Is this the continuation of your work from the hack day? I'll try to find some time this weekend to review this properly.

lxol:
Yes, the hack day was a great jump-start into ensime. Thank you!

api/src/main/scala/org/ensime/api/incoming.scala
```diff
@@ -236,6 +236,16 @@ case class PrepareRefactorReq(
 ) extends RpcAnalyserRequest
 
 /**
+ * Responds with a `RefactorFailure` or a `RefactorDiffEffect`.
+ */
+case class DiffRefactorReq(
+  procId: Int,
+  tpe: Symbol, // tpe is ignored but part of the legacy wire format
```
fommil: if you're adding new API messages you don't need to satisfy the legacy format

api/src/main/scala/org/ensime/api/incoming.scala
```diff
@@ -236,6 +236,16 @@ case class PrepareRefactorReq(
...
+  params: RefactorDesc,
```
fommil: I'd like to see us come up with a cleaner way of doing the refactor params. This legacy object is quite horrible.

project/EnsimeBuild.scala
```diff
@@ -312,7 +312,8 @@ object EnsimeBuild extends Build with JdkResolver {
       "com.typesafe.akka" %% "akka-slf4j" % akkaVersion,
       "org.scala-refactoring" %% "org.scala-refactoring.library" % "0.8.0",
       "commons-lang" % "commons-lang" % "2.6",
-      "commons-io" % "commons-io" % "2.4" % "test,it"
+      "commons-io" % "commons-io" % "2.4" % "test,it",
+      "com.googlecode.java-diff-utils" % "diffutils" % "1.3.0"
```
fommil: great find!

swank/src/main/scala/org/ensime/server/protocol/swank/SwankFormats.scala
```diff
@@ -442,6 +442,7 @@ object SwankProtocolResponse {
   }
   // must be after FileEditFormat
   implicit val RefactorEffectFormat = SexpFormat[RefactorEffect]
+  implicit val RefactorDiffEffectFormat = SexpFormat[RefactorDiffEffect]
```
fommil: I'm impressed that you worked out the changes needed in this file :smile: it's really quite horrible. I'd love to just delete the whole thing, but it requires a big migration on the emacs side.

api/src/main/scala/org/ensime/api/incoming.scala
fommil: ok, I see now that this gives us a really easy migration path onto diffs, which is the most important thing.

core/src/main/scala/org/ensime/util/FileUtil.scala
```diff
@@ -82,4 +82,28 @@ object FileUtils {
...
+  def writeDiffChanges(changes: List[FileEdit], cs: Charset): Either[Exception, File] = {
+    //TODO: add support for NewFile and DeleteFile
+    val editsByFile = changes.collect { case ed: TextEdit => ed }.groupBy(_.file)
+    try {
+      val diffContents =
+        editsByFile.map {
+          case (file, fileChanges) =>
+            readFile(file, cs) match {
+              case Right(contents) =>
+                FileEditHelper.diffFromTextEdits(fileChanges, contents, "/dev/null", file.getAbsolutePath())
```
fommil: I'm surprised that this passes on the Windows CI

fommil:
ok, this is really great stuff, thank you so much! Can you please:

1. remove the redundant fields in the incoming/outgoing
2. change the API so that we don't need to pass this `"/dev/null"` string (input/output name could be the same)
3. squash into one commit

and this looks like it's then ready for merge!

aemoncannon:
Awesome work. What do you think about PrepareRefactorReqV2 over DiffRefactorReq? Ditto for RefactorResultV2. Would make the evolution of the protocol easier to see, and maybe could adopt this approach in general. It's good enough for Sqlite :-)

aemoncannon:
Also please bump the protocol version and add an entry in the protocol change log.

fommil:
ah, yes, forgot about that. it's in the `swank` folder if you're interested.

lxol:
@aemoncannon PrepareRefactorReqV2 is supposed to replace PrepareRefactorReq ExecRefactorRec and CancelRefactorReq. Don't you thinks that we should remove "Prepare" prefix?

lxol:
I will make the changes tomorrow. 
Thank you for the review

aemoncannon:
@lxol I see. So the clients are expected to apply the diff -- is that correct? In that case, maybe just 'RefactorReq' , and 'RefactorDiffResult'. What do you think?

aemoncannon:
Er, RefactorDiffEffect

lxol:
@aemoncannon Yes, the editor will be responsible for the edits and undos. 

I will rename DiffRefactorReq to RefactorReq as you suggested

fommil:
can you please force push onto this PR with your changes?

lxol:
done

fommil:
This is a really top notch PR, thank you! Fancy having a look at moving over Emacs?

lxol:
@fommil I already used Emacs to send/receive diff request/responses using a new api. I will try to make  something production ready. Not sure about ETA though.

fommil:
:grinning:

aemoncannon:
Huzzah!

fommil:
@ensime/vim-panel @ensime/atom-panel @ensime/sublime-panel this PR introduces a much better way to perform refactorings. We'd like to move all clients over to this new (simpler) API, can I bother you all to look at doing this? I'd like to remove the legacy APIs in the 1.0 release (yeah yeah, I know its 10 months overdue... but upgrading to Java 7 is becoming more of a priority so it might happen).

It's important as part of our "ENSIME server shouldn't mutate anything" strategy.

davegurnell:
Sorry for the delayed reply... better late than never!

In general this looks great to me. IIUC these are Git-compatible diffs?

Dave

lxol:
@davegurnell the diff file is in a unified diff format and git-apply can use it. I've just tested git-apply and noticed several issued. 
This command complains that the diff format is wrong.
```` sh
git apply -p0 --unsafe-path /private/var/folders/hc/lg9jzhmn5zj4kp4ldl2hfmk00000gn/T/ensime-diff-1519913091103607566.tmp
````
However if I add an empty line to the end of the diff file, git-apply  works perfectly. I will submit a PR to fix that.
Another issue is that using an external command to apply the patch is no different from applying the changes from the ensime-server. We loose the benefit of keeping the edit history.  We are lucky with `diff-mode` in Emacs.

davegurnell:
Ok awesome. I realised reading your code that I don't know enough about
diffs. I wasn't sure what libraries to use to interpret them. What you've
said there sounds great---a universal or git-compatible diff format is
bound to have support in all the various client languages.

PR #1181, opened 2015-12-17
Lookup definition jumps to function definition

VlachJosef:
Make Lookup definition jump to definition of a function
instead of apply method in FunctionX trait

### DIFF
--- a/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
@@ -126,6 +126,23 @@ class RichPresentationCompilerSpec extends WordSpec with Matchers
         }
     }
 
+    "jump to function definition instead of Function1.apply" in withPresCompiler { (config, cc) =>
+      import ReallyRichPresentationCompilerFixture._
+      runForPositionInCompiledSource(config, cc,
+        "package com.example",
+        "object Bla { val fn: String => Int = str => str.lenght }",
+        "object Abc { def main { Bla.f@@n(\"bal\" } }") { (p, label, cc) =>
+          val sym = cc.askSymbolInfoAt(p).get
+          assert(sym.name == "fn")
+          assert(sym.localName == "fn")
+          assert(sym.declPos.isDefined)
+          assert(sym.declPos.get match {
+            case OffsetSourcePosition(f, i) => i > 0
+            case _ => false
+          })
+        }
+    }
+
     "get symbol info by name" in withPresCompiler { (config, cc) =>
       def verify(
         fqn: String, member: Option[String], sig: Option[String], expectedLocalName: String,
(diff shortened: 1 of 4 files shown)


fommil:
oh awesome! I've been waiting for this one :smile: a bit too late right now but I'll definitely have a good review of this at the weekend.

This technically breaks the API (because we're not actually returning information about the thing at point) but I consider `FunctionX` to be a language implementation detail and therefore we should never be seeing it.

fommil:
actually it was so short it is easy to see what it's doing. Great stuff! Is it possible to write a standard RichPresentationCompilerSpec for this?

VlachJosef:
You right about breaking API, both eclipse and intellij idea give you an option where you wants to jump, but frankly I never wants to jump to `FunctionX.apply` method. A consider this breaking API change as a big productivity booster as it streamline workflow. In the case someone wants to jump into `FunctionX.apply` method, the easy way is just to add `.apply` to a function definition, so in a sense there is no functionality hidden from users. This behaviour should probably be described in a documentation. Where it would be best to put it?

VlachJosef:
I'll try to write spec in RichPresentationCompilerSpec over the weekend.

fommil:
if the user really wants to look at `Function2.apply` they can find it with `ensime-search`. Does this close any tickets?

VlachJosef:
No, I doesn't close any ticket.

fommil:
hmm, interesting... looks like you're getting failed tests on Linux, but not Windows. I wonder if the test is sensitive to the version of Java?

```
[info]   java.lang.AssertionError: assertion failed: expected:  but got unexpected message Some(SymbolInfo(apply,apply,Some(OffsetSourcePosition(/tmp/1450396448450-0/.ensime_cache/dep-src/source-jars/scala/Function1.scala,1665)),ArrowTypeInfo((v1: T1)R,136,BasicTypeInfo(R,137,Nil,scala.R,List(),List(),None,None),List(ParamSectionInfo(List((v1,BasicTypeInfo(T1,135,Nil,scala.T1,List(),List(),None,None))),false))),true,Some(138)))
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsgPF(TestKit.scala:357)
[info]   at akka.testkit.TestKit.expectMsgPF(TestKit.scala:718)
[info]   at org.ensime.intg.BasicWorkflow$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$mcV$sp$2$$anonfun$apply$10$$anonfun$apply$11.apply(BasicWorkflow.scala:188)
```

VlachJosef:
Hm, I will change the tests. It looks like it is failing when checking offset in the `Function1.scala` file. Since this is external source checking offset is not good idea anyway.

VlachJosef:
Ok, I've added RichPresentationCompilerSpec spec and remove checking of offset in `Function1.scala` file. Tests should now pass.

fommil:
Great stuff! This will really improve things for functional libraries

PR #1182, opened 2015-12-18
Add RefactorReq and RefactorDiffEffect (#996)

lxol:


lxol:
changes to the previous pr:
* removed tpe field from RefactorReq
* removed status field from RefactorDiffEffect
* replaced /dev/null with then name of the file in diff chunks
* renamed DiffRefactorReq to RefactorReq
* bumped protocolVersion

fommil:
maybe the diff method should take `File`s instead of `String`s... hmm

lxol:
> maybe the diff method should take Files instead of Strings... hmm

in that case we have to save the  refactored contents as a file and compare it to the original file. I think it is more efficient to compare contents as lists of strings and avoid creating temporary files.

fommil:
@lxol I mean the names of the `File`s. No need to pass those as `String`

lxol:
I can see one use case when it is not convenient to have Files. If newContent is empty then compareContents generates a diff that, when applied, creates a new file and we have to provide a name of that non existent as a string argument to put it to a chunk header.

fommil:
files don't have to exist. You'd only be sending a dummy string anyway.

lxol:
ok, got it. I will submit an updated pr later today..

fommil:
btw, no need to submit new PRs when you want to make a change. Just force push onto the same branch and the PR gets automatically updated.

PR #1187, opened 2015-12-19
add file modification time to a diff header

lxol:
fix a bug in DiffUtil. 
`patch -p0 <diff ` is quite happy with a currently generated diff. However emacs `diff-mode` expects `tab` and  a timestamp after the file name in the diff header as per http://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html


fommil:
I think we had a transient failure. Weird.

PR #1188, opened 2015-12-20
Improved support for case class navigation

VlachJosef:
When symbol under cursor is member field of the case
class then function symbol under cursor (M-. or Control+Left-Click)
jumps to case class definition

### DIFF
--- a/core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
+++ b/core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
@@ -449,6 +449,11 @@ class RichPresentationCompiler(
       && !sym.nameString.contains("$"))
   }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -449,6 +449,11 @@ class RichPresentationCompiler(
       && !sym.nameString.contains("$"))
   }
 
+  private def noDefinitionFound(tree: Tree) = {
+    logger.warn("No definition found. Please report to https://github.com/ensime/ensime-server/issues/492 with description of what did you expected. symbolAt for " + tree.getClass + ": " + tree)
+    Nil
```
VlachJosef: I've improved this method a little with adding more description what is happening

PR #1192, opened 2015-12-22
Ensure that repeated ConnectionInfoReq calls do not cause error

rorygraves:
A recent updated meant that if ConnectionInfo was requested more than once it resulted in a server side error.
When using WebSockets you need to send periodic messages to avoid the WebSocket timing out, and using a NoOp request such as ConnectionInfo makes sense.


rorygraves:
Merging as trivial.

Issue #471, opened 2014-07-01
Summarise file (aka Structure View)


fommil:
it would be great to be able to get a view that shows a summary of the current file and give the ability to quickly jump to classes, traits and members.

Effectively an emacs version of IntelliJ's "Structure View".

edani:
@IvanMalison's PR was added to scala-mode2 a while ago. It does a pretty good job considering that it relies on scala-mode2's syntax parser. Here is a way to see how it looks   (make sure you update to the latest scala-mode2)

```elisp
(require 'scala-mode2)
(add-hook 'scala-mode-hook #'imenu-add-menubar-index)
(require 'speedbar)
(speedbar-add-supported-extension ".scala")
(setq scala-imenu:should-flatten-index nil)
(setq imenu-auto-rescan t)
(speedbar)
(menu-bar-mode 1)
```

It shows the scala file's structure in the "Index" menu a well as the speedbar. Speedbar has a few problems:

- it doesn't auto-refresh
- when refreshed manually, the structure is collapsed again.
- at the moment, trying to expand all descendants causes an infinite loop.

But these problems aren't exactly Ensime's

IvanMalison:
I meant to try to get around to implementing this in ensime, but it never really happened. How difficult would it be to get source code location information from the compiler frontend.  If it's something that could be done relatively quickly, I'd be happy to take a stab at it. I did write a tiny bit of code to grab type information from ensime, but it probably isn't worth including in ensime:
https://github.com/IvanMalison/.emacs.d/blob/3fd9acd615b819fa8f7f34f0d768ba7898458fb6/lisp/ensime-imenu.el

> at the moment, trying to expand all descendants causes an infinite loop.

what do you mean by this?

I actually use the stuff in scala-mode2 with helm-imenu rather than speedbar.

rorygraves:
@IvanMalison The source locations should not be hard to get from Ensime - I think the info might already be available via ``` inspect-type-by-name``` ,  ``` inspect-type-by-id ``` and ``` inspect-type-at-point ``` Ensime calls.

edani:
I don't think it's hard,  SemanticHighlighting.scala already does something similar, maybe you could 
make small changes to it. You'd get the whole structure in one call.

Right now it takes the scala parse tree and returns information about specific leaf nodes but classes/methods/etc .definitions are in the same structure.

[...]

PR #1196, opened 2015-12-31
Structure view (summarise file) backend

ancane:
This is my attempt on structure view backend for #471.
It should be good enough to start with emacs imenu generators. 
But I'm still concerned with couple things:
* Is it ok to have recursive StructureViewMember type?
* Not sure if I did tree traversal correctly.

Thanks in advance for any feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,3 @@ target/
 .sbtopts
 
 TAGS
-
-*#
-.#*
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ target/
 .sbtopts
 
 TAGS
+
+*#
+.#*
--- a/core/src/it/scala/org/ensime/core/StructureViewBuilderSpec.scala
+++ b/core/src/it/scala/org/ensime/core/StructureViewBuilderSpec.scala
@@ -1,0 +1,110 @@
+package org.ensime.core
+
+import org.ensime.fixture._
+import org.ensime.api._
+import org.scalatest._
+import scala.collection.mutable.ListBuffer
+
+class StructureViewBuilderSpec extends WordSpec with Matchers
+    with IsolatedRichPresentationCompilerFixture
+    with RichPresentationCompilerTestUtils
+    with ReallyRichPresentationCompilerFixture {
+
+  def original = EnsimeConfigFixture.EmptyTestProject
+
+  "StructureViewBuilder" should {
+
+    def getStructure(
+      config: EnsimeConfig,
+      cc: RichCompilerControl,
+      content: String
+    ): List[String] = {
+
+      val result = ListBuffer[String]()
+
+      def collect(parent: Option[String], x: StructureViewMember): Unit = {
+        val par = parent.map(_ + ".").getOrElse("")
+        x match {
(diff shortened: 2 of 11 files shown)


fommil:
Great! I'll try to have a close look later today.

.gitignore
```diff
@@ -6,3 +6,6 @@ target/
 .sbtopts
 
 TAGS
+
+*#
```
fommil: please add these to your  `~/.gitignore` (global for you)

jerk/src/main/scala/org/ensime/jerk/JerkFormats.scala
```diff
@@ -41,8 +41,9 @@ private object JerkConversions extends DefaultJsonProtocol with FamilyFormats {
   override implicit def coproductHint[T: Typeable]: CoproductHint[T] = new FlatCoproductHint[T]("typehint")
 
   // note that cachedImplicit increased the compile time
-  val RpcRequestFormat = RootJsonFormat[RpcRequest]
-  val EnsimeServerMessageFormat = RootJsonFormat[EnsimeServerMessage]
+  val RpcRequestFormat = rootFormat(lazyFormat(RootJsonFormat[RpcRequest]))
```
fommil: this is going to confuse the shapeless type inference, can you undo this please?

fommil:
ok, this is looking good! Just a couple of comments.

I've started a fresh build as something went wrong with drone.

fommil:
hmm, drone is not fixing itself. It'll run again when you push changes (please squash)

core/src/it/scala/org/ensime/core/StructureViewBuilderSpec.scala
```diff
@@ -0,0 +1,110 @@
+package org.ensime.core
+
+import org.ensime.fixture._
+import org.ensime.api._
+import org.scalatest._
+import scala.collection.mutable.ListBuffer
+
+class StructureViewBuilderSpec extends WordSpec with Matchers
```
fommil: can you please use `FlatSpec` and the `shouldBe` pattern? We're trying to move away from `WordSpec` and `assert`

core/src/main/scala/org/ensime/core/StructureViewBuilder.scala
```diff
@@ -0,0 +1,84 @@
...
+
+import org.slf4j.LoggerFactory
+
+import scala.collection.mutable.ListBuffer
+import scala.reflect.io.AbstractFile
+import scala.tools.refactoring.common.{ CompilerAccess, PimpedTrees }
+import scala.reflect.internal.util.SourceFile
+import org.ensime.api._
+
+class StructureViewBuilder(val global: RichPresentationCompiler)
```
fommil: the pattern we've been following to date has been to mix in, rather than take as a constructor parameter. Can you please require a mixin from `RichPresentationCompiler`?

ancane:
@fommil Drone doesn't seem to work :( Same error.

fommil:
that's really weird. Try closing this PR and starting a fresh one, we might have hit an internal github hash collision or something.

PR #1204, opened 2016-01-03
Add indirection to FileWatchers.

aemoncannon:
Setting the stage for an optional Java 7 based watcher.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
+++ b/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
@@ -9,21 +9,20 @@ import org.apache.commons.vfs2.impl._
 
 import org.ensime.api._
 
+import scala.collection.Set
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 import scala.util.Try
 import org.ensime.util.file._
 
-trait ClassfileListener {
-  def classfileAdded(f: FileObject): Unit
-  def classfileRemoved(f: FileObject): Unit
-  def classfileChanged(f: FileObject): Unit
+trait FileChangeListener {
+  def fileAdded(f: FileObject): Unit
+  def fileRemoved(f: FileObject): Unit
+  def fileChanged(f: FileObject): Unit
 }
 
-trait SourceListener {
-  def sourceAdded(f: FileObject): Unit
-  def sourceRemoved(f: FileObject): Unit
-  def sourceChanged(f: FileObject): Unit
+trait Watcher {
+  def shutdown(): Unit
 }
 
 /**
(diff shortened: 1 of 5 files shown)


fommil:
nice! Hopefully one of us will get there over the coming weeks.

I can set up an ensime-docker 2.x image with OpenJDK 7 on it, so we can build a standalone project that implements the interface. Ideally, matching the VFS2 API too, but not necessarily. I'd also like to fix the fundamental bugs in the VFS file watcher implementation anyway.

BTW, you failed the formatting test.

fommil:
sorry I think I restarted the servers while this build was going on

fommil:
I'll have a look at this in the next day or two, been a busy start to the year.

fommil:
I'll review this first thing in the morning, getting the file watchers working on big windows projects is going to really help me

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -14,16 +14,14 @@ import scala.concurrent.Future
 import scala.util.Try
 import org.ensime.util.file._
 
-trait ClassfileListener {
-  def classfileAdded(f: FileObject): Unit
-  def classfileRemoved(f: FileObject): Unit
-  def classfileChanged(f: FileObject): Unit
+trait FileChangeListener {
```
fommil: I should never have created two traits in the first place.

fommil: maybe I was trying to give myself the false illusion of type safety.

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -79,10 +144,10 @@ class ClassfileWatcher(
         running = scheduler.schedule(resetter, 1, TimeUnit.SECONDS)
       }
 
-      private val targets: Set[FileName] = for {
-        dir <- config.targetClasspath
+      private val targets: Set[FileName] = (for {
+        dir <- watchedDirs
         ref = vfs.vfile(dir)
-      } yield ref.getName()
+      } yield ref.getName()).toSet
```
fommil: it should be slightly more efficient to take a Set[File] as watchedDirs (avoids an intermediate collection)

fommil:
great, LGTM! If you could just change the `watchedDirs` to a `Set[File]` that will give us a small performance boost, then just merge when it turns green.

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
rorygraves: Breakout......  Does nobody watch my talks :(

fommil: no need! It is actually a `Set[File]` to begin with, but the type sig is `Iterable[Set]` so the compiler doesn't know that.

fommil:
the best thing about this PR is that @aemoncannon's name is going to be all over the file watcher code now, not mine, so the evidence has been destroyed! :laughing:

rorygraves:
You have been marked, it will always be your fault ;)

fommil:
nope! look at the blame log. It's A E M O N . Not Sam, Aemon. It's obviously all his fault. I would never write such shoddy file watcher code. (squashes history, force pushes...)

PR #1205, opened 2016-01-04
Don't insert fqn when completing index matches.

aemoncannon:
A small usability fix. When selecting a completion candidate provided by the indexer, (always a class fqn atm), insert the unqualified name into the buffer, instead of the fully qualified name. User can follow up with ensime-import-type-at-point if they're missing the import.

TODO: send more detailed indexer completion results. For example: constructors with all param info if the user types 'new Foo...'

### DIFF
--- a/core/src/main/scala/org/ensime/core/Completion.scala
+++ b/core/src/main/scala/org/ensime/core/Completion.scala
@@ -167,7 +167,7 @@ trait CompletionControl {
         s.syms.map { s =>
           CompletionInfo(
             s.localName, CompletionSignature(List.empty, s.name),
-            -1, isCallable = false, 40, Some(s.name)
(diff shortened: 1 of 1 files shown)


fommil:
Wasn't somebody doing something client side to make this nicer?

fommil:
yeah, have a look here https://github.com/e8kor/ensime-emacs/commit/695e734c1170a8bbcac55cc7fb96f36b4d5c1f69

does that achieve the same thing? Might be best to do this client side

fommil:
@aemoncannon what do you reckon, should we hold off for https://github.com/ensime/ensime-emacs/issues/166 ?

aemoncannon:
Hmm, that commit is abusing the to-insert field. I think we want my commit,
and modify the elisp one to use the completion signature.

PR #1206, opened 2016-01-05
Add type search results to java completion.

aemoncannon:
Just bringing java completion closer to parity with Scala. Now both will benefit when the indexer can generate better completion results : )


fommil:
niiiiiiiiiiiiice!

Issue #673, opened 2014-10-23
I BET IT IS THIS ONE => red squiggles on valid code <= I BET IT IS THIS ONE


fommil:
This is a gather-all ticket.

Basically we are relying on the scala compiler's interactive mode, also known as the presentation compiler. It is known to have loads of corner case problems and unfortunately these manifest themselves as false positives in ENSIME.

The best thing you can do is report a ticket at https://issues.scala-lang.org/ with a minimal failure case as per http://scala-ide.org/docs/dev/testing/presentation-compiler-tests.html or consider investigating further and submitting a fix via https://github.com/typelevel/scala

If the problem arises from a macro or compiler plugin (e.g. shapeless), you may wish to get involved with the workaround discussed at https://github.com/fommil/imaginary-friend

PR #1210, opened 2016-01-07
False positive Unused import warning crash

VlachJosef:
In some cases when macro is in use, presentation compiler issues false positive of
Unused import warning. Parameter `pos` has in this case field `point` set to -1, what
cause `ArrayIndexOutOfBoundsException` when trying to get `pos.column`. This
fix detect this situation and return 0 instead of `pos.column`

### DIFF
--- a/core/src/main/scala/org/ensime/util/Reporter.scala
+++ b/core/src/main/scala/org/ensime/util/Reporter.scala
@@ -40,6 +40,12 @@ class PresentationReporter(handler: ReportHandler) extends Reporter with Positio
           if (pos.isDefined) {
             val source = pos.source
             val f = source.file.absolute.path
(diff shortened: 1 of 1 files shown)


aemoncannon:
Nice!

VlachJosef:
This is fixing crash in `Reporter.scala` and preventing typechecking of affected files, but on the other hand what remains is another issue worth mentioning in https://github.com/ensime/ensime-server/issues/673

VlachJosef:
Even `scala.tools.nsc.reporters.ConsoleReporter` suffers with the same issue https://github.com/scala/pickling/issues/370

fommil:
no brainer :smile: hope to see you on the 16th!

Issue #345, opened 2014-04-26
ENJINE: java mode


fommil:
aka ENhanced Java INteraction mode for Emacs ;-P

Many of the features of ENSIME could be made available for Java source code editing. Specifically:

* jump to source of symbol at point
* method completion
* package navigation
* classname completion

On the emacs side, this would involve creating an `ensime-java-mode` (or `enjine-mode`) that enables the functions that are to be used, which sends requests to the server. Many of the `ensime-mode` features can be re-used, such as popup/autocomplete boxes and the SWANK protocol (which might not need to be changed).

On the server side would be significant changes, as we would need to use the java compiler in order to produce a [javac.tree](http://docs.oracle.com/javase/6/docs/jdk/api/javac/tree/) that would be used for queries related to `.java` files, instead of the presentation compiler (for `.scala` files).

This might not be as hard as it sounds, since the java compiler doesn't need to perform type inference it can treat each file in isolation and just parse it from fresh every time.

This would pave the way forwards for integration with static analysis tools that could offer hints or automated refactorings, and spring support.

aemoncannon:
I just started working with a VeryLargeJavaCodeBase, so I might make some movements in this direction. My tentative plan is to make minimal adjustments to Ensime to support (probably in this order):
1 navigation
2 inspection
3 completion
As you say, 1,2,3 all require a typed java tree, so the big question is where to get that. I previously integrated with eclipse jdt for java error highlighting, but that has since been ripped out during various refactorings. Might try that direction again.
Other goals:
* don't muddy up the scala story too much
* support VeryLarge projects. will probably borrow from our scala approach here (don't eagerly load sources)

rorygraves:
@fommil Wanted to work on this at some point incase he was made to go back to a VLJCB. 

I definitely think ti would be good to support scala/java mixed projects (and by extension java projects).
The protocol could be made either more agnostic or into language agnostic and language specific subsections quite easily.

I suspect @fommil  has a lot to say on this one - I know he has some ideas about the AST.  

Getting the basics up and working shouldn't be too silly and we support a lot of the infrastrure you will need.

aemoncannon:
@rory Yep. Most of the UI in emacs will work fine. The debugger should
require zero work. I'm guessing the indexer will be easy as well.

rorygraves:
btw, I know there has been some radio silence from me recently - I've actually been improving the tests of the debugger (found a couple of bugs/problems which I'm working on).  PR soon.

[...]

PR #1212, opened 2016-01-08
ENJINE: Implement show error at point

myrjola:
Related to https://github.com/ensime/ensime-server/issues/345.

The function `ensime-print-errors-at-point` wasn't working on ensime-emacs for Java-files. I made a stab at implementing that support. Unfortunately I was unable to get the it-tests working locally with `sbt it:test`, so we'll have to see what the CI server says about my code.

### DIFF
--- a/core/src/main/scala/org/ensime/core/JavaAnalyzer.scala
+++ b/core/src/main/scala/org/ensime/core/JavaAnalyzer.scala
@@ -74,6 +74,10 @@ class JavaAnalyzer(
     case SymbolAtPointReq(file, point) =>
       sender() ! javaCompiler.askSymbolAtPoint(file, point)
 
+    case ImplicitInfoReq(file, range: OffsetRange) =>
+      // Implicit type conversion information is not applicable for Java, so we
+      // return an empty list.
+      sender() ! ImplicitInfos(Nil)
   }
 
 }


aemoncannon:
Hrmmm. I'm thinking we should just special case this in emacs. There's no such thing as implicits in java, so no need to ask for them, right? All the error/warning notes should already be present in client.

myrjola:
Looking at the ensime-emacs code again I noticed my blunder. The compilation errors were actually already stored in Emacs, but the `ensime-print-errors-at-point`-call hangs because fetching of implicit type conversion information isn't implemented or even applicable for Java-files.

@aemoncannon what would be the best way to fix this? I can think of two ways:
1. Fix ensime-emacs to not fetch implicit type conversion information for Java files
2. Make JavaAnalyzer return an empty response for `ImplicitInfoReq` so that clients requesting that wouldn't hang

fommil:
2c. The Emacs client will eventually be refactored so that ensime-mode and enjime-mode are real things. However, the current Java support is a hack to get us into a place where we can perform that refactoring, but giving us real functionality. Therefore, the server should probably implement this. For now.

fommil:
Tests failed due to formatting.

myrjola:
So I tried to implement alternative "2. Make JavaAnalyzer return some kind of empty response for ImplicitInfoReq so that clients requesting that wouldn't hang". Somehow it doesn't work and I could use some help in figuring out why. I updated the PR with the code I think should solve the problem. Calling `ensime-print-errors-at-point` in ensime-emacs hangs and after quitting that operation I get the following log output:

```
01:34:07.631 DEBUG akka://ENSIME/user/ensime-main/tcp-server/con1/5 o.e.s.RequestHandler - ImplicitInfoReq(Left(/Users/martin/Downloads/gradle-2.10/samples/java/quickstart/src/main/java/org/gradle/Person.java),OffsetRange(164,188))
01:34:07.632 DEBUG akka://ENSIME/user/ensime-main/project o.e.c.Project - received handled message ImplicitInfoReq(Left(/Users/martin/Downloads/gradle-2.10/samples/java/quickstart/src/main/java/org/gradle/Person.java),OffsetRange(164,188)) in state handleRequests
01:34:07.632 INFO  akka://ENSIME/deadLetters a.a.DeadLetterActorRef - Message [org.ensime.api.ImplicitInfoReq] from Actor[akka://ENSIME/user/ensime-main/tcp-server/con1/5#-1759615625] to Actor[akka://ENSIME/deadLetters] was not delivered. [1] dead letters encountered. This logging can be turned off or adjusted with con
```

So my questions are:
1. Is it correct to return an empty `ImplicitInfos`-instance for an `ImplicitInfoReq` request to JavaAnalyzer? Would a `VoidResponse` also be ok?
2. What are "dead letters"? It seems that ensime-server manages to handle the request, but something is wrong.
3. How would I go about writing a test to ensure that an `ImplicitInfoReq` to JavaAnalyzer returns an empty ImplicitInfos-instance?

fommil:
hmm, I don't understand why dead letters is getting the response. That's usually a sign that an actor has died. I'll have a closer look in the morning. Sorry, wrong TZ for me.

fommil:
cool, I just realised you're maintaining `diminish`. I just started using `use-package` and this is a really nice feature.

myrjola:
@fommil, thanks for the kind words. I also stumbled upon `diminish` through `use-package`. I became a maintainer as the result of posting a patch. Hopefully this doesn't happen in every project I contribute :wink:.

fommil:
muahahahaha!

myrjola:
The build for Scala 2.10.6 failed because of a timeout. Last time the build for Scala 2.11.7 failed with the same error. I'm only making a minor change, so I think I haven't broken the build. Are the timeouts a known problem?

fommil:
I've been doing some things with the CI recently, I might have done something. Restarting.

fommil:
@myrjola ignore the clahub thing, I changed us to the licence/cla one... can you please sign it.

@aemoncannon other than the CLA, this is green if you're happy with it.

aemoncannon:
lgtm!

myrjola:
I'll make sure I have the paper work signed. I think this can be merged when I have tests implemented and I can get the fix to work in ensime-emacs. @aemoncannon I asked some questions in https://github.com/ensime/ensime-server/pull/1212#issuecomment-170159696, I'll post them below for convenience:

1. Is it correct to return an empty ImplicitInfos-instance for an ImplicitInfoReq request to JavaAnalyzer? Would a VoidResponse also be ok?
2. What are "dead letters"? It seems that ensime-server manages to handle the request, but something is wrong.
3. How would I go about writing a test to ensure that an ImplicitInfoReq to JavaAnalyzer returns an empty ImplicitInfos-instance? Is it better to write a test in the ensime-emacs repository to reproduce my issue?

fommil:
No paperwork, just click the licence/cla link above and you can press a button.

myrjola:
I have signed the CLA. Props to you guys for creating such a simple process!

fommil:
1. Empty is better unless it's defined in the API.
2. Not sure.
3. Look under core/it for integration tests of this sort. There should be a javacompilerspec or something like it.

We'll wait until you've added the test.

fommil:
The tests can wait, this is a fairly easy commit to understand and sounds like it helps out java devs.

PR #1213, opened 2016-01-08
Improved navigation in source files

VlachJosef:
Improvements in RichPresentationCompiler. Method `symbolAt(pos:
Position)` now copy behaviour from Eclipse even more than before.

If point now allow jump to method apply, or some value, we will jump
to value, since it is more reasonable alternative.

### DIFF
--- a/core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
+++ b/core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
@@ -478,32 +478,14 @@ class RichPresentationCompiler(
           }
         case Annotated(atp, _) =>
           List(atp.symbol)
-        case st: SymTree =>
+        case ap @ Select(qualifier, nme.apply) =>
+          // If we would like to give user choice if to go to method apply or value
+          // like Eclipse is doing we would need to return:
+          // List(qualifier.symbol, ap.symbol)
+          List(qualifier.symbol)
+        case st if st.symbol ne null =>
           logger.debug("using symbol of " + tree.getClass + " tree")
-          val treeSymbol =
-            st match {
-              case Select(qualifier, name) if name.decoded == "apply" && qualifier.pos.includes(pos) =>
-                specificOwnerOfSymbolAt(pos) match {
-                  case Some(ownerSymbol) if ownerSymbol.fullName.startsWith("scala.Function") => qualifier.symbol
-                  case _ => tree.symbol
-                }
-              case _ => tree.symbol
-            }
-          List(treeSymbol)
-        case applyTree @ Apply(Select(qualifier, sym), args) =>
-          sym.decoded match {
-            case "apply" =>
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -478,32 +478,14 @@ class RichPresentationCompiler(
           }
         case Annotated(atp, _) =>
           List(atp.symbol)
-        case st: SymTree =>
+        case ap @ Select(qualifier, nme.apply) =>
+          // If we would like to give user choice if to go to method apply or value
+          // like Eclipse is doing we would need to return:
+          // List(qualifier.symbol, ap.symbol)
+          List(qualifier.symbol)
+        case st if st.symbol ne null =>
```
VlachJosef: This is taken from [scala-ide ](https://github.com/scala-ide/scala-ide/blob/b2cb96d3d6735a4eb243a35b09e7835c8f8f8e18/org.scala-ide.sdt.core/src/org/scalaide/core/internal/hyperlink/ScalaDeclarationHyperlinkComputer.scala#L54)

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -478,32 +478,14 @@ class RichPresentationCompiler(
           }
         case Annotated(atp, _) =>
           List(atp.symbol)
-        case st: SymTree =>
+        case ap @ Select(qualifier, nme.apply) =>
```
VlachJosef: This is taken from [scala-ide ](https://github.com/scala-ide/scala-ide/blob/b2cb96d3d6735a4eb243a35b09e7835c8f8f8e18/org.scala-ide.sdt.core/src/org/scalaide/core/internal/hyperlink/ScalaDeclarationHyperlinkComputer.scala#L53)

testing/simple/src/main/scala/org/example/Foo.scala
```diff
@@ -18,15 +18,10 @@ object Foo extends App {
...
+  fn("foobar")
 }
 
 // for SearchServiceSpec
 case class CaseClassWithCamelCaseName()
 case class Bloo()
 case object Blue
-
-object Baz {
-  val fn: String => Int = str => str.size
```
fommil: I assume we were never testing this anyway

fommil:
great stuff!

testing/simple/src/main/scala/org/example/Foo.scala
VlachJosef: This was replaced by [this](https://github.com/ensime/ensime-server/blob/bc4ad93e2fc533f11ead04a66c6d7d476862dd5b/testing/simple/src/main/scala/org/example/Foo.scala#L19-L21).

PR #1224, opened 2016-01-14
fix eof bug in the diff file

lxol:
diff file should ends with an empty line.


fommil:
@dickwall this one goes out to you

Thanks @lxol !

PR #1225, opened 2016-01-15
Banish type-ids to the land of wind and ghosts

aemoncannon:
The emacs client only uses CallCompletionInfo at this point, and I believe I can get rid of that by making small tweaks to CompletionSignature, which is returned with the CompletionInfo objects. In the mean time, let's get all the client teams on the same page (my guess is not many folks use this stuff..).
@rorygraves @fommil

### DIFF
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -143,16 +143,6 @@ case class PackageMemberCompletionReq(
 ) extends RpcAnalyserRequest
 
 /**
- * Responds with `CallCompletionInfo` if valid, or `FalseResponse`.
- */
-case class CallCompletionReq(id: Int) extends RpcAnalyserRequest
-
-/**
- * Responds with `TypeInfo` if valid, or `FalseResponse`.
- */
-case class TypeByIdReq(id: Int) extends RpcAnalyserRequest
-
-/**
  * Responds with `TypeInfo` if valid, or `FalseResponse`.
  */
 case class TypeByNameReq(name: String) extends RpcAnalyserRequest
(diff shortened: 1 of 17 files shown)


fommil:
Great! This is something we've wanted to do for a while

aemoncannon:
@rorygraves @yazgoo @hedefalk
Just a heads up. In the next few days I hope to have a PR to remove these deprecated items from the protocol. Please reply here if you have any concerns.

fommil:
I am extremely concerned by Donald Trump.

aemoncannon:
Don't worry. I'm pretty sure he is just a conceptual art project.

fommil:
curse your infernal documentation formatting!

fommil:
@aemoncannon btw, incase you didn't catch it... this failed because of bad formatting in `b/api/src/main/scala/org/ensime/api/incoming.scala`

rorygraves:
@aemoncannon Sublime does not use any of the deprecated calls, but one of the deprecated types -TypeInfo (specifically ArrowTypeInfo) uses ParamSectionInfo, so that can't go get.

rorygraves:
Whoops, sorry, meant Comment not 'Close and Comment'

aemoncannon:
@rorygraves Thanks! That's exactly the kind of feedback I was hoping for.

aemoncannon:
*phew* Updated this PR to actually remove all type-ids. Please don't merge until I at least fix emacs client to work with this!

aemoncannon:
@rorygraves @yazgoo @hedefalk
See CHANGELOG edits for client team action items.

fommil:
Great stuff! Been wanting to do this for a long time but it was always bottom of the list. The protocol should be a lot easier to understand now

aemoncannon:
Yep!

Btw, emacs integration failure is expected. I'll re-kick the CI once the lisp is updated.

hedefalk:
ensime-atom should be safe. We're not using CallCompletionInfoReq,TypeByIdReq, InspectTypeByIdReq or any typeId's.

aemoncannon:
The server code is green on CI, just the out-of-date elisp that's failing. I'm going to go ahead and merge this as I have time right now to do the necessary cleanup.

PR #1231, opened 2016-01-18
Small index optimisation

fommil:
@rorygraves you're it. Please merge if you're happy and it goes green.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -10,6 +10,7 @@ import org.ensime.indexer.DatabaseService._
 import org.ensime.util.file._
 
 //import scala.concurrent.ExecutionContext.Implicits.global
+import scala.collection.breakOut
 import scala.concurrent._
 import scala.concurrent.duration._
 
@@ -117,11 +118,15 @@ class SearchService(
     // index all the given bases and return number of rows written
     def indexBases(bases: Set[FileObject], checks: Seq[FileCheck]): Future[Int] = {
       log.info("Indexing bases...")
-      val checksLookup: Map[String, FileCheck] = checks.map(check => (check.filename -> check)).toMap
-      val basesWithChecks: Set[(FileObject, Option[FileCheck])] = bases.map { base =>
-        (base, checksLookup.get(base.getName().getURI()))
+      val known: Map[FileObject, FileCheck] = checks.flatMap { check =>
+        val f = check.file
+        if (bases(f)) Some(check.file -> check)
+        else None
+      }(breakOut)
+      val work = bases.map {
+        file => indexBase(file, known.get(file))
       }
-      Future.sequence(basesWithChecks.map { case (file, check) => indexBase(file, check) }).map(_.flatten.sum)
+      Future.sequence(work).map(_.flatten.sum)
     }
 
     def commitIndex(): Future[Unit] = Future {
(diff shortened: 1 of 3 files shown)


fommil:
@jschlather now you get me back, can you have a quick review please incase you spot anything obvious?

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -117,11 +118,15 @@ class SearchService(
     // index all the given bases and return number of rows written
     def indexBases(bases: Set[FileObject], checks: Seq[FileCheck]): Future[Int] = {
       log.info("Indexing bases...")
-      val checksLookup: Map[String, FileCheck] = checks.map(check => (check.filename -> check)).toMap
-      val basesWithChecks: Set[(FileObject, Option[FileCheck])] = bases.map { base =>
-        (base, checksLookup.get(base.getName().getURI()))
+      val known: Map[FileObject, FileCheck] = checks.flatMap { check =>
+        val f = check.file
+        if (bases(f)) Some(check.file -> check)
```
fommil: assumes that `FileObject` correctly implements `equals`, of course.

jschlather: It's not clear to me that FileObject does implement equals properly. Looking at the source for [AbstractFileObject](http://grepcode.com/file/repo1.maven.org/maven2/org.apache.commons/commons-vfs2/2.0/org/apache/commons/vfs2/provider/AbstractFileObject.java#AbstractFileObject) and [ZipFileObject](http://grepcode.com/file/repo1.maven.org/maven2/org.apache.commons/commons-vfs2/2.0/org/apache/commons/vfs2/provider/zip/ZipFileObject.java?av=f) it doesn't seem like it's implemented. If it inherits from `Object`, is the map always going to be empty?

jschlather:
It looks fine to me as long as the set lookup works properly.

Also FWIW, I did some profiling of SearchService startup to see what was taking the longest. As a disclaimer, I only looked at how long the SearchServiceSpec tests took, so it might be beneficial to look at a large project. But, the majority of the time spent seemed to spent in the `ClassfileIndexer` calling `indexClassFile`. I didn't look too deeply at that classfile, but there might be some more performance gains in there.

fommil:
OK, let's hold until we can confirm if equals is implemented correctly.

fommil:
you're quite right, `.equals` is not implemented https://github.com/apache/commons-vfs/blob/VFS-2.0/core/src/main/java/org/apache/commons/vfs/provider/local/LocalFile.java

The VFS library just gets worse and worse. It's such a massive behemoth and doesn't seem to do really basic stuff. Classic Apache library.

fommil:
actually, I'm just closing this PR... it's not on my top priority list. As our resident performance optimisation guru, I leave this to @rorygraves to pick up if he'd like to encourage further use of `breakOut` vs `.toMap` and friends.

fommil:
@jschlather yeah one of the easiest performance boosts we could get would be to rewrite the `DescriptorParser` in parboiled2 instead of parboiled.

jschlather:
Interesting, I might take a crack at that when I have some free time and see how much work it would be. If no one else is working on it.

rorygraves:
Ok, did some quick profiler of the indexer startup - its basically all parboiled

aemoncannon:
:-)

jschlather:
Yeah, that's what it was looking like for me too. @rorygraves  Out of curiosity, what did you use to profile the indexer startup?

rorygraves:
Yourkit - They very nicely provided a couple of OSS licenses for ensime

fommil:
@rorygraves run that on windows and see a very different picture ;-) all h2/slick.

But parboiled2 is definitely going to help a **lot** on all platforms.

Issue #1230, opened 2016-01-18
optional jars instead of class directories


fommil:
In large projects it can be possible to insert jars in-place to where the sbt `packageBin` is expected.

It would be good to support these jars instead of class directories when `exportJars` is enabled.

This is high priority for me, because this is now my workflow at work :-) Will require work in `ensime-sbt` to support.

fommil:
https://github.com/ensime/ensime-server/blob/master/core/src/main/scala/org/ensime/indexer/FileWatchers.scala#L44 is looking specifically for classfiles, it also needs to handle jar files and I think this just works.

Issue #1042, opened 2015-07-11
VFS filewatchers are really inefficient


fommil:
*sigh* seems an upstream bug in VFS that we might have to workaround.

Means that changes to H2 or Lucene and causing VFS to spin and use up all the CPU.

To trigger the extreme behaviour, wipe the cache and startup fresh.

fommil:
@aemoncannon we could use this as a general catch all for the upstream bugs/perfomance problems against VFS2

fommil:
* https://issues.apache.org/jira/browse/VFS-536
* https://issues.apache.org/jira/browse/VFS-537
* https://issues.apache.org/jira/browse/VFS-553

fommil:
this one is really killing me at work too! Been thinking about making a Java 7 filewatcher as a dependency, and then using it if its available, otherwise falling back to the Java 6 implementation from VFS.

fommil:
I'm working on some unit tests around the file watchers, including backporting the latest version of the file monitor (it still has these bugs)

but a massive improvement for the workaround dawned upon me... it seems to be able to detect the deletion of the root directory! So instead of watching the parent, I'll just watch for that instead and then recreate when that happens.

fommil:
WIP https://github.com/fommil/ensime-server/tree/export-jars I should finish this off on Friday evening (going out tomorrow night)

fommil:
@lxol you'll probably want to make use of these tests as well when I'm done

lxol:
@fommil   I had to play with delays to pass some of your tests. Here the problems that I noticed.
-  I used `take` instead of `poll` for change monitoring. This is probably the reason why delay is so big. We can try to use `poll` on schedule if the delay is unacceptable.  I am not sure about performance impact though.
- I ignored tests that register files instead of directories for monitoring.  It looks like Java 7 WatchService can register only directories. I got an exception while trying to do that with a file.  I think as a workaround we can register a parent directory and use a selector equal to the file name. 

I submitted a PR to fommil/ensime-server with the successful test

fommil:
just to be clear, this is closed as WontFix. We no longer need to have file watchers watching the file watchers, which is more efficient, but even if I use the new jar target support I'm still seeing ENSIME hang on startup due to all the filewatcher's time being spent in `getType`... so that operation is obviously extremely slow on windows and I don't really think I care to fix it since the JDK7 file watchers are just round the corner.

fommil:
[...]

PR #1235, opened 2016-01-20
Export jars

### LINKED ISSUE #1234: bundle the JDK 6 tools sources
All the OpenJDK sources are tagged here

  http://hg.openjdk.java.net/jdk8u/jdk8u/tags

instead of requiring the user to bundle this, we should just ship a zip of the official JDK of our build, and then merge this back into `ensime-sbt` to simplify the testing (which also requires this manual step).

fommil:
I've rewritten the file watchers to be more efficient, fixed some epic bugs in the indexer (if it ever worked for watched files, it did it by accident) and added support for `.jar` targets via `useJar` https://github.com/ensime/ensime-sbt/blob/v0.3.3/src/main/scala/EnsimePlugin.scala#L43-L46

closes #1234 #1230 #1042 and some housekeeping.


fommil:
Notes to self:

- take one dir per constructor, not a Set
- make the jar watcher take a different selector and be non-recursive with its own tests.

greenrd:
What does this do? Needs a doc PR as well.

fommil:
@greenrd it's very alpha, just for comment.

fommil:
I'm getting somewhere with this! The JDK6 filewatchers are a lot better already, and the JDK7 ones thanks to @lxol will be far better. I'll try to pull that over the line this weekend.

Issue #1214, opened 2016-01-09
SearchService parsing exception


rlvoyer:
It looks as though the ENSIME indexer is blowing up while trying to index [rojoma-json](https://github.com/rjmac/rojoma-json) classfiles.

```
17:41:11.973 WARN  akka://ENSIME/user/ensime-main/project o.e.c.Project - org.parboiled.errors.ParsingException: Invalid :
Invalid input '/', expected ![;/()EOI] or ';' (line 1, pos 33):
()Lcom/rojoma/json/v3/$minusimpl/util/WrapperJsonCodecImpl;
                                ^

17:41:11.979 ERROR akka.dispatch.Dispatcher a.d.Dispatcher - Invalid :
Invalid input '/', expected ![;/()EOI] or ';' (line 1, pos 33):
()Lcom/rojoma/json/v3/$minusimpl/util/WrapperJsonCodecImpl;
                                ^

org.parboiled.errors.ParsingException: Invalid :
Invalid input '/', expected ![;/()EOI] or ';' (line 1, pos 33):
()Lcom/rojoma/json/v3/$minusimpl/util/WrapperJsonCodecImpl;
                                ^

	at org.ensime.sexp.util.ParboiledParser$$anonfun$parse$1.apply(ParboiledParser.scala:23) ~[sexpress_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at org.ensime.sexp.util.ParboiledParser$$anonfun$parse$1.apply(ParboiledParser.scala:23) ~[sexpress_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at scala.Option.getOrElse(Option.scala:121) ~[scala-library-2.11.7.jar:na]
	at org.ensime.sexp.util.ParboiledParser$class.parse(ParboiledParser.scala:22) ~[sexpress_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at org.ensime.indexer.DescriptorParser$.parse(DescriptorParser.scala:12) ~[core_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
	at org.ensime.indexer.ClassfileIndexer$class.org$ensime$indexer$ClassfileIndexer$$classesInDescriptor(ClassfileIndexer.scala:35) ~[core_2.11-0.9.10-SNAPSHOT.jar:0.9.10-SNAPSHOT]
[...]

Issue #1233, opened 2016-01-19
a slow indexer can dramatically slow down type-at-point


fommil:
I am experiencing serious performance problems with the indexer on a huge project.

When a fresh server starts up against a project, type-at-point is pretty snappy. However, after a while, it starts timing out.

I suspect we are querying the indexer and it is timing out, resulting in the type-at-point timing out. We should have aggressive timeouts for the indexer in these sorts of queries to get the answer back to the user.

fommil:
(I'm just guessing, it might be totally unrelated to the indexer...)

aemoncannon:
I have seen this too. Are there lucene queries that can realistically run
for multiple secs?

fommil:
it's probably more a case of waiting for a connection or hitting a `synchronized` somewhere. I don't think lucene or h2 is ever that slow.

aemoncannon:
that's what i figured. our index is just not *that* big.

rorygraves:
Running the indexer on ensime-server creates ~80M temporary objects

rorygraves:
So it basically comes down to replacing DescriptorParser with a parboiled2 parser

aemoncannon:
Should we consider antlr? I assume parboiled2 is still going to have a lot
of the combinator overhead. Antlr parsers are near optimal.

rorygraves:
To the best of my knowledge, parboiled2 is a complete reimplementation using a PEG parser and is much more efficient (orders of magnitude type)

fommil:
That's during creation, this is all on lookup that I'm sung the slowdown.

fommil:
Sung -> seeing *thanks*, Google

Do we have a separate ticket for the parboiled 2 implementation? Might need to create one because the previous ticket was for sexpparser

rorygraves:
Ok, actually went ahead and re-implemented the DescriptorParser in PB2 - it saves about 30% of the indexing time - the slow one now is the pickling - it has its own parser for loading classfiles to get to the pickle sig, a bit of hacking would probably make that go away (reading the sig information directly)

PR coming shortly.

fommil:
"a bit of hacking" **giggles uncontrollably** (once you see that code, you cannot "unsee" it...)

rorygraves:
I mean - extract the symbols from the classfiles ourselves, its running an entire custom parser when we need at most 2 bits of info from each file.

fommil:
*muahahahahahahaha!*

[...]

PR #1236, opened 2016-01-20
Switch DescriptorParser to use Parboiled2 for performance

rorygraves:
Speeds up https://github.com/ensime/ensime-server/issues/1233 a bit and fixes 
https://github.com/ensime/ensime-server/issues/1214

### DIFF
--- a/core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
+++ b/core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
@@ -23,7 +23,7 @@ trait EnsimeConfigFixture {
   def main(lang: String)(implicit config: EnsimeConfig): File =
     config.subprojects.head.sourceRoots.filter { dir =>
       val sep = JFile.separator
-      dir.getPath.endsWith(s"${sep}main${sep}${lang}")
+      dir.getPath.endsWith(s"${sep}main$sep$lang")
     }.head
   def scalaMain(implicit config: EnsimeConfig): File = main("scala")
   def javaMain(implicit config: EnsimeConfig): File = main("java")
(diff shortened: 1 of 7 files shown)


fommil:
ouch. You might need to do some `intransitive` magic in sbt to fix this. I bet this is parboiled2.

```
[error] Modules were resolved with conflicting cross-version suffixes in {file:/drone/src/github.com/ensime/ensime-server/}sexpress:
[error]    com.chuusai:shapeless _2.10.4, _2.10
```

rorygraves:
Oh happy, happy joy joy I love dependency fun!

rorygraves:
Back to you @fommil

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -108,7 +108,7 @@ case class ArrayDescriptor(fqn: DescriptorType) extends DescriptorType {
   }
   def internalString = "[" + fqn.internalString
 }
-case class Descriptor(params: List[DescriptorType], ret: DescriptorType) {
+case class Descriptor(params: Seq[DescriptorType], ret: DescriptorType) {
```
fommil: can you make these `List` please? Or are you relying on this for some performance reason? Note that you end up calling `.toList` on it anyway.

rorygraves: Whoops forgot about that one, was mainly because I was struggling with PB and got a immutable seq.  will have a look.

aemoncannon:
awesome stuff

fommil:
@sirthias and another (much smaller) parser usage here :smile: ... this one will have a big impact on ENSIME startup performance

PR #1238, opened 2016-01-21
Update SExp parser to PB2, Remove Parboiled1

rorygraves:
So the second stage of switching from Parboiled1 to Parboiled2.
This switches the SexpParser and removes the PB1 dependency.

This probably needs exercising by a few users (@fommil / @aemoncannon ?) before merging as it over an ill defined spec and its quite possible I've subtly changed something not caught by our tests.


rorygraves:
@sirthias You mentioned that you would be willing to take a look at the Parboiled2 parser implementations when we got that far. - I finally got to them.
The two files are 
https://github.com/rorygraves/ensime-server/blob/master/sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala

One issue I had was getting the StringBuilder pattern working, so this could definately be improved:
https://github.com/rorygraves/ensime-server/blob/master/sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala#L76

Other simplier one:
https://github.com/rorygraves/ensime-server/blob/master/core/src/main/scala/org/ensime/indexer/DescriptorParser.scala

There are tests (DescriptorParserSpec and SexpParserSpec) if you fancy tweaking/contributing.

fommil:
btw, I think I'd originally said that `sexpress` would be LGPL but in the CLA we say GPL. We should be consistent, can you change the LICENSE (and any header files) to say GPL instead of LGPL?

sexpress/src/main/scala/org/ensime/sexp/util/ThreadLocalSupport.scala
```diff
@@ -0,0 +1,7 @@
+package org.ensime.sexp.util
+
+trait ThreadLocalSupport {
```
fommil: do we still use this?

fommil: to be more idiomatically scala, it should be an `object`, not a `protected` mixin.

rorygraves: We do, just not in the parsers.  I didn't change this - just moved it out of the file that was deleted.

fommil:
another sweet PR from the Rormeister

sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala
```diff
@@ -1,122 +1,166 @@
...
+  }
 
-  // e.g. for .el files
-  def flatParse(el: String): Sexp = parse("(" + el + "\n)")
+  // https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Char-Syntax.html
+  // https://www.gnu.org/software/emacs/manual/html_node/elisp/Syntax-for-Strings.html
+  // Not supported: https://www.gnu.org/software/emacs/manual/html_node/elisp/Non_002dASCII-in-Strings.html
+  private[sexp] val specialChars = Map[String, String](
+    "\"" -> "\"",
+    "a" -> 7.toChar.toString,
```
sirthias: Would something like `'\u0007'` be more readable here?

sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala
```diff
@@ -1,122 +1,166 @@
...
+class SexpParser(val input: ParserInput) extends Parser with StringBuilding {
+
+  private def SexpP: Rule1[Sexp] = rule {
     SexpAtomP | SexpListP | SexpEmptyList | SexpConsP | SexpQuotedP
   }
 
-  private lazy val SexpAtomP: Rule1[SexpAtom] = rule("Atom") {
+  private def SexpConsP: Rule1[SexpCons] = rule {
+    LeftBrace ~ SexpP ~ Whitespace ~ "." ~ Whitespace ~ SexpP ~ RightBrace ~> {
+      (x: Sexp, y: Sexp) => SexpCons(x, y)
```
sirthias: You should be able to shorten this to `... ~> SexpCons` or at least `... ~> SexpCons(_, _)`, depending on whether you have an explicit companion object defined or not.

sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala
```diff
@@ -1,122 +1,166 @@
...
+    }
+  }
+
+  private def SexpListP: Rule1[Sexp] = rule {
+    LeftBrace ~ SexpP ~ zeroOrMore(Whitespace ~ SexpP) ~ RightBrace ~> {
+      (head: Sexp, tail: Seq[Sexp]) => { SexpList(head :: tail.toList) }
```
sirthias: No need for explicit type annotations here.

sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala
```diff
@@ -1,122 +1,166 @@
...
-  private lazy val Digit = rule("Digit") { "0" - "9" }
-  private lazy val Frac = rule("Frac") { "." ~ Digits }
-  private lazy val Exp = rule("Exp") { ignoreCase("e") ~ optional(anyOf("+-")) ~ Digits }
+  private def EscapedChar: Rule1[String] = rule {
+    '\\' ~ capture(ANY) ~> { s: String => unescape(s) }
+  }
 
-  private lazy val SymbolSpecial = rule("SymbolSpecial") { anyOf("+-*/_~!@$%^&=:<>{}") }
+  private def NormalChar: Rule1[String] = rule {
+    capture(CharPredicate.Printable -- "\"\\") ~> { s: String => s }
```
sirthias: This is not great.
The `--` recreates a new CharPredicate here on *every rule application*!
You should factor this out and put it into the companion.

Also, there is no need for `capture` here. Just use `lastChar` in the action.
This also makes `NormalChar` a `Rule1[Char]` instead of `Rule1[String]`, which appears the more natural choice.

sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala
```diff
@@ -1,122 +1,166 @@
...
-  } ~~> { c => SexpChar(c.head) }
+  private def SexpCharP: Rule1[SexpChar] = rule {
+    '?' ~ NormalChar ~> { (c: String) => SexpChar(c.head) }
+  }
 
-  private lazy val SexpStringP: Rule1[SexpString] = rule("String") {
-    ch('"') ~ zeroOrMore(Character) ~~> (chars => SexpString(chars.mkString(""))) ~ ch('"')
+  private def SexpStringP: Rule1[SexpString] = rule {
+    '"' ~ zeroOrMore(Character) ~ '"' ~> { strs: Seq[String] => SexpString(strs.mkString("")) }
+    //    '"' ~ clearSB() ~ zeroOrMore((Character | "\"\"") ~ appendSB()) ~ '"' ~ push(SexpString(sb.toString))
```
sirthias: What's the problem with the commented out line?
The `mkString` based version above is a lot slower!

sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala
```diff
@@ -1,122 +1,166 @@
...
   }
 
-  private lazy val SexpNumberP: Rule1[SexpNumber] = rule("Number") {
-    group(Integer ~ optional(Frac) ~ optional(Exp))
-  } ~> {
-    value => SexpNumber(BigDecimal(value))
+  def SexpNumberP = rule {
+    capture(Integer ~ optional(Frac) ~ optional(Exp)) ~> { s: String => SexpNumber(BigDecimal(s)) }
```
sirthias: No need for the explicit type annotation.

sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala
```diff
@@ -1,122 +1,166 @@
...
 
-  // Rule0 primitives and helpers...
-  private lazy val Alpha = rule("Alpha") { "a" - "z" | "A" - "Z" }
-  private lazy val Integer = rule("Integer") { optional("-") ~ (("1" - "9") ~ Digits | Digit) }
-  private lazy val Digits = rule("Digits") { oneOrMore(Digit) }
-  private lazy val Digit = rule("Digit") { "0" - "9" }
-  private lazy val Frac = rule("Frac") { "." ~ Digits }
-  private lazy val Exp = rule("Exp") { ignoreCase("e") ~ optional(anyOf("+-")) ~ Digits }
+  private def EscapedChar: Rule1[String] = rule {
+    '\\' ~ capture(ANY) ~> { s: String => unescape(s) }
```
sirthias: `lastChar` instead of `capture`!

sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala
```diff
@@ -1,122 +1,166 @@
...
 
-  private lazy val SymbolSpecial = rule("SymbolSpecial") { anyOf("+-*/_~!@$%^&=:<>{}") }
+  private def NormalChar: Rule1[String] = rule {
+    capture(CharPredicate.Printable -- "\"\\") ~> { s: String => s }
+  }
 
-  private lazy val Whitespace = rule("Whitespace") { zeroOrMore(Comment | anyOf(" \n\r\t\f")) }
-  private lazy val Comment = rule("Comment") { ";" ~ zeroOrMore(noneOf("\n")) ~ ("\n" | EOI) }
+  private def SymbolSpecial = rule {
+    anyOf("+-*/_~!@$%^&=:<>{}")
```
sirthias: A `CharPredicate` for this, defined in the companion, will be faster!

sexpress/src/main/scala/org/ensime/sexp/SexpParser.scala
```diff
@@ -1,122 +1,166 @@
...
+  }
 
-  private lazy val LeftBrace = rule("(") { Whitespace ~ "(" ~ Whitespace }
-  private lazy val RightBrace = rule(")") { Whitespace ~ ")" ~ Whitespace }
+  private def Whitespace: Rule0 = rule {
+    zeroOrMore(Comment | anyOf(" \n\r\t\f"))
```
sirthias: `CharPredicate` instead of `anyOf`

sirthias:
@rorygraves Did a quick review of the parser. (Sorry for the delayed reaction!)
If you improve the few bits I highlighted you should have nice parsing performance...
(Didn't check for higher-level grammar problems, but things appears to be relatively straight-forward in your parser...)

rorygraves:
@sirthias Thank for the feedback thats really helpful.  I would not consider that a delayed reaction!

PR #1241, opened 2016-01-22
Address comments on Parboiled2 upgrade

rorygraves:


rorygraves:
Just pushed a change (thanks @fommil) which should hopefully address the 2.10 PB2/Shapeless

PR #1248, opened 2016-01-25
don't ignore jar targets on startup...

### LINKED ISSUE #1247: jar targets ignored on startup
doh! This would have been so simple to test as well.

fommil:
I really need this one for tomorrow (hotfix), so I might bypass review and insta-merge if this one goes green and nobody gets to it first. Closes #1247

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
+++ b/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
@@ -18,7 +18,7 @@ trait FileChangeListener {
   def fileAdded(f: FileObject): Unit
   def fileRemoved(f: FileObject): Unit
   def fileChanged(f: FileObject): Unit
-  def baseAdded(f: FileObject): Unit = {}
+  def baseReCreated(f: FileObject): Unit = {}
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -121,19 +121,21 @@ private class ApachePollingFileWatcher(
...
     val base = vfs.vfile(watched)
-    if (watched.mkdirs()) listeners.foreach(_.baseAdded(base))
+
+    // we don't send baseReCreated if we create it on startup
+    if (watched.mkdirs() && restarted)
+      listeners.foreach(_.baseReCreated(base))
     fm.addFile(base)
     for {
-      file <- watched.tree
+      file <- if (recursive) watched.tree else watched.children
```
fommil: this should actually give a good boost to performance for the jar case

fommil:
@aemoncannon with this fix, if you set up your project to compile to `.jar` instead of a class directory, the performance might actually be tolerable to use target watching. I added some support for this in `ensime-sbt` recently.

aemoncannon:
interesting..

Issue #1243, opened 2016-01-22
Parse error causes Ensime server exit in sexp


dickwall:
As discussed on gitter, new parser impl breaks on the following, and exits ensime during regular use:

```
[Ensime] 2016-01-22 14:37:29,712 [INFO ]  EnsimeInspectTypeStatus.handleReply: {'outer_type_id': False, 'members': [], 'type_id': 3, 'arrow_type': False, 'name': 'Boolean', 'type_args': [], 'full_name': 'scala.Boolean', 'decl_pos': <Ensime.rpc.SourcePosition object at 0x7fbc0b8f9290>, 'decl_as': class}
[Ensime] 2016-01-22 14:37:34,914 [INFO ]  request #7(sync) send - b'(:swank-rpc (swank:completions (:file "/home/dick/dev/VineSleuth/wine4me-api/app/vinesleuth/dbaccess'...
[Ensime] 2016-01-22 14:37:35,915 [WARNI]  sync_req #7 has timed out (didn't get a response after 1.0 seconds)
[...]

PR #1253, opened 2016-01-27
Parboiled2

fommil:
#1243 might have been a false alarm, we can't reproduce.


fommil:
I think this one is good to merge when @dickwall is convinced that it works with sublime

fommil:
@dickwall I could really do with this at my work, don't curse me if it breaks things again :worried:  (also, I've rebased all my WIP against the new code and I don't want to have to revert it all again, and then again when this is accepted)

If you get any problems, `publishLocal` the commit before this and try to grab me on gitter so I can grab all the debugging I need. I think it might have been a transient failure for you because we couldn't reproduce it last night on sublime.

PR #1255, opened 2016-01-27
Add index on file column to the FqnSymbols table

smootoo:
This will make the batches of deletes quicker. (looking at the code, I see it already batches deletes 1000 at a time).
But this is always going to be a relatively slow operation for large scale deletes. Over a certain threshold, it may well be quicker to wipe and rebuild.


fommil:
thank you so much! :grinning:

PR #1256, opened 2016-01-27
(no-op) rename some modules

fommil:
when on the gitter chat room last night, @magnusart pointed out some "interesting" interpretations of what these packages mean in Swedish (hint: not kid safe!) So renaming just to be on the safe side :grin:


aemoncannon:
classin' it up.

fommil:
classy bender

magnusart:
Haha, I wouldn't be too concerned about the naming. Using Ensime implicitly means you're with grownups right?

fommil:
not necessarily, I know friends and colleagues who teach high school kids with Emacs and they want to use ENSIME to show some scala.

magnusart:
Fair enough. :)

PR #1264, opened 2016-01-28
Fix java file manager performance and fix subproject reference roots.

### LINKED ISSUE #1262: Re-enable file manager caching in JavaCompiler.scala
This is an order-of-magnitude speedup when there are lots of jars (jar caching is file managers job!). I measure 800ms -> 50ms when analyzing .java file with 600 jars on the classpath. Notably ensime is still usable at 800ms, but it is *awesome* at 50ms :-)

aemoncannon:
Fixes #1262

### DIFF
--- a/api/src/main/scala/org/ensime/api/config.scala
+++ b/api/src/main/scala/org/ensime/api/config.scala
@@ -34,7 +34,8 @@ case class EnsimeConfig(
   /* Proposed alternatives to the legacy wire format field names */
   def root = rootDir
   def debugVMArgs = debugArgs
-  def referenceSourceJars = referenceSourceRoots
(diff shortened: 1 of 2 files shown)


fommil:
you're using your work address again :-P do you want to fix that up? I've never ever done that, obviously, but if I did, I would absolutely have run `git config user.email "aemoncannon@gmail.com"` but, I never do that sort of thing.

fommil:
CI is java 6, unless you don't have any tests around this... but I'm ok if this breaks for Java 6 temporarily because its an alpha feature.

fommil:
niiice! I was actually using the java support when I was looking through the apache code, it is really really useful to get some basic features like method completion.

Issue #1259, opened 2016-01-28
utf-8 characters not supported by descriptor parser (scalaz)


fommil:
oh scalaz, aren't you just so cute and smart with your fancy method names...

```
12:03:36.878 ERROR akka.dispatch.Dispatcher a.d.Dispatcher - Failed to parse descriptor type: Invalid input 'η', expected PackageNamePredicate, '/', ClassNameCharPredicate or ';' (line 1, column 57):
Lscalaz/syntax/ToApplicativeOps$ApplicativeIdV$$anonfun$η$1;
```

fommil:
@noelmarkham want to have a look at this? I'm not terribly motivated to support scalaz :wink:

isomarcte:
+1 for this bug! It killed my development session today.

fommil:
@isomarcte should be an easy fix, want to submit it? Create a one-line failing test in `SexpParserSpec.scala` (please remember to include a snarky comment about scalaz being very clever :wink: ), and fix it in `SexpParser.scala`. Once merged, it's released.

fommil:
btw, it shouldn't kill the development session. It should just spit out an exception in the server and continue on without search results for scalaz.

rorygraves:
DescriptorParserSpec

fommil:
doh! I'll update my comment.

fommil:
@isomarcte @rorygraves is fixing it

rorygraves:
@rorygraves has fixed it - just running tests and putting together the PR, - I had to go back and reread the jvm spec for what characters are allowed in classnames (at the jvm level), the short answer is 'all of them except / and ;'

fommil:
omg, really? We **have** to use this as the namespace for our monkey patches http://www.fileformat.info/info/unicode/char/1f435/index.htm

fommil:
oh, darn, by definition we have to use the same as the originals. Maybe we can have that just for fun.

rorygraves:
PR raised - 
Which chars are allowed you ask?  All of them (minus delimiters)!!!
```val ClassNameCharPredicate = CharPredicate.All -- ";/ " ```  

Including 🐵

isomarcte:
Thanks for the fast fix. 

@fommil hmm, I'm not sure what killed my connection between emacs and the server then. Although you are right, the server kept on running after the exception was thrown. My mistake.

rorygraves:
No problem, I broke it - its probably fair I fix it ;)

[...]

PR #1266, opened 2016-01-29
Fix utf-8 characters in classnames/packages in DescriptorParser

rorygraves:
Addresses https://github.com/ensime/ensime-server/issues/1259


fommil:
not enough monkeys 🐵 in this PR

PR #1269, opened 2016-01-31
Re-enable the elisp integrations.

aemoncannon:


aemoncannon:
Emacs integrations look good vs 0.8.20. Re-enabling.

Issue #1267, opened 2016-01-29
ClassCastException when requesting DocUriAtPointReq over HTTP


msiegenthaler:
I'm writing a Scala plugin for Cloud9, while most other things (compile errors, code completions, formatting) work I run into a problem with the tooltips I want to display:

I get a "ClassCastException: Cannot cast scala.Some to org.ensime.api.EnsimeServerMessage" whenever I call ENSIME with a DocUriAtPointReq (and if I'm correct also on TypeAtPointReq).
It does not seem to depend on the contents of the file or the position (aka point).

I use the 0.9.10-SNAPSHOT from today.

Full request
`
{
  "typehint": "DocUriAtPointReq",
  "file": {
    "file": "/home/ubuntu/workspace/src/main/scala/B.scala"
  },
  "point": {
    "from": 40,
    "to": 40
  }
}
`

Full exception:
`
23:03:58.315 ERROR ActorSystem(ENSIME) a.a.ActorSystemImpl - Error during processing of request HttpRequest(HttpMethod(POST),http://localhost:38541/rpc,List(Host: localhost:38541, Connection: close),HttpEntity.Default(application/json,146,akka.stream.scaladsl.Source@237af028),HttpProtocol(HTTP/1.1))
java.lang.ClassCastException: Cannot cast scala.Some to org.ensime.api.EnsimeServerMessage
        at java.lang.Class.cast(Class.java:3369) ~[na:1.8.0_72]
        at scala.concurrent.Future$$anonfun$mapTo$1.apply(Future.scala:405) ~[scala-library-2.11.7.jar:na]
        at scala.util.Success$$anonfun$map$1.apply(Try.scala:237) ~[scala-library-2.11.7.jar:na]
        at scala.util.Try$.apply(Try.scala:192) ~[scala-library-2.11.7.jar:na]
        at scala.util.Success.map(Try.scala:237) ~[scala-library-2.11.7.jar:na]
        at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:235) ~[scala-library-2.11.7.jar:na]
        at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:235) ~[scala-library-2.11.7.jar:na]
        at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:32) ~[scala-library-2.11.7.jar:na]
[...]

PR #1271, opened 2016-01-31
Fix for #1267

msiegenthaler:
Fixes the ClassCastException when executing DocUriAtPointReq over HTTP-RPC (no websocket).
Needed to add the same multi phase logic as in RequestHandler.

### DIFF
--- a/server/src/main/scala/org/ensime/server/WebServerImpl.scala
+++ b/server/src/main/scala/org/ensime/server/WebServerImpl.scala
@@ -22,9 +22,24 @@ class WebServerImpl(
 ) extends WebServer with DocJarReading {
   import system.dispatcher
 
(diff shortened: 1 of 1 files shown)


msiegenthaler:
Sorry for not adding a testcase, but there was no infrastructure in place for this sort of test.

msiegenthaler:
#1272 is closely related, adding a fix for that as well.

fommil:
Can you please squash? We'll accept this for now but I suspect this will be removed. Should be trivial to add a test, no? Isn't there a test framework for akka-http?

msiegenthaler:
Squashed it.

Does it help if I do another pull request removing the whole HTTP-RPC stuff to be merged later on?

fommil:
yeah, please do... send it to us when you are transitioned to WebSockets.

PR #1273, opened 2016-02-01
Evil Monkeys

### LINKED ISSUE #1250: file handles being kept open
we are leaving file handlers open on jars during indexing, which means that windows can't delete the files (among other resource leakage goodness). Oh, Windows!

fommil:
:see_no_evil: :hear_no_evil: :speak_no_evil:

close #1250 (hopefully)... although the java compiler might also be holding references. If this doesn't solve it (almost impossible to test unfortunately... or maybe I should add a test that could fail under windows!), I'll remove the `ZipArchive` monkey and come up with a solution involving copying target jars to a temporary directory every time the analyzer restarts.

### DIFF
--- a/core/src/it/scala/org/ensime/intg/JarTargetTest.scala
+++ b/core/src/it/scala/org/ensime/intg/JarTargetTest.scala
@@ -38,7 +38,8 @@ class JarTargetTest extends EnsimeSpec
     }
   }
 
-  it should "allow jar targets to be deleted" in {
+  // FIXME https://github.com/ensime/ensime-server/pull/1273
+  ignore should "allow jar targets to be deleted" in {
     withEnsimeConfig { implicit config =>
       withTestKit { implicit tk =>
         withProject { (project, asyncHelper) =>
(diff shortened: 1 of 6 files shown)


fommil:
I blame @retronym

fommil:
oops, compile error and also forgot to do the assembly stuff.

fommil:
I'm expecting this one to fail the windows CI. I'll investigate in my VirtualBox image, maybe take a heap dump and analyse it in YourKit. The general principles of `ensime-monkeys` should be obvious and also gives @rorygraves a testbed for performance optimisations of the compiler!

fommil:
ok, the good news is that this monkey patch seems to work. The bad news is that `sbt` seems to have a handle on target `.jar` files as well. I don't even know how that can work, because it needs to have some way to delete that file in order to produce a new one. Maybe deleting the file from the `sbt` process is allowed, otherwise I might have to monkey patch the actual compiler too. **sigh**

fommil:
I'll disable the Windows test, until I can workaround it in sbt, so this can be merged.

fommil:
@aemoncannon if this goes green, and you're happy enough with it, can you please merge?

fommil:
I need this for work today. I'll try to get my repo set up to publish to avoid merging.

PR #1274, opened 2016-02-01
Test no monkeys

fommil:
DO NOT MERGE

I'm testing to see if I can get a failure on appveyor.


fommil:
cool It failed in the expected way. Now to see if it passes with the monkey added.

fommil:
damn, I really hoped that one would pass.

PR #1279, opened 2016-02-04
VFS release file handlers

### LINKED ISSUE #1276: something else other than the ZipFile is holding references to the jar file handlers
I have a heap dump of the problem, but I haven't yet found out what is holding the reference, but its **not** the `ZipFile` anymore (that seems to be fixed).

fommil:
close #1276 without having to write a monkey


fommil:
I actually expect this to fail. Although I've solved the main problem, forcibly closing jars/zips means that the source extraction (e.g. when we parse source to calculate the point position, or to expand it for the user ... which is something I'd like us to stop doing, actually) is failing because of the way Apache VFS works. In a way, Apache VFS is designed to only work by accident, because random exceptions and `null`s are completely normal behaviour if there is no cache.

fommil:
a possible fix is to just not "nuke" source jars after we're done with them... on the assumption that nobody else wants to mutate those files.

fommil:
really? This went green? Ok... I'm suspicious.

fommil:
it would appear that my workaround to the source VFS caching actually worked, by gum. @rorygraves / @aemoncannon merge if you're happy with it. I've been shortcutting reviews lately, which is bad... but now I have publishing set up from my `master` branch I have no urgency for my work stuff anymore.

core/src/test/scala/org/ensime/indexer/ClassfileIndexerSpec.scala
```diff
@@ -3,42 +3,35 @@
...
-    with BeforeAndAfterAll {
+class ClassfileIndexerSpec extends EnsimeSpec with IsolatedEnsimeVFSFixture {
 
-  var vfs: EnsimeVFS = _
+  val indexer = new ClassfileIndexer with SLF4JLogging {}
+  import indexer._
 
-  override def beforeAll(): Unit = {
-    vfs = EnsimeVFS()
+  // TODO: some assertions (currently we're just checking that no exceptions are raised!)
```
rorygraves: Seems... incomplete....

fommil: that's an old comment, reindented

fommil: (I recommend reading each commit individually this time because there is one that just moves files around)

rorygraves:
LGTM

Issue #1245, opened 2016-01-23
rewrite tests to use EnsimeSpec


fommil:
we have a preferred testing style, FlatSpec, which is abstracted in `EnsimeSpec`... which also saves a bunch of boilerplate when it is extended.

It would be good to refactor all our tests to use this style, including converting `assert` => matchers.

fommil:
@noelmarkham you once showed some interest in doing this.

ancane:
I'll take a look at this, if you don't mind ;)

fommil:
@ancane :-) feel free to! You will also find that we have a bunch of unit tests that can use the EnsimeVFS fixture (I caught one in my dev branch) and that some integration tests can actually be moved into the `test` config.

BTW, still loving `M-i`... I use it a lot now.

PR #1282, opened 2016-02-07
Migrate tests to EnsimeSpec

ancane:
Issue #1245
Migrated all tests to become EnsimeSpec and changed asserts to should machers where it was possible. Didn't play with EnsimeVFS yet. I'd prefer to move tests from Int into test with another pr, as this one is already huge. Same for deprecated AkkaFlatSpec. Thanks.

### DIFF
--- a/core/src/it/scala/org/ensime/core/DocFindingSpec.scala
+++ b/core/src/it/scala/org/ensime/core/DocFindingSpec.scala
@@ -2,24 +2,16 @@
 // Licence: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.core
 
-import java.io.File
-
-import akka.event.slf4j.SLF4JLogging
 import org.ensime.fixture._
-import org.scalatest._
-
-import scala.collection.immutable.Queue
-import scala.concurrent.Await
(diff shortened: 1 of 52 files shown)


fommil:
great! This must have taken a long time, it is appreciated.

I want to go through and check everything because it'd be a real pain if a silly regression managed to slip in, so I'll do that before merging.

One comment: I noticed you use `should ===` but I think this can be replaced with `shouldBe` (equality) or `shouldEqual` (uses scalactic equality)

ancane:
In general, I was replasing `==` by `shoulBe` and `===` by `should ===` to preserve compile-time type constraint. Shoul I replace all equality checks with shoudBe?

fommil:
ok, I didn't realise `===` did type checking, we can keep that. The only outstanding comment from me is the `Try` vs `intercept`

ancane:
And I didn't realize I should mixin `TypeCheckedTripleEquals` to have that compile time check working. By defaul `===` is still a runtime check.

ancane:
DescriptorParserSpec now intercepts parser exceptions.

fommil:
great stuff! :-)

PR #1290, opened 2016-02-12
Catch some transient java errors

aemoncannon:
Ideally these wouldn't bring down the entire server... but no need to let them bubble up at all as we have no control and they are caused by broken syntax in the parsed file.


fommil:
We've got a few Scala ones like this too.. I've seen the restarting server issue again recently as a result of git checkouts.

rorygraves:
There is a ticket, typecheck request on missing file kills the Pres compiler.

PR #1291, opened 2016-02-13
Add sbt-buildinfo

### LINKED ISSUE #894: Consider adding sbt-buildinfo to build
https://github.com/sbt/sbt-buildinfo

dwijnand:
Fixes #894

---

This generates a file at `server/target/scala-2.11/src_managed/main/sbt-buildinfo/BuildInfo.scala`

with contents:

```scala
package org.ensime.server

import java.io.File
import java.lang.String
import java.net.URL
import scala._; import Predef._

/** This object was generated by sbt-buildinfo. */
case object BuildInfo {
  /** The value is "server". */
  val name: String = "server"
  /** The value is "0.9.10-SNAPSHOT". */
  val version: String = "0.9.10-SNAPSHOT"
  /** The value is "2.11.7". */
  val scalaVersion: String = "2.11.7"
  /** The value is "0.13.9". */
  val sbtVersion: String = "0.13.9"
  override val toString: String = {
    "name: %s, version: %s, scalaVersion: %s, sbtVersion: %s" format (
      name, version, scalaVersion, sbtVersion
    )
  }
}
```

Let me know if you want to add or change anything, I'm happy to iterate (and yak-shave).

Also let me know if you have a way for testing for this addition.

### DIFF
--- a/core/src/main/scala/org/ensime/config/Environment.scala
+++ b/core/src/main/scala/org/ensime/config/Environment.scala
@@ -1,17 +1,22 @@
 // Copyright: 2010 - 2016 https://github.com/ensime/ensime-server/graphs
 // Licence: http://www.gnu.org/licenses/gpl-3.0.en.html
-package org.ensime.config
+package org.ensime
+package config
 
 import java.net.{ JarURLConnection, URL }
 
 object Environment {
-  def info: String = """
+  def info: String = s"""
     |Environment:
-    |  OS : %s
-    |  Java : %s
-    |  Scala : %s
-    |  Ensime : %s
-  """.trim.stripMargin.format(osVersion, javaVersion, scalaVersion, ensimeVersion)
+    |  OS : $osVersion
+    |  Java : $javaVersion
+    |  Scala version: $scalaVersion
+    |  Ensime : $ensimeVersion
+    |  Built with Scala version: ${BuildInfo.scalaVersion}
+    |  Built with sbt version: ${BuildInfo.sbtVersion}
+    |  Built from git sha: ${BuildInfo.gitSha}
+    |  Built on: ${BuildInfo.builtAtString}
+  """.trim.stripMargin
 
   private def osVersion: String =
     System.getProperty("os.name")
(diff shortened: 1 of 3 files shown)


rorygraves:
Can you get the guy hash in there?

rorygraves:
Git even. That will teach me for responding go my phone.

rorygraves:
Would love the server startup to log the info.

rorygraves:
Creation date?

dwijnand:
Add git sha and build time, which adds:

```scala
  /** The value is "682c5913069cbb651ac50e5c48009765bbd10923". */
  val gitSha: String = "682c5913069cbb651ac50e5c48009765bbd10923"
  /** The value is "2016-02-13 14:35:03.218". */
  val builtAtString: String = "2016-02-13 14:35:03.218"
  /** The value is 1455374103218L. */
  val builtAtMillis: Long = 1455374103218L
```

and tweaks the `toString`

rorygraves:
Sweet

dwijnand:
Added the build info to `Environment.info` so it gets logged together with it on server startup.

dwijnand:
The build has passed for Scala 2.11, which means there's no test for the previous `Environment.info`..

How can I add one?

dwijnand:
Welp so much for that test..

fommil:
Hehe. For testing the make sense, i think we'd want the escalate this to the protocol layer. As is, this is just logged as a side effect.

rorygraves:
You can still add one, @fommil  is getting over excited ;)
I think you can build a test that asserts that the update date is within the last hour (to show that the file is being regenerated), but not sure how you would definitively test the entire class.

rorygraves:
@fommil Yes - it would be great if this was within the connection info protocol.

rorygraves:
Nice job btw :)

fommil:
Yeah, this has been on the list for ages

dwijnand:
Given this feature came from a feature request I wouldn't classifying the value it's adding to the program as a side-effect - it's an effect you're interested in the program having, because you think it's of interest to your users.

The test is exercising the program such that the effect desired is obtained, as closely as possible to how the user would obtain it. We would just need to capture (retroactively) the requirements for that (for instance that the build date is correctly up to date).

That's why I asked the question: how can I add a test that as closely as possible simulates a user starting the ensime server and seeing some useful logging information about it?

fommil:
Intercept the log in an integration test?

dwijnand:
Sounds like it could be viable. Where should I start from? What should I look at?

Meanwhile let me just test if I can actually run integration tests with JDK 7...

fommil:
It'd be far easier just to add this to the api response for server info and then check explicitly. That'd actually be useful as well. Add to the api project, as a member of serverinformation in outgoing.scala

dwijnand:
In `ConnectionInfo`? In a new case class? How am I checking it explicitly then? I don't know this build or this codebase remember :)

fommil:
The beauty is that if you add something to that class, a test will fall. Probably the jerky/swanky protocol formats and also there is an integration test somewhere that asserts on the values in the connection info. It's rarely used within the server, so a git grep should find it easily. If not, I'll hunt it down for you when I'm at a computer.

PR #1292, opened 2016-02-13
tidy the sbt build, move sensible defaulty things into their own file

fommil:


fommil:
Oh joy

fommil:
bizarre consistent failure

```
Some("docs/scala-library-2.11.7-javadoc.jar/index.html#scala.Some@++[B>:A,That](that:scala.collection.GenTraversableOnce[B])(implicitbf:scala.collection.generic.CanBuildFrom[Repr,B,That]):That") was not equal to Some("docs/scala-library-2.11.7-javadoc.jar/index.html#scala.Some@++[B](that:scala.collection.GenTraversableOnce[B]):Option[B]") (DocResolverSpec.scala:286)
```

fommil:
ok, I think this is the cachedResolution bug (doesn't see shapeless)

fommil:
ok, I think I see the 2.10 problem. Fixing now. Still no idea about the javadoc failures.

fommil:
I wonder if we have a corrupted docker image... I'll check

fommil:
nope, that scaladoc file matches the one I have locally.

fommil:
now it fails because of the guava bump, that I can understand.

fommil:
@aemoncannon this failure is driving me crazy, it's supposed to be a no-op refactoring to the build definition, but it's resulting in this consistent error in the doc resolver test

```
Some("docs/scala-library-2.11.7-javadoc.jar/index.html#scala.Some@++[B>:A,That](that:scala.collection.GenTraversableOnce[B])(implicitbf:scala.collection.generic.CanBuildFrom[Repr,B,That]):That") was not equal to Some("docs/scala-library-2.11.7-javadoc.jar/index.html#scala.Some@++[B](that:scala.collection.GenTraversableOnce[B]):Option[B]") (DocResolverSpec.scala:286)
```

any ideas why that might be happening?

fommil:
also, it passes locally and on the AppVeyor. I'll try nuking the docker ivy cache and re-running

aemoncannon:
@fommil 
This seems to hinge on a @usecase difference. Signatures with implicit CanBuildFrom are often replaced in the docs with dumbed-down signatures without the CanBuildFrom. We handle this insanity by actually loading the html in question and checking if the the @usecase signature is present. See DocUsecaseHandling in core.

Is it possible the doc html is available in one case and not in another... or different?

fommil:
@aemoncannon seems to be an identical javadoc jar in both cases.

fommil:
interestingly, we don't close the `JarFile`, but I don't think that's causing this madness.

fommil:
ok, can confirm that the usecase regex isn't catching. I can reproduce in my local docker, but not my local non-docker. I'll close this until I can work out what the problem is.

fommil:
FYI, the usecase matching is failing with "input length = 1"

fommil:
oh, ffs, this might be the `javaOptions += "-Dfile.encoding=UTF8"` line.

fommil:
yeah, it was dropped and it must not be the default in the drone runs. I'm such a moron. I'll reopen after squashing and fixing.

fommil:
@aemoncannon interestingly, if we upgrade to guava 19, the doc resolver tests fail... its finding the scala names for everything instead of the java names.

fommil:
I'm merging before the CI gods change their mind.

aemoncannon:
Hey good catch :)

PR #1297, opened 2016-02-15
re-use tcp and http ports from previous sessions

### LINKED ISSUE #1277: use ports in the files, instead of bailing
as discussed on gitter with @hedefalk and @rorygraves it would be advantageous if the ensime server could use the same port as in a previous session. For Atom, it sounds like they want to use this as the mechanism for restarting the server when the `.ensime` file changes (maybe emacs could do the same, although there is some old startup initialisation logic in there) but for Windows it is a good workaround to https://github.com/ensime/ensime-emacs/issues/338

I suppose the server, when it sees a port file (or websockets file) would attempt to use that port... and only if it fails to obtain the exclusive ownership of that port would it fail.

fommil:
close #1277 

FYI @hedefalk 

@rorygraves you're it.


rorygraves:
lgtm

Issue #1089, opened 2015-09-10
Improve error messages on class watcher errors


rorygraves:
An unwrapped NPE during actor initialisation is not very helpful, we should 

http://pastebin.com/sqkimp9w

a) move it to a post create step.
b) Catch the raw exception and provide details of files etc.

rorygraves:
Closing as I think this is handled.

PR #1299, opened 2016-02-16
Actorize ClassfileWatcher

lukasz-golebiewski:
Prerequisite for #1089

### DIFF
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -53,7 +53,7 @@ class Project(
     def fileRemoved(f: FileObject): Unit = reTypeCheck()
     override def baseReCreated(f: FileObject): Unit = reTypeCheck()
   }
-  private val classfileWatcher = new ClassfileWatcher(config, searchService :: reTypecheck :: Nil)
+  private val classfileWatcher = context.actorOf(Props(new ClassfileWatcher(config, searchService :: reTypecheck :: Nil)), "classFileWatcher")
 
   def receive: Receive = awaitingConnectionInfoReq
 
@@ -106,7 +106,6 @@ class Project(
 
   override def postStop(): Unit = {
     // make sure the "reliable" dependencies are cleaned up
-    Try(classfileWatcher.shutdown())
     Try(sourceWatcher.shutdown())
     searchService.shutdown() // async
     Try(vfs.close())
--- a/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
+++ b/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
@@ -2,6 +2,7 @@
 // Licence: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.indexer
 
+import akka.actor.{ Actor, Props }
 import akka.event.slf4j.SLF4JLogging
 import org.apache.commons.vfs2._
 import org.apache.commons.vfs2.impl.DefaultFileMonitor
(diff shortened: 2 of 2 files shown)


fommil:
this keeps running out of memory. I'm starting to suspect that this is not a problem with our CI because a few other builds went through. Can you think of any memory leaks that you might have introduced?

fommil:
@chipsenkbeil I've checked the logs and all the builds for this are running on your machine, e.g. http://fommil.com/ensime/ensime-server/193 can you please check that your dockers are definitely able to give up to 8GB of RAM?

fommil:
yeah, more

```
[error] Uncaught exception when running org.ensime.core.SemanticHighlightingSpec: java.lang.OutOfMemoryError: unable to create new native thread
```

@lukasz-golebiewski can you have a think about this please?

lukasz-golebiewski:
Sure, I'll have a look

chipsenkbeil:
@fommil so I can see one container actively running right now on my machine related to ensime.

screen shot 2016-02-21 at 1 18 45 pm
screen shot 2016-02-21 at 1 19 20 pm

I do see the OutOfMemory exception in the logs, but I'm being told that the container has access to nearly 16 GB of RAM. Is there something I need to provide to the Docker daemon to allow it to use that much memory? I haven't done anything to it; so, I assumed that it already has access to available memory.

fommil:
@chipsenkbeil we still haven't isolated if its your box or the code itself. From the look of that chart, it does seem like it shouldn't have a memory problem :-/

chipsenkbeil:
I also executed `docker exec -it ensime-container-name /bin/bash`, installed the free program via `apt-get install procps`, and then ran `free -mh`:

screen shot 2016-02-21 at 1 29 32 pm

fommil:
thanks for looking into this @chipsenkbeil ! Where do you get your machine hosting from? I'd like to use that instead of contabo.

lukasz-golebiewski:
I executed the `it:test`s on my machine several times on `master` and on this branch. The memory usage never went above 800 MB and the number of live threads never exceeded 40. Peculiar

fommil:
oh, but we fork all our tests... this is just the compiler process.

lukasz-golebiewski:
Right, I'll investigate further tomorrow

fommil:
yeah, I'm pretty confident this has introduced a regression. There have been tonnes of other tests today and none are exhibiting the same problem.

I honestly can't see the bug though.

lukasz-golebiewski:
Tests for this PR were successful one time: http://fommil.com/ensime/ensime-server/191
Is it possible that other tests were executed on other worker nodes on which the problems do not occur? Could we trigger a build of `master` on the one that's failing? I'm just guessing here, as I also cannot see the error - the new Actor should be destroyed along with it's parent and the resources should be freed.

fommil:
yeah, it's altogether very confusing.

fommil:
ok, definitely a docker problem. I just seen this on a totally unrelated build http://fommil.com/ensime/ensime-server/212/1

fommil:
@chipsenkbeil I'm taking your machine off the network for a bit, we should investigate what is going on

chipsenkbeil:
@fommil okay, let me know what you need done to help figure out this problem. I'm not seeing anything obvious on my end in terms of memory usage, just that the exception is thrown.

fommil:
I actually don't know how to fix it, but I suspect it'll be something to do with docker version / system config? fyi we're on `docker --version ; Docker version 1.10.1, build 9e83765`

chipsenkbeil:
Same build for me. I'll send you my docker.service file privately on Gitter.

fommil:
@lukasz-golebiewski sorry to leave you hanging on this one. Can you please rebase and squash your commits?

lukasz-golebiewski:
@fommil Done

PR #1306, opened 2016-02-19
Add stack traces to shutdown error messages.

jschlather:
This is motivated by an issue I had where ensime failed to start and the log just said

```
15:56:17.331 ERROR None Server - Shutdown requested due to internal error: java.lang.NumberFormatException: For input string: "55720
"
```

Which was a little confusing. I grepped my entire project directory for 55720 and found the culprit to be my `.ensime_cache/http` file which I deleted and the issue was resolved. The fact that the number is in the right range for ports, is a hint, but I figure it would have been easier to debug if I had known where in the flow ensime had the error.

### DIFF
--- a/server/src/main/scala/org/ensime/server/Server.scala
+++ b/server/src/main/scala/org/ensime/server/Server.scala
@@ -34,6 +34,7 @@ class ServerActor(
 
   override val supervisorStrategy = OneForOneStrategy() {
     case ex: Exception =>
+      log.error(s"Error with monitor actor ${ex.getMessage}", ex)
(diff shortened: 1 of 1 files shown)


fommil:
why not just log out the exceptions using the logger where they are received? This extra level if indirection could be confusing in itself :confused:

fommil:
cool, that looks better :smile:

jschlather:
Good :smile:.

Let me know if there's anything else I should do.

fommil:
this job seems to have ended up in a zombied state on my build machine. No idea why.

fommil:
thanks!

PR #1311, opened 2016-02-21
Minor test cleanup

### LINKED ISSUE #1245: rewrite tests to use EnsimeSpec

ancane:
* Move DiffUtilSpec from it to test config
* Use TestKitFix instead of AkkaTestUtil and drop the latter
* Use EnsimeVFS fixture at ClassfileDepickler spec
This PR should close #1245.


fommil:
cool! Been meaning to do some of this for a long time

PR #1312, opened 2016-02-21
coursier for fun and profit

### LINKED ISSUE #1296: use coursier for ivy resolution
purely a developer enhancement, enable the sbt plugin.

https://github.com/alexarchambault/coursier#sbt-plugin

unfortunately 1.0.0-M7 uses java 7.

https://github.com/alexarchambault/coursier/issues/137

fommil:
close #1296

workaround https://github.com/sbt/sbt/issues/2266 and https://github.com/sbt/sbt/issues/1930

things to be cautious of:

- https://github.com/alexarchambault/coursier#important-sbt-plugin-might-mess-with-published-artifacts (our emacs integration tests will hopefully work this one out, and I'm also seeing locally good results)
- https://github.com/alexarchambault/coursier/issues/138 <= not as fast as it could be, but of course @alexarchambault has a fix up his sleeve...

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -7,6 +7,7 @@ build:
     - EMACS=/opt/emacs-24.5/bin/emacs
     - AKKA_TEST_TIMEFACTOR=10
     - SBT_TASK_LIMIT=6
+    - COURSIER_MODE=update-local-changing,missing
   commands:
     - if $(git grep -qE "TODO|FIXME" -- `git ls-files | grep -v .drone.yml`) ; then
         echo "Please remove TODO or FIXME. Create an issue at GitHub instead." ;
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,6 +21,7 @@ install:
   - cmd: SET SBT_OPTS=-XX:MaxPermSize=256m -Xss2m -Xms512m -Xmx512m
   - cmd: SET AKKA_TEST_TIMEFACTOR=10
   - cmd: SET SBT_TASK_LIMIT=2
+  - cmd: SET COURSIER_MODE=update-local-changing,missing
 build_script:
   - sbt clean compile test:compile it:compile
 before_test:
@@ -31,6 +32,7 @@ cache:
   - C:\sbt\
   - C:\Users\appveyor\.m2
   - C:\Users\appveyor\.ivy2
+  - C:\Users\appveyor\.coursier
   - C:\Program Files\Java\jdk1.7.0\src.zip
 on_failure:
   - appveyor PushArtifact -Path '**\*.log' -FileName logs -Type zip
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -23,3 +23,5 @@ scalacOptions in Compile ++= Seq("-feature", "-deprecation")
 ivyLoggingLevel := UpdateLogging.Quiet
 
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.0")
+
+addSbtPlugin("io.get-coursier" % "sbt-coursier-java-6" % "1.0.0-M11")


alexarchambault:
Just one thing: like the CLI tool, the snapshot dependencies won't get automatically updated here either. Wipe all the `*-SNAPSHOT` things in `~/.coursier` as a (temporary) workaround.

fommil:
ah, good point. Although in our CI we don't depend on SNAPSHOTs... and this might just answer our debate about third party SNAPSHOT deps.

fommil:
@alexarchambault badness in our CI http://fommil.com/ensime/ensime-server/213/1 does this look transient to you?

fommil:
@alexarchambault are you doing some weird bash stuff insider coursier? https://ci.appveyor.com/project/aemoncannon/ensime-server/build/1.0.344 lots of windows messages about bash tty

fommil:
the logging is a bit excessive, it would be good to silence it and only show when something is being downloaded. I like to see network activity and nothing else.

fommil:
interesting, Windows appears to have fared better than Linux. Could be a transient thing. Should you perhaps put in some retry logic so a bad download of a pom/jar doesn't nuke an entire build?

alexarchambault:
That's the progress bars output flowing the terminal. Set the environment variable `COURSIER_NO_TERM` (to `1` say) to get the same kind of messages as http://fommil.com/ensime/ensime-server/213/1.

fommil:
Windows is giving some really funky logback warnings that I've never seen before. I wonder if perhaps the intransitive / exclude rules are being ignored?

fommil:
@alexarchambault cool, thanks. Even the Linux logs are a bit noisy, is there any way to get rid of everything except lines that begin "Downloading "

fommil:
the linux 2.11 build keeps seeing stuff like

```
Caused by: java.io.FileNotFoundException: /root/.coursier/cache/v1/https/jcenter.bintray.com/org/netbeans/api/org-netbeans-modules-projectapi/RELEASE731/org-netbeans-modules-projectapi-RELEASE731.pom.lock (No such file or directory)
```

there are obviously a few kinks to iron out, but this is looking sweet.

alexarchambault:
Could be a problem with rights... Are you sharing the `~/.coursier` caches between builds?

alexarchambault:
About logging, isn't there really missing log messages? In particular to have details about the timeout of errors like
```
Caused by: akka.event.Logging$LoggerInitializationException: Logger log1-Slf4jLogger did not respond with LoggerInitialized, sent instead [TIMEOUT]
```

fommil:
I've never seen these logback errors on windows before, so I suspect the classpath is not as it should be.

I can't think of a rights problems, these are all running is isolated docker containers.

alexarchambault:
I have a Windows VM around, I'll try to reproduce the Appveyor error.

What did you make to have the drone CI pass?

fommil:
it was transient, so I think maybe if coursier could be more resilient to bad downloads that would be good

fommil:
created a ticket for the logging issue https://github.com/alexarchambault/coursier/issues/183

@alexarchambault do you want me to create a ticket for the transient download failures too? And perhaps the windows slf4j stuff. I wonder if its a classpath ordering thing.

alexarchambault:
@fommil Done here: https://github.com/alexarchambault/coursier/issues/187.

About the Windows errors, there's a small chance https://github.com/alexarchambault/coursier/pull/181 could address it. Wait for the next milestone.

I have one idea to make the exception-prone line more secure, that could address the transient errors (Not 100% sure though).

fommil:
@alexarchambault the latest CI run after bumping to M9 doesn't look healthy. Missing sbt `update` or `updateClassifiers` key :confused:

alexarchambault:
@fommil See the `Not found` messages. Some updates fail because it can't find some netbeans-related dependencies. Likely because of missing repositories, `http://bits.netbeans.org/nexus/content/groups/netbeans` I think.

About the missing task key, it looks like [this error](https://github.com/ensime/ensime-sbt/blob/1ea2fa8d4e59c514c51d494ed4c25ce5c94ee269/src/main/scala/EnsimePlugin.scala#L573), so ensime-sbt not handling failed updates well.

fommil:
It happening reliably. The netbeans jars are on bintray.

alexarchambault:
Which I don't see in the resolvers, e.g. for the `core` module,
```
Repositories:
  ivy:file:/root/.sbt/0.13/plugins/target/resolution-cache[organization]/[module](/scala_[scalaVersion])(/sbt_[sbtVersion])/[revision]/resolved.xml.[ext]
  inter-project
  ivy:file://${ivy.home}/local/[organisation]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]
  https://repo1.maven.org/maven2/
  https://oss.sonatype.org/content/repositories/snapshots/
```

First two are added by coursier, the remaining ones are `~/.ivy2/local`, Central, and Sonatype snapshots. No bintray here :-)

alexarchambault:
You're bitten by [this point](https://github.com/alexarchambault/coursier#why-does-coursier-seem-not-to-find-some-artifacts-whereas-sbt-does) here.

fommil:
hmm, but we do have

```
    fullResolvers += Resolver.jcenterRepo
```

alexarchambault:
The coursier SBT plugin only looks at `externalResolvers`... Any reason you used `fullResolvers` instead of just `resolvers`? Maybe the plugin should look at `fullResolvers`...

fommil:
I've changed to use just `resolvers` in this latest update, I can't remember why I didn't just use `resolvers` in the first place.

alexarchambault:
By the way, maybe you should wait for the next milestone of coursier (in a few days I think). It will have some [important fixes](https://github.com/alexarchambault/coursier/pull/181), that could address the issue on Appveyor here. I'd like to make the plugin less verbose by default too, and add some possible safeguards against the transient errors we were seeing on the Linux CI (https://github.com/alexarchambault/coursier/issues/187).

fommil:
planning on doing that, I just wanted to test out the latest milestone :smile:

btw, I think the logger error in appveyor is unrelated to coursier. I've seen it in sbt/ivy branches and requires further testing. Probably a classpath initialisation problem with akka doing some reflection on something that's not loaded yet.

The transient error updates sound great! Also looking forward to silent logs (except when downloading a file)

fommil:
heh, appveyor just failed with https://ci.appveyor.com/project/fommil/ensime-server-circ6/build/1.0.2

```
Fetching artifacts
bash: /dev/tty: No such device or address
Fetching artifacts: done
[info]   Compilation completed in 23.736 s
java.lang.RuntimeException: Missing scala-library.jar
	at scala.sys.package$.error(package.scala:27)
	at sbt.Defaults$$anonfun$scalaInstanceFromUpdate$1$$anonfun$file$1$1.apply(Defaults.scala:392)
	at sbt.Defaults$$anonfun$scalaInstanceFromUpdate$1$$anonfun$file$1$1.apply(Defaults.scala:392)
	at scala.Option.getOrElse(Option.scala:120)
	at sbt.Defaults$$anonfun$scalaInstanceFromUpdate$1.file$1(Defaults.scala:392)
	at sbt.Defaults$$anonfun$scalaInstanceFromUpdate$1.apply(Defaults.scala:394)

```

alexarchambault:
Yeah, this one is suspicious :disappointed:

alexarchambault:
Found [this](https://ci.appveyor.com/project/fommil/ensime-server-circ6/build/1.0.2#L457) which can explain these. Several concurrent accesses to the same JAR that hinder themselves. Related to https://github.com/alexarchambault/coursier#no-wait-on-locked-file, but an explicit error, or some re-attempts, would have been better.

fommil:
you should really test on Windows if you're ever doing concurrent FS access ;-)

if the gratis version of appveyor is too slow, add me as a collaborator and then I can add appveyor under my paid account.

fommil:
I get that `bash` error even in the linux builds. What is calling out to `bash`?

alexarchambault:
This should be fixed since https://github.com/alexarchambault/coursier/pull/186 (so for the next milestone). That was originating from a call to ammonite-terminal to get the terminal width.

fommil:
cool, looking forward to the milestone update!

alexarchambault:
I just released the java 6 version of coursier 1.0.0-M10. Could be worth a try! (although I don't know what causes the current Appveyor error...).

fommil:
@alexarchambault I've updated to M10, we'll see how the CI goes but locally it looks like there is still a load of logging going on. Do I need to do anything to reduce the level? Basically the second time I run it I would like there to be no logging from coursier because it is only logging when it is downloading something (in CI I wouldn't expect to see those download meters).

Is there anything I need to do to ensure that it picks up on the latest version of our SNAPSHOT dependencies?

fommil:
@alexarchambault btw, the windows logging thing was totally unrelated :smile: turned out that our CI box got a little slower and the loggers were timing out inside akka, because akka is awesome and all.

fommil:
let's track those requirements. Blocked on

- [ ] noisy logging https://github.com/alexarchambault/coursier/issues/119 and https://github.com/alexarchambault/coursier/issues/183
- [x] cache missing classifiers https://github.com/alexarchambault/coursier/issues/138
- [ ] ensure SNAPSHOT dependencies are retried

fommil:
@alexarchambault I think Maven Central hasn't disted across the globe yet. We have some build machines in the USA.

alexarchambault:
The organization and name of the plugin changed, see https://github.com/alexarchambault/coursier#sbt-plugin (updated a second ago). Add `-java-6` to the name here.

alexarchambault:
Also, you can set the environment variable `COURSIER_MODE` to `update-local-changing,missing` for the snapshot artifacts to be checked for updates.

I would have expected something like
```scala
coursierCachePolicies in ThisBuild := Seq(
  coursier.CachePolicy.LocalUpdateChanging,
  coursier.CachePolicy.FetchMissing
)
```
to work too, but it looks like it doesn't.

fommil:
oh, so we don't get that by default?

alexarchambault:
TTL should be enabled by default. I was (still am actually) having the feeling that systematic update check by default was kind of annoying.

fommil:
yeah, it is kind of annoying but it's also a reminder why one shouldn't depend on `SNAPSHOT`s :smile:

alexarchambault:
Yeah, I guess that's even. TTL support will fix that.

fommil:
so what should we do, use the flag for now?

alexarchambault:
Yes, on the CI at least. (And locally, one can set it from time to time.)

fommil:
**I'M PRESSING THE BUTTON!!!!**

PR #1317, opened 2016-02-23
FIXME: eliminate all the Option responses

### LINKED ISSUE #1316: RequestHandler FIXME: Find and eliminate all the Option responses legacy
Eliminate Option responses according to [FIXME](https://github.com/ensime/ensime-server/blob/cbacc3d37577be93e68a70c597442961d0e998c6/server/src/main/scala/org/ensime/server/RequestHandler.scala#L49)

ancane:
Heya! Trying to fix #1316
Request handler isn't listening to Options any more. FalseResponses returned from project actors instead. There is a non-RpcResponse family DocSigPair responses left. Wasn't sure if I should change it or not.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -4,7 +4,7 @@ build:
   environment:
     - SONATYPE_USERNAME=$$SONATYPE_USERNAME
     - SONATYPE_PASSWORD=$$SONATYPE_PASSWORD
-    - EMACS=/opt/emacs-24.4/bin/emacs
+    - EMACS=/opt/emacs-24.5/bin/emacs
     - AKKA_TEST_TIMEFACTOR=10
   commands:
     - git log | head -n 20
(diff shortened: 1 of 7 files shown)


fommil:
cool! This looks sensible. There might be another handler in the WebSockets side, I seem to recall it being added recently.

Tests failing, can you see the error?

ancane:
Thanks for the hint, I'll check websocket handler. Testing failed because of flickering FileWatcherSpec.

ancane:
Hm, I was wrong, flickering test doesn't fail all the suite. The true error is strange:
cask pkg-file: error: Emacs does not exist at /opt/emacs-24.4/bin/emacs.  Did you install Emacs?
[info] build failed (exit code 1)

fommil:
oh damn, I think I might have removed that from the docker image (thought it was unused). Change it to emacs 24.5 in `.drone.yml`

ancane:
Ya! It worked. I changed emacs to 24.5 in drone.yml and removed Option response handling in webserver.

fommil:
Awesome!

PR #1323, opened 2016-02-27
Stop PC failing when completion point outside file

rorygraves:

### DIFF
--- a/core/src/main/scala/org/ensime/core/Completion.scala
+++ b/core/src/main/scala/org/ensime/core/Completion.scala
@@ -48,7 +48,6 @@ import scala.collection.mutable
 import scala.concurrent.duration._
 import scala.concurrent.{ Await, Future }
 import scala.reflect.internal.util.{ BatchSourceFile, SourceFile }
-import scala.tools.nsc.io.AbstractFile
 
 trait CompletionControl {
   self: RichPresentationCompiler =>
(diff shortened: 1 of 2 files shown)


rorygraves:
I think this is an unrelated failure - possibly due to the snapshot releases of the refactoring library?  @fommil 

Btw - PR is easier to read in this form: https://github.com/ensime/ensime-server/pull/1323/files?w=1

core/src/main/scala/org/ensime/core/Completion.scala
```diff
@@ -349,9 +357,10 @@ object CompletionUtil {
   def matchesPrefix(m: String, prefix: String,
     matchEntire: Boolean, caseSens: Boolean): Boolean = {
     val prefixUpper = prefix.toUpperCase
-    ((matchEntire && m == prefix) ||
+
+    (matchEntire && m == prefix) ||
```
rorygraves: Removal of bonus brackets.

core/src/main/scala/org/ensime/core/Completion.scala
```diff
@@ -312,8 +321,9 @@ object Keywords {
     "yield"
   )
 
-  val keywordCompletions = keywords map { CompletionInfo(_, CompletionSignature(List(), "", false), false, 100, None) }
-
+  val keywordCompletions = keywords map {
```
rorygraves: Relayout only

core/src/main/scala/org/ensime/core/Completion.scala
```diff
@@ -77,87 +76,97 @@ trait CompletionControl {
   import CompletionUtil._
 
   def completionsAt(inputP: Position, maxResultsArg: Int, caseSens: Boolean): CompletionInfoList = {
-    val maxResults = if (maxResultsArg == 0) Int.MaxValue else maxResultsArg
+    val origContents = inputP.source.content
+    val point = inputP.endOrCursor
+
+    if (point > origContents.length) {
```
rorygraves: 90% of the change is the same code - I've wrapped it in a defensive if block.

fommil:
I see what you mean. Do you mind pushing a small commit to address the changes in the refactoring library, unless you think it's just wrong. Looks small.

rorygraves:
I will do some more tests on that first- but yeah

fommil:
I'm not too keen on the refactoring-library change. It is making things come out as

```scala
   package org.ensime.testing

   import java.lang.Integer
   trait Temp {
     valueOf(5)
     vo("5")
     toBinaryString(27)
   }
```

```scala
   package org.ensime.testing

   import java.lang.Integer
   
   trait Temp {
     valueOf(5)
     vo("5")
     toBinaryString(27)
   }
```

instead of 

```

```

rorygraves:
Yeah, looks like a regression on the library or some weirdness do to the recent fun with repositories on their side?

fommil:
@sschaef @mlangc any ideas what could have caused the above regression?

fommil:
it's got to be https://github.com/scala-ide/scala-refactoring/pull/139

fommil:
ok, it's intentional.

core/src/main/scala/org/ensime/core/Completion.scala
fommil: oh, excellent. We need a few more in ensime-emacs.

sschaef:
The new behavior makes the reafactoring library safer to use. Before merging, we should have verified first if it breaks your test suite. Sorry about that!

The plan is to let organize imports do the cleanups, i.g. the insertion of newlines if they are necessary.

rorygraves:
@fommil AppVeyor has failed for an unrelated slf4j initialisation issue.

fommil:
Pull requests that have a failing status can’t be merged from a phone.

Ha! Nice feature

PR #1324, opened 2016-02-27
Adapt new behavior of add import refactoring in test suite

sschaef:
scala-refactoring no longer prints new lines after in the add import
refactoring. See PR https://github.com/scala-ide/scala-refactoring/pull/139

----

We may rethink the decision to no longer print new lines in the add refactoring import but it is more likely that we stay with the decision to run the organize import refactoring afterwards to do the cleanups. Anyway, these are the problems one gets when depending on snapshots. Feel free to go back to the latest final version of scala-refactoring if you don't want to deal with these issues.


rorygraves:
Oh cool - thanks Simon - I was just making the same change ;)
I think the right answer for us is to stay in line with the refactoring library.

PR #1329, opened 2016-02-28
short circuits in pattern matchers

### LINKED ISSUE #1328: short circuits
I found out today that, despite what Odersky's book implies, `&` will not short circuit in a pattern match guard.

Therefore the following two uses are too eager

```
git grep '[^&]&[^&]' *.scala
core/src/main/scala/org/ensime/config/package.scala:      if file.isFile & file.getName.endsWith(".scala")
core/src/main/scala/org/ensime/indexer/domain.scala:    case ClassName(op, on) if pack == op & on.startsWith(name) =>
```

fommil:
close #1328


aemoncannon:
Lgtm

PR #1330, opened 2016-02-28
Optimise class prepare handling

greenrd:
This avoids:
- Creating duplicate breakpoints
- O(n^2) walking of a source file

Also fix a bug in JVM attaching that if a class was already loaded but not one of its
anonymous inner classes, a breakpoint might get set only in the former but not the
latter, whereas if neither were loaded at the start, breakpoints would be set in both.

### DIFF
--- a/core/src/it/scala/org/ensime/intg/DebugTest.scala
+++ b/core/src/it/scala/org/ensime/intg/DebugTest.scala
@@ -181,7 +181,6 @@ class DebugTest extends EnsimeSpec
                 activeBreakpoints should contain theSameElementsAs Set(
                   Breakpoint(breakpointsFile, 11), Breakpoint(breakpointsFile, 13)
                 )
-                pendingBreakpoints shouldBe empty
             }
 
             // check clear works again
--- a/core/src/main/scala/org/ensime/core/debug/DebugManager.scala
+++ b/core/src/main/scala/org/ensime/core/debug/DebugManager.scala
@@ -51,13 +51,12 @@ class DebugManager(
 
   def setBreakpoint(file: File, line: Int): Boolean = {
     val applied = maybeVM.exists { vm => vm.setBreakpoint(file, line) }
+    val bp = Breakpoint(file, line)
+    addPendingBreakpoint(bp)
     if (applied) {
-      activeBreakpoints += Breakpoint(file, line)
-      true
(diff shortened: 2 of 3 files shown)


fommil:
I think @chipsenkbeil is planning Steve big stuff in this area, probably best if he takes a look

chipsenkbeil:
@greenrd, good find! The debugger API doesn't do this yet because I didn't think of that scenario. Just another reason why it's always a good thing that multiple pairs of eyes are looking at the same part of a codebase. :)

I'd love to have you check out the Scala debugger implementation of pending breakpoint requests (https://github.com/ensime/scala-debugger/blob/master/scala-debugger-api/src/main/scala/org/scaladebugger/api/lowlevel/breakpoints/PendingBreakpointSupport.scala#L126). Since I'm planning to go through and replace all debug calls with the debugger API implementations, I don't want to cause any regressions.

The only test change I've seen you make is that pending breakpoints hang around, presumably to always check for new classes loaded for the same source file. If that's enough to ensure that my changes to use the debugger API don't regress this change, then code looks fine to me for the present Ensime.

Of course, the less changes that go into the debug code before I change everything over, the easier and smoother it will be for me, but I'm just now getting some free time to tackle that problem and don't want to block any commits that will improve Ensime's debugger experience before I get there.

fommil:
looks like the tests are failing for realsies

fommil:
@greenrd ! Ping

fommil:
duplicate breakpoints, interesting. We see multiple breakpoint messages in the integration tests and we have to filter them out. Do you think that's related?

fommil:
@chipsenkbeil is your refactor going to cover all this? If so, perhaps we should close this and you take it as an action in your todo list?

chipsenkbeil:
@fommil, this one wasn't on the list yet. Hence why I was hoping @greenrd could take a look at the debugger api later to see what he thought needed to be done to its pending breakpoint support.

fommil:
I think this one is in the safe hands of @chipsenkbeil so closing for now unless you want to pick it up again.

chipsenkbeil:
This needs to be tackled in the debugger API (Ensime server shouldn't have to deal with this). Issue opened here: https://github.com/ensime/scala-debugger/issues/252

Issue #1293, opened 2016-02-14
remove the old refactor API


fommil:
@lxol is it time? :smile: start with a PR to the emacs client?

Then we'll need the thumbs up from @hedefalk  @jvican and @dickwall

rorygraves:
This is not live in sublime yet.

hedefalk:
:+1:

lxol:
shall we keep the old refactoring as a not-default option or get rid of it all together?

fommil:
remove it completely. This was always an opt-in beta test period since it was such a big change, but I'm really happy with it. We can even remove all the old emacs code for showing the preview of the refactor. Might be a big PR of deletions!

lxol:
ok, I will do that next week

aemoncannon:
+1 and that will be a nice chunk of elisp we can purge...

Valdaer:
This will be blocked until removal of old api in all clients, right?

fommil:
@dickwall any ETAs on the refactor move over?

hedefalk:
It has been removed in Atom just to clarify.

jvican:
We didn't even use it. So it's ok :smile:

rorygraves:
Sublime now users new ractoring api - so this is unblocked and good to go.

rorygraves:
PR already open - waiting for rebase.

PR #1333, opened 2016-03-04
Remove old refactoring api

Valdaer:
Close: https://github.com/ensime/ensime-server/issues/1293

Please add Blocked label.


core/src/main/scala/org/ensime/util/FileUtil.scala
```diff
@@ -84,14 +84,14 @@ object FileUtils {
     }
   }
 
-  def writeDiffChanges(changes: List[FileEdit], cs: Charset): Either[Exception, File] = {
+  def writeDiffChanges(changes: List[FileEdit])(implicit cs: Charset): Either[Exception, File] = {
     //TODO: add support for NewFile and DeleteFile
```
Valdaer: should we create separate issue for that? I haven't yet looked which of our refactorings can create/delete files and in how many cases.

fommil: it's ok to add it as a separate commit here. Since it is implicit, can you perhaps call it implicitly instead of explicitly?

fommil:
interesting, we got this transient again on Windows. Seems maybe we need to do better initialisation of SLF4J.

```
akka.ConfigurationException: Logger specified in config can't be loaded [akka.event.slf4j.Slf4jLogger] due to [akka.event.Logging$LoggerInitializationException: Logger log1-Slf4jLogger did not respond with LoggerInitialized, sent instead [TIMEOUT]]
	at akka.event.LoggingBus$$anonfun$4$$anonfun$apply$1.applyOrElse(Logging.scala:116)
	at akka.event.LoggingBus$$anonfun$4$$anonfun$apply$1.applyOrElse(Logging.scala:115)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
	at scala.util.Failure$$anonfun$recover$1.apply(Try.scala:216)
	at scala.util.Try$.apply(Try.scala:192)
	at scala.util.Failure.recover(Try.scala:216)
	at akka.event.LoggingBus$$anonfun$4.apply(Logging.scala:115)
	at akka.event.LoggingBus$$anonfun$4.apply(Logging.scala:110)

```

core/src/main/scala/org/ensime/util/FileUtil.scala
Valdaer: Sry, I didn't have time last week to respond to this.

In here I meant that TODO entry however after studying our library refatoring possibilities I've come to conclusion that none of the refactorings will create/delete files.

As for implicit charset, it's passed explicitly in this case because of ambiguous implicit values from method parameter and https://github.com/ensime/ensime-server/blob/master/util/src/main/scala/org/ensime/util/file.scala#L31

Valdaer:
I've seen the same exception in couple of other pull request last week and now it was gone in most. 

How is our CI set up? Why we have 2 CI's, how AppVeyor is different from drone? Can I/someone trigger the build for a pull request manually?

fommil:
we're still blocked by sublime.

scalanatortest:
Why would you say that? Didn't you see the latest commit?

scalanatortest:
;)

fommil:
I stand corrected! @Valdaer please rebase

rorygraves:
probably have to release it mind

rorygraves:
@Valdaer  This is unblocked - all editors now use the newer refactoring api - can you rebase and then we are good for merge.

Valdaer:
I rebased twice and messed up the branch.
New PR: https://github.com/ensime/ensime-server/pull/1354

PR #1334, opened 2016-03-06
Add support for `deprecating` annotation.

lukasz-golebiewski:
This PR adds support for `deprecating` annotation

### DIFF
--- a/core/src/main/scala/org/ensime/core/SemanticHighlighting.scala
+++ b/core/src/main/scala/org/ensime/core/SemanticHighlighting.scala
@@ -53,9 +53,15 @@ class SemanticHighlighting(val global: RichPresentationCompiler) extends Compile
         } else if (sym.hasFlag(PARAM)) {
           add(ParamSymbol)
         } else {
+
           if (sym.ownerChain.exists(_.isDeprecated)) {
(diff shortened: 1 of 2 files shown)


fommil:
I think you accidentally removed the old test too :-)

fommil:
Can you please squash into one commit?

marekklis:
Commits were squashed

core/src/main/scala/org/ensime/core/SemanticHighlighting.scala
```diff
@@ -53,9 +53,15 @@ class SemanticHighlighting(val global: RichPresentationCompiler) extends Compile
         } else if (sym.hasFlag(PARAM)) {
           add(ParamSymbol)
         } else {
+
           if (sym.ownerChain.exists(_.isDeprecated)) {
             add(DeprecatedSymbol)
           }
+
+          if (sym.ownerChain.exists(_.annotations.exists(_.atp.toString().endsWith("deprecating")))) {
```
fommil: this might be better as an or condition on the above if statement, to avoid double additions.

fommil:
excellent! Thanks for this :smile:

Issue #1321, opened 2016-02-25
H2 delete is really inefficient


fommil:
Effectively reopening https://github.com/ensime/ensime-server/pull/1255

I'm noticing insane wait times on Windows for ENSIME to reindex a big jar.

@smootoo's fix is to add an index, but a quick hack didn't work due to performance reasons of creating the index (on a string column). Perhaps we can look at making the filenames foreign keys, which should also reduce the memory footprint.

This makes filewatching useless for me.

I was hoping the graphpocalypse would fix this, but its ages away.

smootoo:
Is there a way for me to reproduce this? Maybe a test that exercises the problematic code paths?

fommil:
Yup, I'll try to create one. For now manual test is the run ENSIME on a big compiled project and then `sbt clean`

fommil:
@smootoo I'll try to make a reproducible test case for this today or tomorrow. If you don't see anything by the end of today, please pester me about it. Now descending into the last of imaginary things ... https://github.com/fommil/imaginary-friend

fommil:
oh drat, I totally forgot to do this. I'll try to find this week as I managed to break the back of the imaginary friend stuff.

smootoo:
I had a quick look at this on a running server and I think there might be some things that can be done to speed things up. Will need to play more.

fommil:
fantastic! Do you still need me to put together an automated test case?

What I was going to do was to create an integration test for a project with loads of dependencies (maybe 100) and a module with loads of target classfiles (say 10,000). After indexing (which will take a couple of minutes), save the H2 for future use. Then we simulate removing the classes all at once.

The test should be more dramatic if run on Windows, or a slow hard drive. I have a Windows 7 VirtualBox image with sbt/java/emacs all preloaded if you'd like access to it?

I'm pretty sure a more judicious use of foreign keys, instead of using `String` keys everywhere, would be far better.

Automating this would hugely slow down and complicate our CI, so I'm not sure I'd want it to run all the time, but as a local test it could be ok (and also then normalising the timings of before / after any fixes are more reliable if restricted to one machine).

fommil:
Hmm, I wonder if running ensime on `ensime-server` itself would generate a good enough example.  Then it should be a simple case of saving the `.ensime_cache` after everything is all committed for a fully compiled `ensime-server` and then firing off delete messages for every single `.class` file that is in the database, in a random order, to the `SearchService`.

smootoo:
[...]

PR #1338, opened 2016-03-08
Add foreign key to FqnSymbols table and use cascade delete for file r…

smootoo:
This should help with https://github.com/ensime/ensime-server/issues/1321

inSetBind will bind positional parameters rather than having deletes use literal strings
Cascade deletes mean fewer round trips to the database to achieve the same result

@fommil the foreign key implicitly adds the index that seemed to slow things down for you last time I tried to fix this. You should test it again. My tests suggest no measurable slow-down on inserts.

A bigger win would come from noticing that directories were deleted on a ```clean``` and then doing something like ```delete from fileChecks where filename like 'dir%'``` instead of a filewatcher deleting a file at a time.
The server knows, because I see it logging that ```a watched base was deleted```. But it is much more work to wire that through.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -63,12 +63,8 @@ class DatabaseService(dir: File) extends SLF4JLogging {
   def removeFiles(files: List[FileObject])(implicit ec: ExecutionContext): Future[Int] =
     db.run {
       val restrict = files.map(_.getName.getURI)
-
-      (
-        fqnSymbols.filter(_.file inSet restrict).delete
-        andThen
-        fileChecks.filter(_.filename inSet restrict).delete
-      )
+      // Deletion from fqnSymbols relies on fk cascade delete action
+      fileChecks.filter(_.filename inSetBind restrict).delete
     }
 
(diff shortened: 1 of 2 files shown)


fommil:
Oh wow, this is really great! I'm on holiday this week, will look in more detail when I get back on Saturday, hopefully I'll be at the hack day.

fommil:
I was expecting a really huge change, this is like magic of the best variety.

fommil:
Did you mean to close?

smootoo:
No. I was just testing to see if the AppVeyor fail was due to this commit (with a force push to my repo). Forgot that this would close the PR.
Will be pushing again in a few mins with the version bumped to 1.2 (as per your comment)

fommil:
In the error about debugging messages or logback initialisation nonsense inside akka? I've seen both a few times recently.

rorygraves:
yeah its the logging init one.

fommil:
I have no idea what's going on there. Maybe we need to manually load some logback classes and send a log message before starting any actors or something? I thought we were doing that in EnsimeSpec already.

We have bigger and faster appveyor machines now, which could be the reason we're seeing this more often.

I consider this green if that's the only problem.

fommil:
Really looking forward to using this one!

PR #1340, opened 2016-03-11
Gc idle

rabbitonweb:
Thank you for contributing to ENSIME!

If this is your first time submitting a PR, this is what you can expect:

1. You'll have to agree to the licences we use. You'll get a link to follow in a moment.
2. Your code will be automatically tested. That can take as long as an hour before the results come back. You can read the `.drone.yml` and `appveyor.yml` files to see what is being run.
3. We'll try to review your code and give feedback quickly.
4. Please squash your commits into independent units. We use the commit log as our changelog so it is good if your commit messages are clear and concise.
5. You can request an ENSIME sticker from `@fommil`, be sure to ask if you want one!
6. If everything goes silent, please ping us.

Now, delete this line and everything above it, then tell us which tickets this PR closes (please don't reference tickets directly from your commit messages).

### DIFF
--- a/core/src/main/scala/org/ensime/core/IdleScheduler.scala
+++ b/core/src/main/scala/org/ensime/core/IdleScheduler.scala
@@ -1,0 +1,28 @@
+package org.ensime.core
+
+import akka.actor.{ Actor, ActorRef, Cancellable }
+import scala.concurrent.duration._
+
+case object Idle
+/*
+Actor that sends Idle  message every times seconds to Project actor.
+ */
+class IdleScheduler(project: ActorRef) extends Actor {
+
+  import context.{ dispatcher, system }
+
+  var idleCallback: Option[Cancellable] = None
+
+  override def preStart(): Unit = {
+    idleCallback = Some(context.system.scheduler.schedule(5.seconds, 5.seconds, project, Idle))
+  }
+
+  override def postStop(): Unit = {
+    idleCallback.foreach(_.cancel)
+  }
+
+  def receive: Receive = {
+    case _ =>
+  }
+
+}
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -2,7 +2,6 @@
 // Licence: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.core
 
-import akka.actor.Actor._
 import akka.actor._
 import akka.event.LoggingReceive
 import akka.event.LoggingReceive.withLabel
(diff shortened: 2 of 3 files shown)


rabbitonweb:
Missing unit tests

core/src/main/scala/org/ensime/core/IdleScheduler.scala
```diff
@@ -0,0 +1,28 @@
...
+Actor that sends Idle  message every times seconds to Project actor.
+ */
+class IdleScheduler(project: ActorRef) extends Actor {
+
+  import context.{ dispatcher, system }
+
+  var idleCallback: Option[Cancellable] = None
+
+  override def preStart(): Unit = {
+    idleCallback = Some(context.system.scheduler.schedule(5.seconds, 5.seconds, project, Idle))
```
fommil: can you put this in the server config? 5 seconds is way too short as well, probably every 1 minute.

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -55,6 +55,20 @@ class Project(
   }
   private val classfileWatcher = context.actorOf(Props(new ClassfileWatcher(config, searchService :: reTypecheck :: Nil)), "classFileWatcher")
 
+  var seenWorkSinceIdle = true
+  //  var lastMessageReceived = System.currentTimeMillis()
```
fommil: commented code, please remove

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -55,6 +55,20 @@ class Project(
...
+
+  def processIdleMessage(): Unit = {
+    if (seenWorkSinceIdle)
+      seenWorkSinceIdle = false
+    else
+      System.gc()
```
fommil: you need to run finalizers as well, see https://github.com/fommil/sbt-big-project/blob/master/src/main/scala/BigProjectSettings.scala#L106-L108

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -114,7 +128,13 @@ class Project(
...
   private var rechecking: Cancellable = _
 
-  def handleRequests: Receive = withLabel("handleRequests") {
+  def idleMonitor(r: Receive)(implicit context: ActorContext): Receive = {
+    case i @ Idle =>
+      processIdleMessage(); r(i)
+    case i => seenWorkMessage; r(i)
+  }
+
+  def handleRequests: Receive = withLabel("handleRequests")(idleMonitor {
```
fommil: the more canonical thing to do here would be `idleMonitor orElse handleRequests` when you set the listening state of the actor.

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -55,6 +55,20 @@ class Project(
   }
   private val classfileWatcher = context.actorOf(Props(new ClassfileWatcher(config, searchService :: reTypecheck :: Nil)), "classFileWatcher")
 
+  var seenWorkSinceIdle = true
```
fommil: should this not be `false` by default?

fommil:
thanks for putting a PR together! A couple of comments in the commits.

One thing that comes to mind is that we might have a cleaner way of implementing this logic in the `Project`. Instead of recording if we have seen anything since the last idle message, why not just fire off a `Cancellable` every time we get a message?

fommil:
@rabbitonweb ! Ping

and more thoughts on this? I might be picking up something in this area on Monday if I get round to it.

fommil:
I'm thinking that remote GC through REST is probably a better way to do this. Would you like to raise a new PR for that or are you OK if I work on it?

Issue #692, opened 2014-11-10
loose ends in S-Express


fommil:
* SexpData unapply should allow dupes, on a first key wins basis
* SexpList apply should have a tail recursive instead of recurse that could blow up the stack
* ensure that Either/Option have alternative handlers (be strict by default, but allow the lazier "common sense" form)

fommil:
* move back into `org.ensime.sexp` (in prep for separation)
* option not to write out case class fields that are `nil` (@aemoncanon requested)

fommil:
* upgrading scalacheck reveals lots of bugs

fommil:
example of a transient failure https://travis-ci.org/ensime/ensime-server/builds/52514899

aemoncannon:
Still seeing the recursion fails occasionally in sexplist apply

fommil:
I might use shapless in sexpress in the same way its used in the JSON stuff... all the hard work has been done, this should be a much simpler task.

fommil:
re the tail recursion failure. Yeah, I see this in really big files now. That's probably because we're trying to respond with a really big list of semantic information. Bugger.

fommil:
another transient http://fommil.com/github.com/ensime/ensime-server/shapeless_upgrade/e31a41db3e97b28d2183c5b2b322b3138ff50c1e

PR #1344, opened 2016-03-12
Bugfix 692

idanov:
This PR fixes:
 - `SexpList` apply should have a tail recursive instead of recurse that could blow up the stack
 - option not to write out case class fields that are `nil`

These can be found in issue #692 
@iyadi @konradwudkowski

### DIFF
--- a/s-express/src/test/scala/org/ensime/sexp/SexpSpec.scala
+++ b/s-express/src/test/scala/org/ensime/sexp/SexpSpec.scala
@@ -4,6 +4,8 @@ package org.ensime.sexp
 
 import org.ensime.util.EnsimeSpec
 
+import scala.util.{ Success, Try }
+
 class SexpSpec extends EnsimeSpec {
 
   val foostring = SexpString("foo")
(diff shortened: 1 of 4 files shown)


s-express/src/test/scala/org/ensime/sexp/SexpSpec.scala
```diff
@@ -40,6 +42,12 @@ class SexpSpec extends EnsimeSpec {
     }
   }
 
+  it should "support a list with 1000000 elements" in {
+    val data = List.fill(1000000)(SexpChar('a'))
+    val res = Try(SexpList(data))
+    res shouldBe a[Success[Sexp]]
```
fommil: the CI doesn't like this line due to erasure. maybe do a pattern match on the Try?

fommil:
I'll close this and raise a PR with your test slightly tweaked.

fommil:
thanks! Merged in a different PR where I made the one line fix.

idanov:
@fommil great :)

PR #1345, opened 2016-03-12
fixup 1344

### LINKED ISSUE #692: loose ends in S-Express

fommil:
cleanup #1344 which closes #692

### DIFF
--- a/s-express/src/main/scala/org/ensime/sexp/Sexp.scala
+++ b/s-express/src/main/scala/org/ensime/sexp/Sexp.scala
@@ -57,9 +57,8 @@ object SexpNumber {
 object SexpList {
   def apply(els: Sexp*): Sexp = apply(els.toList)
 
-  def apply(els: List[Sexp]): Sexp = els match {
-    case Nil => SexpNil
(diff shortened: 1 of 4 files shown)


fommil:
drone failure is transient

aemoncannon:
Quick question: does the nil skipping also skip empty list?

aemoncannon:
That is, does it omit a key if the key's value is an empty sexplist? (I'm hoping it does).

aemoncannon:
That can actually save a lot of traffic.

fommil:
Yep, but it's not enabled by default yet.

fommil:
I'll see if its easy to update this...

fommil:
looking easy, I'll raise a PR after my dinner (only time I've been able to write any code today, hack day was insane. 8 new users and almost 5 PRs. wow)

PR #1348, opened 2016-03-12
Unit tests have application.conf that allow test time factor to be ap…

smootoo:
…plied on appveyor.

Remove search-service-dispatcher from it:test application.conf

This should help with the transient failures on Appveyor that have this cause in FileWatcherSpec
"Cause: java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.ensime.indexer.Added"


fommil:
excellent! First green light on Windows since they moved from the magnetic tapes to stone tablets.

PR #1350, opened 2016-03-13
TEST ONLY - Debugging investigation

rorygraves:
This code is to investigate the appveyor failures only


fommil:
you can turn off drone by deleting the `.drone.yml` BTW :stuck_out_tongue:

rorygraves:
Interesting, this one appears to have failed because of the logger:

https://ci.appveyor.com/project/fommil/ensime-server-circ6/build/1.0.40

rorygraves:
@fommil @smootoo Another interesting one - seems like 
a timeout on actor system shutdown triggering the logger failure - https://ci.appveyor.com/project/fommil/ensime-server-circ6/build/1.0.41

rorygraves:
Success - (well a successful failed test) 

Looks 
```
1559228374 - DebugStart
18854763 - DebugThreadStartEvent: 1  main
1559228374 - DebugThreadStartEvent: 347  Attach Listener
```

@chipsenkbeil This might be of interested to you.
There are a number of default JVM threads - 
http://stackoverflow.com/questions/19427339/jvm-core-threads

I'm wondering how we fix this, a couple of options:
1) Filter them out of the notifications (ThreadStartEvent) - so the internal threads do not show up to the debugger.
2) Update the test to be more flexible, the background threads after being started in the background.

Thoughts @fommil  @smootoo?

fommil:
filtering sounds good, I find that when one tries to inspect JVM internals the facade breaks down anyway because the boundary between native and VM is a lie.

chipsenkbeil:
@rorygraves, I already knew about those threads. Didn't know not filtering them would be an issue for Ensime. Pretty trivial to filter. I'm able to see them easily with the REPL.

screen shot 2016-03-13 at 4 13 21 pm

rorygraves:
I was aware of the threads in general.  had not really thought about it and ofc the test runs beautifully locally.  Really shows up on the appveyer side, because its very slow (I'm guessing 1 real core running the tests in parrallel.)

PR #1352, opened 2016-03-16
Always write the port file message

rorygraves:
This is a quick fix - Sublime relies on the log message to notice that the server is ready for connection.
This change ensures the message is always written.

### DIFF
--- a/server/src/main/scala/org/ensime/server/PortUtil.scala
+++ b/server/src/main/scala/org/ensime/server/PortUtil.scala
@@ -19,11 +19,12 @@ object PortUtil extends SLF4JLogging {
   def writePort(cacheDir: File, port: Int, name: String): Unit = {
     val portFile = cacheDir / name
     if (!portFile.exists()) {
-      log.info("creating port file: " + portFile)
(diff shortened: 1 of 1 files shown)


server/src/main/scala/org/ensime/server/PortUtil.scala
```diff
@@ -19,11 +19,12 @@ object PortUtil extends SLF4JLogging {
   def writePort(cacheDir: File, port: Int, name: String): Unit = {
     val portFile = cacheDir / name
     if (!portFile.exists()) {
-      log.info("creating port file: " + portFile)
       portFile.createNewFile()
     }
 
     portFile.deleteOnExit() // doesn't work on Windows
     portFile.writeString(port.toString)
+    // Some clients grep the log waiting for this file to be written - so always write the log message.
```
fommil: "Some" clients? I think we need to be pointing the finger of blame squarely on Sublime :stuck_out_tongue:

rorygraves:
@fommil  Please review - its one of the windows transients thats failing - this will make the windows sublime startup much more stable.

PR #1354, opened 2016-03-18
Removal of old refactoring Api.

Valdaer:
Closes: https://github.com/ensime/ensime-server/issues/1293

Updated and fixed branch for: https://github.com/ensime/ensime-server/pull/1333

### DIFF
--- a/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
@@ -3,15 +3,18 @@
 package org.ensime.core
 
 import java.io.File
+import java.nio.charset.Charset
 import java.text.SimpleDateFormat
 import java.util.Date
 import org.ensime.api._
 import org.ensime.fixture._
 import org.ensime.util.EnsimeSpec
+import org.scalatest.Assertions
 
 class RefactoringHandlerSpec extends EnsimeSpec
     with IsolatedAnalyzerFixture
-    with RichPresentationCompilerTestUtils {
+    with RichPresentationCompilerTestUtils
+    with RefactoringHandlerTestUtils {
 
   val encoding = "UTF-16"
   def original = EnsimeConfigFixture.EmptyTestProject.copy(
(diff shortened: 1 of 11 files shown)


fommil:
Next time you get into rebase trouble, just ping us on gitter and we'll help :smile:

fommil:
I'm restarting the Windows CI, it might be a transient

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
```diff
@@ -268,27 +224,31 @@ class RefactoringHandlerSpec extends EnsimeSpec
...
+    val relevantExpectedPart = s"""|@@ -2,3 +2,3 @@
+                                   | trait Foo {
+                                   |-def doIt(
+                                   |+def doItNow(
+                                   | ) = ""
+                                   |""".stripMargin
+    val expectedContents = expectedDiffContent(file.path, relevantExpectedPart)
+
+    diffContents should ===(expectedContents)
+    diffFile.delete()
```
fommil: these `.delete()` calls should be in a `finally` block or they'll not be cleaned up when the tests fail.

core/src/main/scala/org/ensime/core/Refactoring.scala
```diff
@@ -24,7 +25,16 @@ abstract class RefactoringEnvironment(file: String, start: Int, end: Int) {
     procId: Int,
     tpe: RefactorType,
     parameters: refactoring.RefactoringParameters
-  ): Either[RefactorFailure, RefactorEffect] = {
+  )(implicit charset: Charset): Either[RefactorFailure, RefactorDiffEffect] = {
```
fommil: do we really need to pass around the `Charset` everywhere? Are we not free to choose the encoding of the diff file?

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
```diff
@@ -367,23 +325,33 @@ class RefactoringHandlerSpec extends EnsimeSpec
...
+                                     |  \n""".stripMargin
+      val expectedContents = expectedDiffContent(file.path, relevantExpectedPart)
 
       diffContents should ===(expectedContents)
       diffFile.delete()
     }
   }
 }
+
+trait RefactoringHandlerTestUtils extends Assertions {
```
fommil: why does this extend `Assertions`?

fommil:
@Valdaer while we look into the windows problems (we might merge with it red) there are a few review comments for you to look at. Thanks!

Valdaer:
K, I will try to address/fix those in the evening/tommorow.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
Valdaer: because I wanted to do bigger refactoring, backed off from it and I've forgotten to remove it, however now I added another helper method which extracts diff from response and fails a test when response is different type than expected, so now the Assertions trait is needed.

core/src/main/scala/org/ensime/core/Refactoring.scala
Valdaer: Why we would need to choose different `Charset` for the diff file? And on what level? request? this class only? 

I think the `Charset` should be consistent to avoid potential bugs, in all these references in `Refactoring.scala` the `Charset` is taken from https://github.com/ensime/ensime-server/blob/master/core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala#L69

Valdaer:
I rebased to master, I don't know why Needs Rebase label is still there, shouldn't it detect automatically and remove the label?

core/src/main/scala/org/ensime/core/Refactoring.scala
fommil: the charset is used when reading the user's files, but we can choose to use utf-8 for our diff file.

fommil:
@Valdaer ! Ping

fommil:
The Emacs tests are being quite fragile with this lisp connection thing.

fommil:
https://github.com/ensime/ensime-emacs/issues/399 needs fixing, sorry.

fommil:
I think I fixed the build, so I'll rerun and merge if this goes green.

fommil:
finally the build passes, thank you!

Issue #355, opened 2014-06-10
RFE: order ensime-search results by "closest" module/library


fommil:
e.g. hits in current module first, then deps, then non-dep modules, then libraries in the same order.

Also, exact hits of simple names should be preferred over prefixes (there is a trick in Lucene to do this that I once created, I'll have to remember how I did it).

fommil:
this is the search brother of #126

ChrisCoffey:
I'll take a stab at this in the next couple of days

fommil:
https://github.com/ensime/ensime-server/blob/master/core/src/main/scala/org/ensime/indexer/IndexService.scala#L23F

add a boost to classes in your project. Not obvious how that bit of information is passed to the lucene analyser.

fommil:
I think all you need to do is work out an importance score in IndexService.persist and then add a parameter to SimpleLucene.create that encodes it as an index-time boost. I'm a bit embarrassed about lack of unit test coverage of SimpleLucene, because we rely on the integration level testing of IndexService. It's designed to hide the mutable (and incredibly specialist) Java API from Scala. The reason I added the Blocked tag is because I'm refactoring SimpleLucene in the graphpocalypse, but I can resolve any conflicts. Just please try to minimise your changes in this code (i.e. don't refactor anything that isn't necessary)

fommil:
oh I bet IndexService will need to take an `EnsimeConfig` or you'll need to add an extra parameter onto `persist` so that the caller can pass the bit of information.

ChrisCoffey:
Turns out the boosting fix for this ticket actually exacerbates the issue in #1287. I'd like to close my current PR, port the general boosting concept over to a new branch, address the inner-class matching, then come back to this. Any objections to that approach?

fommil:
@ChrisCoffey your PR is good enough to merge as-is, but just make the boost 1.1f :stuck_out_tongue:

fommil:
by as-is, I mean when you implement the `isUser` method

PR #1355, opened 2016-03-20
prioritise the user's classes in search results

ChrisCoffey:
This partially addresses #355, but only provides a partial ordering rather than the total ordering of symbols suggested in the ticket.

The implemented solution first boosts any symbols stored beneath the root director (but not in the cache) to ensure Lucene returns them at the top of it's query results (assuming a normal score-based result order). However, the final result set is actually extracted from the SQL Db, so I also had to implement an ordering for those results based on whether files should have been boosted. It's a little ugly, but does the trick.

The end result is that your own project files which match the search request are returned at the top of the results, followed by everything else.

I had some trouble writing integration tests for this because of how the it config paths are structured. If you'd like me to add an integration test I wouldn't mind a bit of guidance on the integration testing setup.

### DIFF
--- a/api/src/main/scala/org/ensime/api/config.scala
+++ b/api/src/main/scala/org/ensime/api/config.scala
@@ -52,6 +52,9 @@ case class EnsimeConfig(
     }.toSet
   } ++ javaLibs
 
+  val allTargets: Set[File] =
+    subprojects.flatMap(sm => sm.targets ::: sm.testTargets).toSet
+
   def allDocJars: Set[File] = modules.values.flatMap(_.docJars).toSet
 
   def scalaLibrary: Option[File] = allJars.find(_.getName.startsWith("scala-library"))
--- a/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
+++ b/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
@@ -158,6 +158,21 @@ class SearchServiceSpec extends EnsimeSpec
     hits.head shouldBe "org.example2.Baz"
   }
 
+  it should "return user created classes first" in withSearchService { implicit service =>
+    val hits = service.searchClasses("File", 10).map(_.fqn)
+    hits.head should startWith("org.boost.File")
+    hits should contain("java.io.File")
+
+    val hits2 = service.searchClasses("Function1", 25).map(_.fqn)
+    hits2.head should startWith("org.boost.Function1")
+    hits2 should contain("scala.Function1")
+  }
+
+  it should "return user methods first" in withSearchService { implicit service =>
+    val hits = service.searchClassesMethods("toString" :: Nil, 10).map(_.fqn)
+    all(hits) should startWith regex ("org.example|org.boost")
+  }
+
   "exact searches" should "find type aliases" in withSearchService { implicit service =>
     service.findUnique("org.scalatest.fixture.ConfigMapFixture$FixtureParam") shouldBe defined
   }
--- a/core/src/main/scala/org/ensime/indexer/IndexService.scala
+++ b/core/src/main/scala/org/ensime/indexer/IndexService.scala
@@ -84,13 +84,17 @@ class IndexService(path: File) {
 
   private val lucene = new SimpleLucene(path, analyzers)
 
-  def persist(check: FileCheck, symbols: List[FqnSymbol], commit: Boolean): Unit = {
+  def persist(check: FileCheck, symbols: List[FqnSymbol], commit: Boolean, boost: Boolean): Unit = {
     val f = Some(check)
     val fqns: List[Document] = symbols.map {
       case FqnSymbol(_, _, _, fqn, Some(_), _, _, _, _) => MethodIndex(fqn, f).toDocument
       case FqnSymbol(_, _, _, fqn, _, Some(_), _, _, _) => FieldIndex(fqn, f).toDocument
       case FqnSymbol(_, _, _, fqn, _, _, _, _, _) => ClassIndex(fqn, f).toDocument
     }
+    if (boost) {
+      fqns foreach { _.boostText("fqn", 1.1f) }
+    }
+
     lucene.create(fqns, commit)
   }
 
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -34,6 +34,10 @@ class SearchService(
     with FileChangeListener
     with SLF4JLogging {
 
+  private[indexer] def isUserFile(file: FileName): Boolean = {
+    (config.allTargets map (vfs.vfile)) exists (file isAncestor _.getName)
+  }
+
   private val QUERY_TIMEOUT = 30 seconds
 
   /**
@@ -110,8 +114,9 @@ class SearchService(
       val outOfDate = fileCheck.map(_.changed).getOrElse(true)
       if (!outOfDate) Future.successful(None)
       else {
+        val boost = isUserFile(base.getName())
         val check = FileCheck(base)
-        extractSymbolsFromClassOrJar(base).flatMap(persist(check, _, commitIndex = false))
+        extractSymbolsFromClassOrJar(base).flatMap(persist(check, _, commitIndex = false, boost = boost))
       }
     }
 
@@ -146,8 +151,8 @@ class SearchService(
 
   def refreshResolver(): Unit = resolver.update()
 
-  def persist(check: FileCheck, symbols: List[FqnSymbol], commitIndex: Boolean): Future[Option[Int]] = {
-    val iwork = Future { blocking { index.persist(check, symbols, commitIndex) } }
+  def persist(check: FileCheck, symbols: List[FqnSymbol], commitIndex: Boolean, boost: Boolean): Future[Option[Int]] = {
+    val iwork = Future { blocking { index.persist(check, symbols, commitIndex, boost) } }
     val dwork = db.persist(check, symbols)
     iwork.flatMap { _ => dwork }
   }
@@ -313,7 +318,8 @@ class IndexingQueueActor(searchService: SearchService) extends Actor with ActorL
             case Success(_) => indexed.collect {
               case (file, syms) if syms.isEmpty =>
               case (file, syms) =>
-                searchService.persist(FileCheck(file), syms, commitIndex = true).onComplete {
+                val boost = searchService.isUserFile((file.getName))
+                searchService.persist(FileCheck(file), syms, commitIndex = true, boost = boost).onComplete {
                   case Failure(t) => log.error(s"failed to persist entries in $file", t)
                   case Success(_) =>
                 }
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -2,7 +2,6 @@
 // Licence: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.indexer
 
-import java.io.File
 import java.sql.Timestamp
 
 import akka.event.slf4j.SLF4JLogging
@@ -12,6 +11,7 @@ import org.ensime.indexer.DatabaseService._
 
 import org.ensime.api._
 import org.ensime.vfs._
+import org.ensime.util.file._
 
 import scala.concurrent._
 import scala.concurrent.duration._
--- a/core/src/main/scala/org/ensime/indexer/lucene/package.scala
+++ b/core/src/main/scala/org/ensime/indexer/lucene/package.scala
@@ -32,5 +32,9 @@ package object lucene {
 
   implicit class RichDocument(d: Document) {
     def toEntity[T](implicit p: DocumentRecovery[T]) = p.toEntity(d)
+
+    def boostText(f: String, boost: Float) = {
+      d.getField(f).asInstanceOf[TextField].setBoost(boost)
+    }
   }
 }
--- a/testing/simple/src/main/scala/org/boost/Effs.scala
+++ b/testing/simple/src/main/scala/org/boost/Effs.scala
@@ -1,0 +1,5 @@
+package org.boost
+
+case class File()
+
+case class Function1()


core/src/main/scala/org/ensime/indexer/IndexService.scala
```diff
@@ -84,13 +84,19 @@ class IndexService(path: File) {
...
     val fqns: List[Document] = symbols.map {
       case FqnSymbol(_, _, _, fqn, Some(_), _, _, _, _) => MethodIndex(fqn, f).toDocument
       case FqnSymbol(_, _, _, fqn, _, Some(_), _, _, _) => FieldIndex(fqn, f).toDocument
       case FqnSymbol(_, _, _, fqn, _, _, _, _, _) => ClassIndex(fqn, f).toDocument
     }
+    // I don't like casting here, but the alternative is some messy param threading through the
+    // serialization code.
+    // The enormous boost is to ensure we always return user FQNs first
+    if (boost) 
+      fqns foreach { _.getField("fqn").asInstanceOf[TextField].setBoost(1000f) }
```
fommil: yeah, gnarly. Perhaps we could isolate this a little by creating a `RichDocument` with a `boost` method on it? That at least pushes it into the Lucene integration layer instead of our more idiomatic scala code. Is it really not possible to boost a whole document? That must have changed.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -34,6 +34,10 @@ class SearchService(
     with FileChangeListener
     with SLF4JLogging {
 
+  private[indexer] def applyBoost(path: String) =
+    (path startsWith config.rootDir.getAbsolutePath) &&
```
fommil: this is definitely not guaranteed to work. What you need to do is to check if it is contained within any of the modules roots.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -210,13 +215,17 @@ class SearchService(
   /** free-form search for classes */
   def searchClasses(query: String, max: Int): List[FqnSymbol] = {
     val fqns = index.searchClasses(query, max)
-    Await.result(db.find(fqns), QUERY_TIMEOUT) take max
+    (Await.result(db.find(fqns), QUERY_TIMEOUT) take max) sortBy { f =>
+      !applyBoost(f.file drop 7)
```
fommil: why are you using boosts in searching? This should not be necessary.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -210,13 +215,17 @@ class SearchService(
...
+    }
   }
 
   /** free-form search for classes and methods */
   def searchClassesMethods(terms: List[String], max: Int): List[FqnSymbol] = {
     val fqns = index.searchClassesMethods(terms, max)
-    Await.result(db.find(fqns), QUERY_TIMEOUT) take max
+    (Await.result(db.find(fqns), QUERY_TIMEOUT) take max) sortBy { f =>
+      !applyBoost(f.file drop 7)
```
fommil: ditto

core/src/main/scala/org/ensime/indexer/SearchService.scala
fommil: we have a few convenience methods around to check if a file is an ancestor of another, perhaps that might be better (and avoids using the raw `path: String` which is fragile, especially on Windows).

fommil:
Great stuff! I think this needs to go through another round, comments above, but it certainly looks like you've broken the back of the problem.

Can you please add a few tests for this? Maybe create a few classes in the test project with the same name as things that are in the Scala / Java stdlibs and make sure they come out on top. You might want to write them against the current code to convince yourself that they fail without your fix.

core/src/main/scala/org/ensime/indexer/IndexService.scala
ChrisCoffey: Yeah, sadly it looks like `Document` level boosting was dropped with Lucene 4.0. I like the `RichDocument` approach.

core/src/main/scala/org/ensime/indexer/SearchService.scala
ChrisCoffey: in hindsight `applyBoost` is a terrible name, it should be something like `isUserFile`. I'm using the check here to order the `FQN`s we get from the Db, since as far as I could tell there's no guarantee the ordering of fqns that goes into `db.find` is the same that comes out.

core/src/main/scala/org/ensime/indexer/SearchService.scala
ChrisCoffey: sounds great, I'll track those down & move to checking against all of the module roots.

ChrisCoffey:
Thanks @fommil, I'll send this your way soon

core/src/main/scala/org/ensime/indexer/SearchService.scala
fommil: yeah the DB should return hits in the same order that the indexer finds them. If you can prove that's not the case (with a unit test), I'll look at it and fix it (unless you get there first).

fommil:
it's really amazing, I've been able to survive without this for so long, but now that I know it's just around the corner **I WANT IT YESTERDAY ALREADY** :laughing:

ChrisCoffey:
@fommil This is ready for review again

fommil:
darn, you have conflicts

ChrisCoffey:
All better ( I hope). Back to you @fommil

core/src/main/scala/org/ensime/indexer/IndexService.scala
```diff
@@ -84,13 +84,19 @@ class IndexService(path: File) {
...
     val fqns: List[Document] = symbols.map {
       case FqnSymbol(_, _, _, fqn, Some(_), _, _, _, _) => MethodIndex(fqn, f).toDocument
       case FqnSymbol(_, _, _, fqn, _, Some(_), _, _, _) => FieldIndex(fqn, f).toDocument
       case FqnSymbol(_, _, _, fqn, _, _, _, _, _) => ClassIndex(fqn, f).toDocument
     }
+    // I don't like casting here, but the alternative is some messy param threading through the
+    // serialization code.
+    // The enormous boost is to ensure we always return user FQNs first
+    if (boost)
+      fqns foreach { _.boostText("fqn", 1000f) }
```
fommil: we'll have to see what experience tells us but my gut instinct is that this boost is too high. It'll drown out sensible hits from third party libraries with noise from our project.

util/src/main/scala/org/ensime/util/file.scala
```diff
@@ -115,6 +115,9 @@ package object file {
         case t: Throwable => file.getAbsoluteFile
       }
 
+    def descendsFrom(parent: File): Boolean =
+      file.getParent.startsWith(parent.getAbsolutePath)
```
fommil: this needs a test. Also, there is a VFS approach to do this which may avoid the need to write the implementation.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -34,6 +34,9 @@ class SearchService(
     with FileChangeListener
     with SLF4JLogging {
 
+  private[indexer] def isUserFile(file: File): Boolean =
+    config.allRoots exists (r => file descendsFrom r)
```
fommil: I don't understand how this works. `allRoots` are source files, but isn't `file: File` a binary class file or a jar? It would be better to take a solid type here instead of `File`.

core/src/main/scala/org/ensime/indexer/IndexService.scala
ChrisCoffey: I agree. I wasn't sure how extreme to go so started pretty high. I'll drop it down to 10 & see how it looks during a local QA.

util/src/main/scala/org/ensime/util/file.scala
ChrisCoffey: That'd be great, I'll track the VFS approach down.

core/src/main/scala/org/ensime/indexer/SearchService.scala
ChrisCoffey: Agreed. After revisiting this I'm unsure of it myself.

util/src/main/scala/org/ensime/util/file.scala
ChrisCoffey: The only VFS based approach I can find would be transforming each of the module roots into a `FileObject` then using `getChild`. I haven't tested this yet, so please let me know if there's another  way of performing this check that you had in mind.

fommil: `getName` => https://commons.apache.org/proper/commons-vfs/apidocs/org/apache/commons/vfs2/FileName.html#isAncestor(org.apache.commons.vfs2.FileName)

ChrisCoffey: Thanks for the guidance @fommil, I'm not familiar with `vfs2` yet. `isAncestor` or `isDescenant` look to be what I'll need.

fommil: I don't think you need to do any conversions if you pass the right higher level object to your filter.

core/src/main/scala/org/ensime/indexer/SearchService.scala
fommil: I think we need a test that fails on the current codebase to be sure that we're doing the right thing.

util/src/main/scala/org/ensime/util/file.scala
ChrisCoffey: yup, 100% agree. The conversions are a result of my lack of knowledge around the `vfs2` api.

core/src/main/scala/org/ensime/indexer/SearchService.scala
ChrisCoffey: Sure. I'll push a failing branch this afternoon or tomorrow morning(EST).

fommil: just push it on here, no need for a separate branch.

ChrisCoffey: I misspoke. I'll push it here.

fommil:
@ChrisCoffey ! Ping

ChrisCoffey:
The result ordering for the searches in my test cases is the same whether I apply a boost or not (i.e. already the order I want w/o the boost code). I can prove the boost is being applied and impacting search result order if I apply a negative boost to `org.example` classfiles, but unfortunately it doesn't fail on the current codebase.

Would that sort of contrived example suffice?

fommil:
This is probably too do with index insertion order. You could try indexing something again and see if that bumps it to lower priority. I'm not keen on introducing different behaviour just for testing.

ChrisCoffey:
no joy. Tried reindexing a couple classfile with `SearchService.fileChanged()`, but the order is the same. I'm open to other suggestions/ means of manipulating Lucene that can prove the boost is applied (beyond a negative boost)

fommil:
Maybe a good old fashioned unit test inspecting the analysis results?

fommil:
did you actually `touch` the class files when you called fileChanged? Because I think we have logic to skip unchanged files.

fommil:
I might try to attempt a failing test case tonight, no promises :smile:

fommil:
yeah, `findBases` always puts the user stuff first and java stdlib last, so we definitely need to reindex everything.

This is probably why this hasn't been such a high priority for me so far... the user classes implicitly come up first, unless the user recompiles them.

fommil:
well this is a massive hack, and no way in hell will this make it into the repo, but it demonstrates the current code failing!

```diff
Staged changes (9)
modified   core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
@@ -32,7 +32,7 @@ trait IsolatedAnalyzerFixture
     withVFS { implicit vfs =>
       withTestKit { testkit =>
         import testkit._
-        withSearchService { (config, search) =>
+        withSearchService { (config, search, _) =>
           implicit val c = config
           testCode(config, AnalyzerFixture.create(search))
         }
modified   core/src/it/scala/org/ensime/fixture/JavaCompilerFixture.scala
@@ -73,7 +73,7 @@ trait IsolatedJavaCompilerFixture
     withVFS { implicit vfs =>
       withTestKit { testkit =>
         import testkit._
-        withSearchService { (config, search) =>
+        withSearchService { (config, search, _) =>
           withEnsimeConfig { config =>
             val reportHandler = new JavaStoreReporter
             val cc = JavaCompilerFixture.create(config, reportHandler, search)
modified   core/src/it/scala/org/ensime/fixture/RichPresentationCompilerFixture.scala
@@ -64,7 +64,7 @@ trait IsolatedRichPresentationCompilerFixture
     withVFS { implicit vfs =>
       withTestKit { testkit =>
         import testkit._
-        withSearchService { (config, search) =>
+        withSearchService { (config, search, _) =>
           import org.ensime.fixture.RichPresentationCompilerFixture._
           val pc = create(config, search)
           try {
modified   core/src/it/scala/org/ensime/fixture/SearchServiceFixture.scala
@@ -11,10 +11,10 @@ import scala.concurrent.duration._
 
 trait IsolatedSearchServiceFixture extends IsolatedSourceResolverFixture {
 
-  def withSearchService(testCode: (EnsimeConfig, SearchService) => Any)(implicit actorSystem: ActorSystem, vfs: EnsimeVFS): Any = withSourceResolver { (config, resolver) =>
+  def withSearchService(testCode: (EnsimeConfig, SearchService, EnsimeVFS) => Any)(implicit actorSystem: ActorSystem, vfs: EnsimeVFS): Any = withSourceResolver { (config, resolver) =>
     val searchService = new SearchService(config, resolver)
     try {
-      testCode(config, searchService)
+      testCode(config, searchService, vfs)
     } finally {
       Await.ready(searchService.shutdown(), Duration.Inf)
       actorSystem.shutdown()
@@ -42,6 +42,10 @@ trait SharedSearchServiceFixture
   }
 
   def withSearchService(
+    testCode: (EnsimeConfig, SearchService, EnsimeVFS) => Any
+  ): Unit = testCode(_config, _search, _vfs)
+
+  def withSearchService(
     testCode: (EnsimeConfig, SearchService) => Any
   ): Unit = testCode(_config, _search)
 
modified   core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
@@ -2,6 +2,7 @@
 // Licence: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.indexer
 
+import com.google.common.io.Files
 import org.ensime.fixture._
 import org.ensime.util.EnsimeSpec
 import org.ensime.util.file._
@@ -158,7 +159,19 @@ class SearchServiceSpec extends EnsimeSpec
     hits.head shouldBe "org.example2.Baz"
   }
 
-  it should "return user created classes first" in withSearchService { implicit service =>
+  it should "return user created classes first" in withSearchService { (config, service, vfs) =>
+    import org.ensime.vfs._
+    implicit val s = service
+    implicit val c = config
+
+    // have to re-index these user files because they are otherwise
+    // given a slight boost because of the findBases ordering.
+    val userFile = mainTarget / "org/example/File.class"
+    Files.touch(userFile)
+    service.fileChanged(vfs.vfile(userFile))
+
+    Thread.sleep(10000) // how do we do this properly?
+
     val hits = service.searchClasses("File", 10).map(_.fqn)
     hits.head should startWith("org.example.File")
     hits.length shouldBe 8
modified   core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -145,7 +145,6 @@ object DatabaseService {
     // this is just as a helper until we can use more sensible
     // domain objects with slick
     def sourceFileObject(implicit vfs: EnsimeVFS) = source.map(vfs.vfile)
-    def underlyingFile: File = File(file)
 
     // legacy: note that we can't distinguish class/trait
     def declAs: DeclaredAs =
modified   core/src/main/scala/org/ensime/indexer/IndexService.scala
@@ -91,11 +91,8 @@ class IndexService(path: File) {
       case FqnSymbol(_, _, _, fqn, _, Some(_), _, _, _) => FieldIndex(fqn, f).toDocument
       case FqnSymbol(_, _, _, fqn, _, _, _, _, _) => ClassIndex(fqn, f).toDocument
     }
-    // I don't like casting here, but the alternative is some messy param threading through the
-    // serialization code.
-    // The enormous boost is to ensure we always return user FQNs first
     if (boost)
-      fqns foreach { _.boostText("fqn", 1000f) }
+      fqns foreach { _.boostText("fqn", 1000f) } // FIXME reduce the boost
 
     lucene.create(fqns, commit)
   }
modified   core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -34,8 +34,8 @@ class SearchService(
     with FileChangeListener
     with SLF4JLogging {
 
-  private[indexer] def isUserFile(file: File): Boolean =
-    config.allRoots exists (r => file descendsFrom r)
+  private[indexer] def isUserFile(file: File): Boolean = false
+  //config.allRoots exists (r => file descendsFrom r)
 
   private val QUERY_TIMEOUT = 30 seconds
 
@@ -214,14 +214,13 @@ class SearchService(
   /** free-form search for classes */
   def searchClasses(query: String, max: Int): List[FqnSymbol] = {
     val fqns = index.searchClasses(query, max)
-    fqns foreach { f => log.debug(f.fqn) }
-    (Await.result(db.find(fqns), QUERY_TIMEOUT) take max)
+    Await.result(db.find(fqns), QUERY_TIMEOUT) take max
   }
 
   /** free-form search for classes and methods */
   def searchClassesMethods(terms: List[String], max: Int): List[FqnSymbol] = {
     val fqns = index.searchClassesMethods(terms, max)
-    (Await.result(db.find(fqns), QUERY_TIMEOUT) take max)
+    Await.result(db.find(fqns), QUERY_TIMEOUT) take max
   }
 
   /** only for exact fqns */
modified   util/src/main/scala/org/ensime/util/file.scala
@@ -115,9 +115,6 @@ package object file {
         case t: Throwable => file.getAbsoluteFile
       }
 
-    def descendsFrom(parent: File): Boolean =
-      file.getParent.startsWith(parent.getAbsolutePath)
-
   }
 
 }

```

I wonder if perhaps we should put user code to the last of the indexing list **anyway** because there is no good reason to put it at the front. I'll look into that.

fommil:
hmm, I can't get it to fail even if I reorder the `findBases`. Weird. Also, the earlier "refresh when files have changed" test should be doing exactly what we want.

fommil:
I wonder if its an alphabetical thing.

fommil:
`SimpleLucene` has trace-level explanations, I'm going to dig into that.

ChrisCoffey:
I'll take a deeper look tonight when I'm back too. Thanks for the updates

fommil:
some fine logging

```
21:39:21.757 TRACE  o.ensime.indexer.lucene.SimpleLucene - Document<stored,indexed,tokenized,omitNorms,indexOptions=DOCS_ONLY<TYPE:ClassIndex> stored,indexed,tokenized<fqn:zzz.File>> scored 11.206611 = (MATCH) sum of:
  10.126803 = (MATCH) sum of:
    0.08923305 = (MATCH) ConstantScore(fqn:File fqn:File$ fqn:FileBuilder fqn:FileCacheImageInputStream fqn:FileCacheImageOutputStream fqn:FileChannel fqn:FileChannel$MapMode fqn:FileChooserUI fqn:FileDataSource fqn:FileDescriptor fqn:FileDialog fqn:FileDialogPeer fqn:FileFilter fqn:FileHandler fqn:FileIcon16 fqn:FileImageInputStream fqn:FileImageOutputStream fqn:FileImpl fqn:FileInput fqn:FileInputStream fqn:FileLock fqn:FileLockInterruptionException fqn:FileNameExtensionFilter fqn:FileNameMap fqn:FileNotFoundException fqn:FileObject fqn:FileOutput fqn:FileOutputStream fqn:FilePermission fqn:FileProcessLogger fqn:FileProp fqn:FileReader fqn:FileRenderer fqn:FileReporter fqn:FileReporterConfiguration fqn:FileReporterConfiguration$ fqn:FileSystemView fqn:FileTypeMap fqn:FileView fqn:FileWriter fqn:FilenameFilter fqn:Filer fqn:FilerException), product of:
      1.0 = boost
      0.08923305 = queryNorm
    10.03757 = (MATCH) weight(fqn:File in 14) [DefaultSimilarity], result of:
      10.03757 = score(doc=14,freq=1.0 = termFreq=1.0
), product of:
        0.9464053 = queryWeight, product of:
          10.605995 = idf(docFreq=6, maxDocs=103975)
          0.08923305 = queryNorm
        10.605995 = fieldWeight in 14, product of:
          1.0 = tf(freq=1.0), with freq of:
            1.0 = termFreq=1.0
          10.605995 = idf(docFreq=6, maxDocs=103975)
          1.0 = fieldNorm(doc=14)
  1.0798072 = (MATCH) weight(TYPE:ClassIndex in 14) [DefaultSimilarity], result of:
    1.0798072 = score(doc=14,freq=1.0 = termFreq=1.0
), product of:
      0.3104102 = queryWeight, product of:
        3.478646 = idf(docFreq=8718, maxDocs=103975)
        0.08923305 = queryNorm
      3.478646 = fieldWeight in 14, product of:
        1.0 = tf(freq=1.0), with freq of:
          1.0 = termFreq=1.0
        3.478646 = idf(docFreq=8718, maxDocs=103975)
        1.0 = fieldNorm(doc=14)
```

vs

```
21:39:21.748 TRACE  o.ensime.indexer.lucene.SimpleLucene - Document<stored,indexed,tokenized,omitNorms,indexOptions=DOCS_ONLY<TYPE:ClassIndex> stored,indexed,tokenized<fqn:java.io.File>> scored 11.206611 = (MATCH) sum of:
  10.126803 = (MATCH) sum of:
    0.08923305 = (MATCH) ConstantScore(fqn:File fqn:File$ fqn:FileBuilder fqn:FileCacheImageInputStream fqn:FileCacheImageOutputStream fqn:FileChannel fqn:FileChannel$MapMode fqn:FileChooserUI fqn:FileDataSource fqn:FileDescriptor fqn:FileDialog fqn:FileDialogPeer fqn:FileFilter fqn:FileHandler fqn:FileIcon16 fqn:FileImageInputStream fqn:FileImageOutputStream fqn:FileImpl fqn:FileInput fqn:FileInputStream fqn:FileLock fqn:FileLockInterruptionException fqn:FileNameExtensionFilter fqn:FileNameMap fqn:FileNotFoundException fqn:FileObject fqn:FileOutput fqn:FileOutputStream fqn:FilePermission fqn:FileProcessLogger fqn:FileProp fqn:FileReader fqn:FileRenderer fqn:FileReporter fqn:FileReporterConfiguration fqn:FileReporterConfiguration$ fqn:FileSystemView fqn:FileTypeMap fqn:FileView fqn:FileWriter fqn:FilenameFilter fqn:Filer fqn:FilerException), product of:
      1.0 = boost
      0.08923305 = queryNorm
    10.03757 = (MATCH) weight(fqn:File in 12639) [DefaultSimilarity], result of:
      10.03757 = score(doc=12639,freq=1.0 = termFreq=1.0
), product of:
        0.9464053 = queryWeight, product of:
          10.605995 = idf(docFreq=6, maxDocs=103975)
          0.08923305 = queryNorm
        10.605995 = fieldWeight in 12639, product of:
          1.0 = tf(freq=1.0), with freq of:
            1.0 = termFreq=1.0
          10.605995 = idf(docFreq=6, maxDocs=103975)
          1.0 = fieldNorm(doc=12639)
  1.0798072 = (MATCH) weight(TYPE:ClassIndex in 12639) [DefaultSimilarity], result of:
    1.0798072 = score(doc=12639,freq=1.0 = termFreq=1.0
), product of:
      0.3104102 = queryWeight, product of:
        3.478646 = idf(docFreq=8718, maxDocs=103975)
        0.08923305 = queryNorm
      3.478646 = fieldWeight in 12639, product of:
        1.0 = tf(freq=1.0), with freq of:
          1.0 = termFreq=1.0
        3.478646 = idf(docFreq=8718, maxDocs=103975)
        1.0 = fieldNorm(doc=12639)
```

fommil:
all I can say is that I can see the index boost for `org.example` classes with this code, but I'm totally bamboozled about how to create a failing test. It seems that the random sort order just keeps coming out the way the user wants it.

```diff
modified   core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -145,7 +145,6 @@ object DatabaseService {
     // this is just as a helper until we can use more sensible
     // domain objects with slick
     def sourceFileObject(implicit vfs: EnsimeVFS) = source.map(vfs.vfile)
-    def underlyingFile: File = File(file)
 
     // legacy: note that we can't distinguish class/trait
     def declAs: DeclaredAs =
modified   core/src/main/scala/org/ensime/indexer/IndexService.scala
@@ -91,11 +91,8 @@ class IndexService(path: File) {
       case FqnSymbol(_, _, _, fqn, _, Some(_), _, _, _) => FieldIndex(fqn, f).toDocument
       case FqnSymbol(_, _, _, fqn, _, _, _, _, _) => ClassIndex(fqn, f).toDocument
     }
-    // I don't like casting here, but the alternative is some messy param threading through the
-    // serialization code.
-    // The enormous boost is to ensure we always return user FQNs first
     if (boost)
-      fqns foreach { _.boostText("fqn", 1000f) }
+      fqns foreach { _.boostText("fqn", 1.1f) }
 
     lucene.create(fqns, commit)
   }
modified   core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -35,7 +35,8 @@ class SearchService(
     with SLF4JLogging {
 
   private[indexer] def isUserFile(file: File): Boolean =
-    config.allRoots exists (r => file descendsFrom r)
+    file.getCanonicalPath.contains("example")
+  //config.allRoots exists (r => file descendsFrom r)
 
   private val QUERY_TIMEOUT = 30 seconds
 
@@ -214,14 +215,13 @@ class SearchService(
   /** free-form search for classes */
   def searchClasses(query: String, max: Int): List[FqnSymbol] = {
     val fqns = index.searchClasses(query, max)
-    fqns foreach { f => log.debug(f.fqn) }
-    (Await.result(db.find(fqns), QUERY_TIMEOUT) take max)
+    Await.result(db.find(fqns), QUERY_TIMEOUT) take max
   }
 
   /** free-form search for classes and methods */
   def searchClassesMethods(terms: List[String], max: Int): List[FqnSymbol] = {
     val fqns = index.searchClassesMethods(terms, max)
-    (Await.result(db.find(fqns), QUERY_TIMEOUT) take max)
+    Await.result(db.find(fqns), QUERY_TIMEOUT) take max
   }
 
   /** only for exact fqns */
modified   logback-it.xml
@@ -12,6 +12,7 @@
   <root level="INFO">
     <appender-ref ref="file" />
   </root>
+  <logger name="org.ensime.indexer.lucene" level="ALL" />
   <logger name="org.ensime" level="DEBUG" />
   <logger name="akka" level="DEBUG" />
 </configuration>
modified   util/src/main/scala/org/ensime/util/file.scala
@@ -115,9 +115,6 @@ package object file {
         case t: Throwable => file.getAbsoluteFile
       }
 
-    def descendsFrom(parent: File): Boolean =
-      file.getParent.startsWith(parent.getAbsolutePath)
-
   }
 
 }
```

fommil:
i.e. I'm admitting that we don't need a failing test case right now :smile: (don't commit the logging changes)

ChrisCoffey:
Haha, I'm glad to hear it! I was going nuts this morning trying to find a
simple failing case :)

ChrisCoffey:
@fommil any idea what the 2.10.6 drone failure is? My [local compiile](https://gist.github.com/ChrisCoffey/01f9af0069f5dff24885) didn't have any code warnings.

fommil:
Yeah I created a ticket for it in ensime-emacs

ChrisCoffey:
okay, this is ready for review again in that case.

api/src/main/scala/org/ensime/api/config.scala
```diff
@@ -52,6 +52,9 @@ case class EnsimeConfig(
     }.toSet
   } ++ javaLibs
 
+  def allTargets: Set[File] =
```
fommil: this can be a val, to avoid recomputation

core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -158,6 +158,22 @@ class SearchServiceSpec extends EnsimeSpec
     hits.head shouldBe "org.example2.Baz"
   }
 
+  it should "return user created classes first" in withSearchService { implicit service =>
+    val hits = service.searchClasses("File", 10).map(_.fqn)
+    hits.head should startWith("org.example.File")
+    hits.length shouldBe 8
```
fommil: asserting on length is going to be very fragile across systems, just assert the first hit and that the `java.io.File` result is in the response *somewhere* (likewise below)

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -176,8 +176,12 @@ class BasicWorkflow extends EnsimeSpec
               BasicTypeInfo("Blue$", DeclaredAs.Object, "org.example.Blue$", List(), List(), Some(_)),
               BasicTypeInfo("CaseClassWithCamelCaseName", DeclaredAs.Class, "org.example.CaseClassWithCamelCaseName", List(), List(), Some(_)),
               BasicTypeInfo("CaseClassWithCamelCaseName$", DeclaredAs.Object, "org.example.CaseClassWithCamelCaseName$", List(), List(), Some(_)),
+              BasicTypeInfo("File", DeclaredAs.Class, "org.example.File", List(), List(), Some(_)),
```
fommil: if you put your test file in a different package, you won't need to change this test. I recommend doing that, e.g. `org/user/mythings.scala`

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -34,6 +34,11 @@ class SearchService(
     with FileChangeListener
     with SLF4JLogging {
 
+  private[indexer] def isUserFile(file: FileName): Boolean = {
+    //may want to cache the mapped results. 
+    (config.allTargets map { vfs.vfile }) exists { file isAncestor _.getName }
```
fommil: when using the syntax sugar for function application, always use `(...)` instead of `{...}` because there are some corner cases about how the desugaring works.

fommil: no need to cache, that should be handled at the layer beneath us already (of course, it isn't, but that's because VFS is crap...)

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -110,8 +115,9 @@ class SearchService(
       val outOfDate = fileCheck.map(_.changed).getOrElse(true)
       if (!outOfDate) Future.successful(None)
       else {
+        val boostFile = isUserFile(base.getName())
```
fommil: `boostFile` implies that it is a `File`. Maybe prefer just `boost`

fommil: nice optimisation btw, this means it only needs to calculate for the jar. It would be good to do something similar for the directories, but that would mean pushing the logic one more level up.

fommil: (or maybe we should be passing that containing directory information along anyway, it would help to reduce `String` duplication and reduce the database size)

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -313,7 +319,8 @@ class IndexingQueueActor(searchService: SearchService) extends Actor with ActorL
             case Success(_) => indexed.collect {
               case (file, syms) if syms.isEmpty =>
               case (file, syms) =>
-                searchService.persist(FileCheck(file), syms, commitIndex = true).onComplete {
+                val boostFile = searchService.isUserFile((file.getName))
```
fommil: ditto `boostFile` vs `boost`

fommil:
almost there! Just a couple of minor changes.

core/src/main/scala/org/ensime/indexer/SearchService.scala
ChrisCoffey: Happy to move away from `{}` block syntax, but can you share a link or example where curly braces fall down? I understand the single vs. multi-line behavior between the two.

core/src/main/scala/org/ensime/indexer/SearchService.scala
ChrisCoffey: v2 perhaps?

ChrisCoffey:
blocked on the usages at point issue?

fommil:
this is weird. It's like the `Retry` is being ignored.

ChrisCoffey:
anything I should look into?

fommil:
nah, I think it's my fault. Sorry about all this.

ChrisCoffey:
no worries. I want to understand the build setup better, so I'll ping you once this finally lands.

ChrisCoffey:
@fommil Victory!!?

fommil:
I think blind luck on the build, I still have things to fix there. I'll merge quickly while it's in a good mood

Issue #1298, opened 2016-02-16
print tree at point


fommil:
would be very useful for debugging and working around presentation compiler issues

aemoncannon:
+1

I was thinking it'd be cool to have a namespace for experimental rpcs --
XGetTree or whatever.

rorygraves:
To add details - 
Add an API call which defers down to ```showRaw``` on the compiler API.
This would need a companion ticket on emacs/etc to implement the client side rendering.

Thought from @fommil  'There is another way of visualising too, which shows the pseudo scala code'

sugakandrey:
This seems easy enough, going to give it a shot. Also, could someone provide a simple use case example for this?

fommil:
Use case: a macro developer, wanting to print out the AST of the code at point.

or, ENSIME developer wanting to see what the AST looks like in the presentation compiler wanting to do the same.

PR #1356, opened 2016-03-20
Implemented ASTAtPointReq, which returns raw representation of the tree at point

sugakandrey:
Closes:  #1298

There is a slight issue with `showRaw` flags addressed in comment.

### DIFF
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -265,6 +265,17 @@ case class ExpandSelectionReq(file: File, start: Int, end: Int) extends RpcAnaly
  */
 case class StructureViewReq(fileInfo: SourceFileInfo) extends RpcAnalyserRequest
 
+/**
+ * Responds with `ASTInfo`
+ *
+ * @param file source
+ * @param offset in file to inspect
+ */
+case class AstAtPointReq(
+  file: SourceFileInfo,
+  offset: OffsetRange
+) extends RpcAnalyserRequest
+
 sealed trait RpcDebuggerRequest extends RpcRequest
 
 /**
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -565,3 +565,8 @@ case class StructureViewMember(
   position: SourcePosition,
   members: List[StructureViewMember]
 )
+
+/**
+ * @param ast raw representation of the tree at point
+ */
+case class AstInfo(ast: String) extends RpcResponse
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -178,6 +178,7 @@ class BasicWorkflow extends EnsimeSpec
               BasicTypeInfo("CaseClassWithCamelCaseName$", DeclaredAs.Object, "org.example.CaseClassWithCamelCaseName$", List(), List(), Some(_)),
               BasicTypeInfo("Foo", DeclaredAs.Class, "org.example.Foo", List(), List(), Some(_)),
               BasicTypeInfo("Foo$", DeclaredAs.Object, "org.example.Foo$", List(), List(), Some(_)),
+              BasicTypeInfo("Qux", DeclaredAs.Class, "org.example.Qux", List(), List(), Some(_)),
               BasicTypeInfo("Test1", DeclaredAs.Class, "org.example.Test1", List(), List(), None),
               BasicTypeInfo("Test1$", DeclaredAs.Object, "org.example.Test1$", List(), List(), None),
               BasicTypeInfo("Test2", DeclaredAs.Class, "org.example.Test2", List(), List(), None),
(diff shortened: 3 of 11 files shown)


api/src/main/scala/org/ensime/api/incoming.scala
```diff
@@ -265,6 +265,24 @@ case class ExpandSelectionReq(file: File, start: Int, end: Int) extends RpcAnaly
  */
 case class StructureViewReq(fileInfo: SourceFileInfo) extends RpcAnalyserRequest
 
+/**
+ * Responds with `ASTInfo`
+ *
+ * @param file source
+ * @param offset in file to inspect
+ */
+case class ASTAtPointReq(
```
fommil: can you please capitalise like `Ast`? (java/scala convention)

api/src/main/scala/org/ensime/api/incoming.scala
```diff
@@ -265,6 +265,24 @@ case class ExpandSelectionReq(file: File, start: Int, end: Int) extends RpcAnaly
...
+  file: Either[File, SourceFileInfo],
```
fommil: no need for `Either`, it is captured in the `SourceFileInfo` type already.

api/src/main/scala/org/ensime/api/incoming.scala
```diff
@@ -265,6 +265,24 @@ case class ExpandSelectionReq(file: File, start: Int, end: Int) extends RpcAnaly
...
+  offset: OffsetRange,
+  //A bunch of flags, used to customize output tree format
+  printTypes: Option[Boolean] = None,
```
fommil: let's omit these and set sensible defaults on the server side. They change in scala releases so we want to isolate our API from scalac internals.

api/src/main/scala/org/ensime/api/outgoing.scala
```diff
@@ -565,3 +565,8 @@ case class StructureViewMember(
   position: SourcePosition,
   members: List[StructureViewMember]
 )
+
+/**
+ * @param ast raw representation of the tree at point
+ */
+case class ASTInfo(ast: String) extends RpcResponse
```
fommil: ditto for naming conventions.

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -213,6 +214,22 @@ class BasicWorkflow extends EnsimeSpec
           expectMsgPF() {
             case RefactorResult(1234, RefactorType.Rename, List(`fooFile`), _) =>
           }
+
+          import scala.reflect.runtime.universe._
+          import scala.reflect.runtime.{ currentMirror => cm }
+          import scala.tools.reflect.ToolBox
+
+          val tree1 = q"List(1, 2, 3).head + 2"
+          val ast1 = showRaw(cm.mkToolBox().typecheck(tree1), printTypes = true)
```
fommil: I'd prefer to hardcode the expected response... it's more reliable than finding it out through the toolbox. We can always have multiple valid expectations to handle changes in the scala version.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -57,7 +57,9 @@ class Analyzer(
   private var settings: Settings = _
   private var reporter: PresentationReporter = _
 
-  protected var scalaCompiler: RichCompilerControl = _
+  //Declared a more specific type in order to get access to RichPresentationCompiler `showRaw` method.
+  //If this is a bad decision, let me know and I will simply write wrapper method in `RichCompilerControl`
+  protected var scalaCompiler: RichCompilerControl with RichPresentationCompiler = _
```
fommil: yeah, would be nice to have it in the RichCompilerControl so that we can keep track of what we're exporting.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -267,6 +269,22 @@ class Analyzer(
...
+      sender ! withExisting(file) {
+        val p = pos(file, offset)
+        scalaCompiler.askLoadedTyped(p.source)
+        val ast = scalaCompiler.locateTree(p)
+        //This is really weird, but if I supply all 6 flags here, compilation fails with type error as
+        //compiler fails to find implicit conversion Option[Boolean] => BoooleanFlag
+        //val rawAst = scalaCompiler.showRaw(ast, printTypes, printIds, printOwners,
+        //                                   printKinds, printMirrors, printPositions)
+        //val rawAst = scalaCompiler.showRaw(ast, true, true, true, true, true, true) (this fails to compile too)
+        val rawAst = scalaCompiler.showRaw(ast, printTypes, printIds, printOwners,
```
fommil: yeah, I've noticed that as well when working on compiler plugins. also it changes across scala versions, there is something really funky going on. Did you notice that named parameters fail too? I think it has something to do with crazy amountsn of method overloading inside scalac.

protocol-swanky/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
-Protocol Version: 0.8.20 (Must match version at ConnectionInfo.protocolVersion)
+Protocol Version: 0.8.21 (Must match version at ConnectionInfo.protocolVersion)
 
 Protocol Change Log:
+  0.8.21
```
fommil: hehe, everybody always forgets to do this. well done :smile:

protocol-swanky/src/main/scala/org/ensime/server/protocol/swank/SwankFormats.scala
```diff
@@ -513,6 +513,8 @@ object SwankProtocolResponse {
   }
   implicit def StructureViewFormat = SexpFormat[StructureView]
 
+  implicit def ASTInfoFormat = SexpFormat[ASTInfo]
```
fommil: I hate this file so much, sorry you had to see it.

testing/simple/src/main/scala/org/example/Foo.scala
```diff
@@ -25,3 +25,7 @@ object Foo extends App {
 case class CaseClassWithCamelCaseName()
 case class Bloo()
 case object Blue
+
+class Qux {
+  List(1, 2, 3).head + 2
+}
```
fommil: needs newline at end

fommil:
cool stuff! Couple of comments that won't take you long to fix. You might want to check the scala 2.10 build, I bet you it fails.

core/src/main/scala/org/ensime/core/Analyzer.scala
sugakandrey: Named parameters fail as well.

api/src/main/scala/org/ensime/api/incoming.scala
sugakandrey: What should sensible flags include? printTypes seems useful, other than that I'm not really sure.

fommil: true for all the things.

fommil:
btw, I was half tempted to suggest that this just `println` the AST, but of course that makes it really hard to test so you've done the right thing.

sugakandrey:
This should solve all of the problems above. PrintIds and printOwners flags in `showRaw` are disabled because they are compilation-unique and thus make testing hard. Also, I doubt they have much use anyways.

sugakandrey:
Hmm, 2.10 build fails with the same implicit conversion problem in `showRaw` 
```
[error] /drone/src/github.com/ensime/ensime-server/core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala:248: overloaded method value showRaw with alternatives:
[error]   (flags: RichCompilerControl.this.FlagSet)String <and>
[error]   (name: RichCompilerControl.this.Name)String <and>
[error]   (any: Any,printTypes: RichCompilerControl.this.BooleanFlag,printIds: RichCompilerControl.this.BooleanFlag,printKinds: RichCompilerControl.this.BooleanFlag,printMirrors: RichCompilerControl.this.BooleanFlag)String
[error]  cannot be applied to (Any, Boolean, Boolean, Boolean, Boolean, Boolean, Boolean)
[error]     showRaw(any, true, false, false, true, true, true)
[error]     ^
[error] one error found
```
Not really sure how to fix this one.

fommil:
@sugakandrey parameters were added in 2.11. I think I solved this in other projects by looking at the type signature of `showRaw` in 2.10 and then just passing in those exact parameters... and ignore the 2.11 extra options. If you really want, you could add a method with 2.10 and 2.11 implementations into `PresentationCompilerBackCompat.scala`

sugakandrey:
@fommil Apparently, not only the flags have changed but the output format is different in 2.10 too. How should I test it then? Should I create version specific file containing expected output?

fommil:
just have both strings and say "either be this or this". I think we use a pattern matcher elsewhere but its up to you how to do it.

sugakandrey:
@fommil Appveyor keeps failing for reasons unreleated to my PR. Other than that everything seems to be working.

fommil:
ignore the windows fails, we know about it

testing/macros/src/main/scala/org/macros/Macros.scala
```diff
@@ -0,0 +1,16 @@
+package org.macros
+
+import scala.language.experimental.macros
+import scala.reflect.macros.blackbox.Context
+
+/**
+ * Created by sugakandrey.
+ */
+object Macros {
```
fommil: what is this file for? Last comment, honest :smile: Remove this file and then somebody (maybe me will merge)

sugakandrey:
@fommil Accidentally added this, my bad. Removed now.

fommil:
wOooooooooooet

PR #1358, opened 2016-03-23
use java7-file-watcher if javaVersion > 1.6

lxol:

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
+++ b/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
@@ -11,6 +11,8 @@ import org.ensime.api._
 import org.ensime.vfs._
 
 import org.ensime.util.file._
+import java.util.UUID
+import scala.util.Properties
 
 trait FileChangeListener {
   def fileAdded(f: FileObject): Unit
@@ -18,6 +20,7 @@ trait FileChangeListener {
   def fileChanged(f: FileObject): Unit
   def baseReCreated(f: FileObject): Unit = {}
   def baseRemoved(f: FileObject): Unit = {}
+  def baseRegistered(): Unit = {}
 }
 
 trait Watcher {
@@ -43,10 +46,38 @@ class ClassfileWatcher(
 
   private val impls =
     if (config.disableClassMonitoring) Nil
-    else config.targetClasspath.map { target =>
-      val (selector, dir, rec) =
-        if (target.isJar) (JarSelector, target.getParentFile, false) else (ClassfileSelector, target, true)
-      new ApachePollingFileWatcher(dir, selector, rec, listeners)
+    else {
+      def buildJava6Watcher() = {
+        config.targetClasspath.map { target =>
+          val (selector, dir, rec) =
+            if (target.isJar)
+              (JarSelector, target.getParentFile, false)
+            else
+              (ClassfileSelector, target, true)
+          if (log.isDebugEnabled())
+            log.debug(s"creating a ApachePollingFileWatcher watcher for $dir")
+          new ApachePollingFileWatcher(dir, selector, rec, listeners).asInstanceOf[Watcher]
+        }
+      }
+      def buildJava7Watcher() = {
+        val jarJava7WatcherBuilder = new JarJava7WatcherBuilder()
+        val classJava7WatcherBuilder = new ClassJava7WatcherBuilder()
+        config.targetClasspath.map { target =>
+          if (target.isJar) {
+            if (log.isDebugEnabled())
+              log.debug(s"creating a Java 7 jar watcher for ${target}")
+            jarJava7WatcherBuilder.build(target, listeners)
+          } else {
+            if (log.isDebugEnabled())
+              log.debug(s"creating a Java 7 class watcher for ${target}")
+            classJava7WatcherBuilder.build(target, listeners)
+          }
+        }
+      }
+      if (Properties.isJava6)
+        buildJava6Watcher()
+      else
+        buildJava7Watcher()
     }
 
   override def receive: Receive = {
@@ -56,7 +87,6 @@ class ClassfileWatcher(
   override def postStop(): Unit = {
     impls.foreach(_.shutdown())
   }
-
 }
 
 class SourceWatcher(
@@ -68,13 +98,36 @@ class SourceWatcher(
 ) extends Watcher with SLF4JLogging {
   private val impls =
     if (config.disableSourceMonitoring) Nil
-    else for {
-      module <- config.modules.values
-      root <- module.sourceRoots
-    } yield {
-      new ApachePollingFileWatcher(root, SourceSelector, true, listeners)
+    else {
+      def buildJava6Watcher() =
+        for {
+          module <- config.modules.values
+          root <- module.sourceRoots
+        } yield {
+          if (log.isDebugEnabled())
+            log.debug(s"creating ApachPollingFileWatcher source watcher for $root")
+          new ApachePollingFileWatcher(root, SourceSelector, true, listeners)
+        }
+
+      def buildJava7Watcher() = {
+        val sourceJava7WatcherBuilder = new SourceJava7WatcherBuilder()
+        for {
+          module <- config.modules.values
+          root <- module.sourceRoots
+        } yield {
+          if (log.isDebugEnabled())
+            log.debug(s"creating a Java 7 source watcher for $root")
+          sourceJava7WatcherBuilder.build(root, listeners)
+        }
+      }
+      if (Properties.isJava6)
+        buildJava6Watcher()
+      else
+        buildJava7Watcher()
     }
+
   override def shutdown(): Unit = impls.foreach(_.shutdown)
+
 }
 
 /**
@@ -153,5 +206,162 @@ private class ApachePollingFileWatcher(
 
   override def shutdown(): Unit = {
     fm.stop()
+  }
+}
+
+trait Java7WatcherBuilder extends SLF4JLogging {
+  import org.ensime.filewatcher.WatcherListener
+  val serviceBuilder = new Java7WatchServiceBuilder()
+  def build(
+    watched: File,
+    listeners: Seq[FileChangeListener]
+  )(
+    implicit
+    vfs: EnsimeVFS
+  ): Watcher = {
+    val watcherId = UUID.randomUUID()
+    serviceBuilder.build(watcherId, watched,
+      listeners.map { l => toWatcherListener(l, watched, watcherId, vfs) })
+  }
+  def toWatcherListener(
+    l: FileChangeListener,
+    baseFile: File,
+    uuid: UUID,
+    vfs: EnsimeVFS
+  ): WatcherListener
+}
+
+class JarJava7WatcherBuilder() extends Java7WatcherBuilder {
+  import org.ensime.filewatcher.WatcherListener
+  override def toWatcherListener(
+    l: FileChangeListener,
+    baseFile: File,
+    uuid: UUID,
+    vfs: EnsimeVFS
+  ) = {
+    new WatcherListener() {
+      override val base = baseFile
+      override val recursive = false
+      override val extensions = JarSelector.include
+      override val watcherId = uuid
+      override def fileCreated(f: File) =
+        l.fileAdded(vfs.vfile(f))
+      override def fileDeleted(f: File) = {}
+      override def fileModified(f: File) =
+        l.fileChanged(vfs.vfile(f))
+      override def baseRegistered(): Unit =
+        l.baseRegistered()
+      override def baseRemoved(): Unit =
+        l.fileRemoved(vfs.vfile(baseFile))
+      override def missingBaseRegistered(): Unit =
+        l.fileAdded(vfs.vfile(baseFile))
+      override def baseSubdirRegistered(f: File): Unit = {}
+      override def proxyRegistered(f: File): Unit = {}
+      override def existingFile(f: File): Unit = {}
+    }
+  }
+
+}
+
+private class SourceJava7WatcherBuilder() extends Java7WatcherBuilder {
+  import org.ensime.filewatcher.WatcherListener
+  override def toWatcherListener(
+    l: FileChangeListener,
+    baseFile: File,
+    uuid: UUID,
+    vfs: EnsimeVFS
+  ) = {
+    new WatcherListener() {
+      override val base = baseFile
+      override val recursive = true
+      override val extensions = SourceSelector.include
+      override val watcherId = uuid
+      @volatile private var notifyExisting = false
+      override def fileCreated(f: File) =
+        l.fileAdded(vfs.vfile(f))
+      override def fileDeleted(f: File) =
+        l.fileRemoved(vfs.vfile(f))
+      override def fileModified(f: File) =
+        l.fileChanged(vfs.vfile(f))
+      override def baseRegistered(): Unit = {
+        notifyExisting = true
+        l.baseRegistered()
+      }
+      override def baseRemoved(): Unit =
+        l.baseRemoved(vfs.vfile(baseFile))
+      override def missingBaseRegistered(): Unit =
+        l.baseReCreated(vfs.vfile(baseFile))
+      override def baseSubdirRegistered(f: File): Unit = {}
+      override def proxyRegistered(f: File): Unit =
+        notifyExisting = true
+      override def existingFile(f: File): Unit =
+        if (notifyExisting)
+          l.fileAdded(vfs.vfile(f))
+    }
+  }
+}
+
+private class ClassJava7WatcherBuilder() extends Java7WatcherBuilder {
+  import org.ensime.filewatcher.WatcherListener
+  override def toWatcherListener(
+    l: FileChangeListener,
+    baseFile: File,
+    uuid: UUID,
+    vfs: EnsimeVFS
+  ) = {
+    new WatcherListener() {
+      override val base = baseFile
+      override val recursive = true
+      override val extensions = ClassfileSelector.include
+      override val watcherId = uuid
+      @volatile private var notifyExisting = false;
+
+      override def fileCreated(f: File) =
+        l.fileAdded(vfs.vfile(f))
+      override def fileDeleted(f: File) =
+        l.fileRemoved(vfs.vfile(f))
+      override def fileModified(f: File) =
+        l.fileChanged(vfs.vfile(f))
+      override def baseRegistered(): Unit = {
+        notifyExisting = true
+        l.baseRegistered()
+      }
+      override def baseRemoved(): Unit =
+        l.baseRemoved(vfs.vfile(baseFile))
+      override def missingBaseRegistered(): Unit =
+        l.baseReCreated(vfs.vfile(baseFile))
+      override def baseSubdirRegistered(f: File): Unit = {}
+      override def proxyRegistered(f: File): Unit =
+        notifyExisting = true
+      override def existingFile(f: File): Unit =
+        if (notifyExisting)
+          l.fileAdded(vfs.vfile(f))
+    }
+  }
+}
+
+class Java7WatchServiceBuilder extends SLF4JLogging {
+  import org.ensime.filewatcher.FileWatchService
+  import org.ensime.filewatcher.WatcherListener
+
+  val fileWatchService: FileWatchService = new FileWatchService
+  def build(
+    watcherId: UUID,
+    base: File,
+    listeners: Seq[WatcherListener]
+  )(implicit vfs: EnsimeVFS) = {
+    if (log.isDebugEnabled())
+      log.debug("watching {}", base)
+
+    trait EnsimeWatcher extends Watcher {
+      val w = fileWatchService.spawnWatcher(watcherId, base, listeners.toSet)
+      override def shutdown(): Unit = {
+        if (log.isDebugEnabled())
+          log.debug("shutdown watcher {}", w.watcherId)
+        w.shutdown()
+      }
+    }
+    val ensimeWatcher = new EnsimeWatcher {}
+    ensimeWatcher
   }
 }
--- a/core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
+++ b/core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
@@ -11,6 +11,8 @@ import org.ensime.util._
 import org.ensime.util.file._
 import org.scalatest._
 import org.scalatest.tagobjects.Retryable
+import scala.util.Properties
+import scala.concurrent.duration._
 
 sealed trait FileWatcherMessage
 final case class Added(f: FileObject) extends FileWatcherMessage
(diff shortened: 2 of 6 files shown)


fommil:
the java 7 code isn't going to be tested as part of CI so I think to be on the safe side we should have a couple of QA testers for this. I'll volunteer, @aemoncannon ?

lxol:
unfortunately all tests failed. @fommil  can you help with integration. I guess this message indicates that something wrong with Sonatype publishing

> Found intransitive dependency (org.ensime:java7-file-watcher_2.11:1.0.0-SNAPSHOT) while publishMavenStyle is true, but Maven repositories
[warn]   do not support intransitive dependencies. Use exclusions instead so transitive dependencies
[warn]   will be correctly excluded in dependent projects.

fommil:
have a look at how I did it in my previous branch https://github.com/fommil/ensime-server/tree/java7-file-watcher

fommil:
failure is

```
[error] missing or invalid dependency detected while loading class file 'FileWatchService.class'.
[error] Could not access term file in package java.nio,
[error] because it (or its dependencies) are missing. Check your build definition for
[error] missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
[error] A full rebuild may help if 'FileWatchService.class' was compiled against an incompatible version of java.nio.
[error] two errors found
```

so looks like a signature thing somewhere.

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -155,3 +167,116 @@ private class ApachePollingFileWatcher(
     fm.stop()
   }
 }
+
+object JarJava7Watcher {
+  lazy val jarWatcher = new Java7Watcher()
```
fommil: I'm not too keen on this as an `object` as it introduces global, mutable, state.

project/EnsimeBuild.scala
```diff
@@ -141,6 +141,7 @@ object EnsimeBuild extends Build {
       unmanagedJars in Compile += JavaTools,
       EnsimeKeys.unmanagedSourceArchives += file(".").getCanonicalFile / "openjdk-langtools/openjdk6-langtools-src.zip",
       libraryDependencies ++= Seq(
+        "org.ensime" %% "java7-file-watcher" % "1.0.0-SNAPSHOT" intransitive () exclude ("commons-logging", "commons-logging"),
```
fommil: hmm, looks like you did it the right way here...

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
lxol: yes that's not ideal, but to reduce the number of threads we do need a small number of global Java7 WatchServices. I din't find a better way to pass them to `SourceWatcher` and `ClassfileWatcher` classes.

fommil: just create an instance in the containing class above and pass in as a value. This is going to cause all kinds of trouble in the integration tests.

lxol:
I removed global java7 watcher objects and still trying to fix `missing or invalid dependency detected while loading class file FileWatchService.class`. 

 I also can not compile  https://github.com/fommil/ensime-server/tree/java7-file-watcher with java6. I got 
> [error] /home/lxol/Documents/projects/tmp/ensime-server/core/src/main/scala/org/ensime/indexer/FileWatchers.scala:181: object FileMonitor is not a member of package org.ensime.filewatcher
[error]   import org.ensime.filewatcher.FileMonitor
[error]          ^
[warn] /home/lxol/Documents/projects/tmp/ensime-server/core/src/main/scala/org/ensime/indexer/FileWatchers.scala:181: Unused import

lxol:
ahh...  java7-file-watcher no longer has FileMonitor class. that's why  https://github.com/fommil/ensime-server/tree/java7-file-watcher  doesn't compile

lxol:
reproduced the the error  in a minimal project
https://github.com/lxol/watcher-sandbox/blob/master/src/main/scala/com/example/Hello.scala

> > compile
[info] Compiling 1 Scala source to /home/lxol/Documents/projects/tmp/minimalscala/target/scala-2.11/classes...
[error] missing or invalid dependency detected while loading class file 'FileWatchService.class'.
[error] Could not access term file in package java.nio,
[error] because it (or its dependencies) are missing. Check your build definition for
[error] missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
[error] A full rebuild may help if 'FileWatchService.class' was compiled against an incompatible version of java.nio.
[error] one error found
[error] (compile:compileIncremental) Compilation failed
[error] Total time: 1 s, completed 24-Mar-2016 20:04:58
>

fommil:
try using FQNs instead of imports maybe?

fommil:
have you put any global state in the classes that it's importing? Because just loading the classes may trigger the execution of that code.

lxol:
FQN doesn't help, checking global state....

fommil:
damn. I used to have to hack around problems like this all the time back in the J2ME days, it's black magic I swear it is.

What about exposed constructor parameters?

fommil:
(analysing the `.class` files with a decompiler might even be needed to see if scala is adding anything static to the class)

lxol:
another suspect is implicit conversion of nio.WatchKey to File.

lxol:
I think I know what is the issue.  Watch events that I use are in fact static variable in java.nio.file.StandardWhatchEventKinds.

fommil:
btw, I'm trying to get Windows tests stable again for you

fommil:
@lxol ! Ping

lxol:
Great. I will continue with the watcher integration on Sunday/Monday, not sure  if I find time before that.

lxol:
The latest java7-file-watcher fixed 2.11.7 tests. 2.10.6 still fails. I found the scala 2.11.x PR that can be relevant. https://github.com/scala/scala/pull/4111 .  btw https://github.com/fommil/ensime-server/tree/java7-file-watcher with the previous java-only watcher behaves exactly the same . It  works fine with 2.11.7 and doesn't compile in 2.10.6

fommil:
Interesting.

lxol:
hmm... I was wrong -  https://github.com/fommil/ensime-server/tree/java7-file-watcher compiles with 2.10.6

fommil:
oh, I think I know. Did the snapshot dependency for 2.10 fail to build?

fommil:
once we're comfortable with the java7-file-watchers jar, I'll make a release to maven central. You're ok with Apache 2 licence, yeah? SBT might be interested in using it as well as us.

lxol:
ensime-server build for 2.10 fails with the error:
>scala.reflect.internal.Types$TypeError: bad symbolic reference. A signature in FileWatchService.class refers to term file
in package java.nio which is not available

2.11 was ok, and yes I am happy with Apache 2 license.

fommil:
I'll it possible we're seeing an old build of the file watcher in 2.10?

lxol:
the error message says that term file in package java.nio is not available. I created  `java.nio.file

lxol:
the error message says that  the term `file` in package` java.nio` is not available. I reproduced the 2.10 error in a minimal project then created a fake   `java.nio.file` package.  That fixed the error. I will try the same workaround in the ensime-server project later today.

fommil:
ok, what about when this runs on Java 7? Will that conflict or will it be ok?

lxol:
I will test that as well.

lxol:
build is okay in my small project for all 4 combinations of scala 2.10, 2.11 and java6, 7 if I create `package java.nio.file`
(2.10 - java6 ) fails without the package. 

I think it has something to do with the bug that was fixed in 2.11.x https://github.com/scala/scala/pull/4111

fommil:
heh, great work! Is CI failing due to that bloody emacs thing? I really need to look at it.

lxol:
that was in my tiny project where I reproduced 2.10 issue. I updated the PR with this ugly hack https://github.com/ensime/ensime-server/pull/1358/commits/1e607022134d5310bd3e5d604b64c66810be6447
waiting now for CI to complete...

fommil:
I bet you get hit by https://github.com/ensime/ensime-emacs/issues/399 :sob: 

I'm swamped with classloader hell right now but as soon as I get a moment I'll try to fix the build.

core/src/main/scala/java/nio/Foo.scala
```diff
@@ -0,0 +1,2 @@
+package java.nio.file {
```
fommil: you can put this in the `scala-2.10` directory only if you want to limit the hackiness.

lxol: ok, will do

fommil:
looks like the new machines don't have enough RAM. Can you please change this line in `project/SensibleSettings.scala`?

```
    // 4 x 1GB = 4GB
    concurrentRestrictions in Global := Seq(Tags.limitAll(4)),
```

to be something like

```
    concurrentRestrictions in Global := Seq(Tags.limitAll(sys.env.get("APPVEYOR").map{_ => 2}.getOrElse(4))),
```

maybe even 1 instead of 2

(sorry, typing on a phone)

fommil:
there is probably a much better helper in `scala.util.Properties` for doing that

fommil:
I'll try to fix the Windows build today. Looks like you got a real failure on the GNU/Linux build.

fommil:
looks like something else crept into the build, breaking it. I'll investigate this evening. *sigh*

fommil:
the windows (Java 7) tests are still failing very badly, but it looks like a real error this time. Server is not coming up.

fommil:
I'll try to run it on Windows tonight

lxol:
Several tests are failing because java7 ClassfileWatcher sends some extra notifications.  It treats exiting files as new which is different form the current watcher.

fommil:
hmm, I thought we went through and fired messages for existing files so I'm surprised that has changed the behaviour. Are we getting duplicate messages maybe?

lxol:
I added logging to the reTypecheck listener in Project.scala  and run failing UnsavedFileTest in both java7 and the current project.  Java7 project triggered `fileAdded` and `baseRecreated` and nothing in the java6 project.

fommil:
Interesting. I'll try to fix the other aspects of the failing build tomorrow and try to look into this also.

fommil:
very sorry, I didn't get round to this (I had a migraine on Saturday and couldn't do anything)

lxol:
I am  trying to fix`JarTargetTest` and `MissingJarTargetTest`. One of them fails depending on the `treatExistingAsNew` flag. This is because the new watcher triggers `addFile` in both cases - when it registers a previously missing jar and an existing jar. The java6 watcher - only missing jars.

fommil:
@lxol don't forget to rebase to master, the CI has gotten a lot faster

lxol:
ok

fommil:
argh, have they lost the history of CLA signers? That's annoying.

lxol:
no, I pushed from a new machine with git not properly configured

lxol:
All green now but I updated several tests to check only the final results and disregard intermediate messages. They can be different on different OSs because of os specific WatchService implementations.

fommil:
oh wow, I'll try to review as a priority. Please ping me every day that I forget.

core/src/it/scala/org/ensime/fixture/ProjectFixture.scala
```diff
@@ -58,12 +58,19 @@ object ProjectFixture extends Matchers {
...
+    else {
+      type Fish = PartialFunction[Any, Boolean]
+      val fishFor3Messages: Fish = {
+        case Broadcaster.Persist(AnalyzerReadyEvent) => true
+        case Broadcaster.Persist(FullTypeCheckCompleteEvent) => true
+        case Broadcaster.Persist(IndexerReadyEvent) => true
+        case _ => false
+      }
+
+      probe.fishForMessage(2.minutes)(fishFor3Messages)
```
fommil: hmm, not too keen on this change. We need to assert that it is those exact three messages that we get. Maybe there is another way to do it (and don't forget to dilate).

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -155,3 +198,202 @@ private class ApachePollingFileWatcher(
...
+    vfs: EnsimeVFS
+  ) = {
+    new WatcherListener() {
+      override val base = baseFile
+      override val recursive = false
+      override val extensions = JarSelector.include
+      override val watcherId = uuid
+
+      override def fileCreated(f: File) = {
+        //log.debug("event: fileCreated fileAdded {}", f)
```
fommil: best not to commit commented code

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
```diff
@@ -131,7 +157,7 @@ abstract class FileWatcherSpec extends EnsimeSpec
       }
     }
 
-  it should "detect removed parent base directory" taggedAs (Retryable) in
+  it should "detect removed parent base directory" in
```
fommil: no longer tagged retryable?

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
```diff
@@ -195,20 +230,39 @@ abstract class FileWatcherSpec extends EnsimeSpec
...
+
             val foo = (dir / "foo.class")
             val bar = (dir / "b/bar.class")
 
             waitForLinus()
 
-            foo.createWithParents() shouldBe true
-            bar.createWithParents() shouldBe true
+            if (!foo.createWithParents()) {
+              Thread.sleep(1000)
```
fommil: do we need the sleeps here?

core/src/it/scala/org/ensime/fixture/ProjectFixture.scala
lxol: The current watcher somehow ignores existing files on startup only. In case of java7 watcher we have an extra `FullTypeCheckCompleteEvent` as a result of finding existing files. I will try to fix that.

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
lxol: ahh.. forgot to put it back. It was 10 min wait to fail and 20 with Retryable

core/src/it/scala/org/ensime/fixture/ProjectFixture.scala
fommil: yeah the current watcher has a hack to go scan everything on startup. So you probably don't need that hack anymore in java 7.

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
lxol: `foo.createWithParents` often fails here on Windows but always succeeds with retry.

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
fommil: you can hardcode the wait times here instead of using the defaults, which are perhaps excessive. I think our CI boxes are more reliable now so we could probably reduce the timeouts generally.

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
fommil: does it need the sleep ?

fommil:
I was using this build tonight (GNU/Linux, Java 7) and all seemed good :smile:

Just a few small comments above, and then I think this is good to go.

lxol:
I am testing a fix for the watcher startup issue. Looks good so far on Linux. I think I need a couple of days for proper testing on all platforms.

lxol:
I think it is okay now. The java7 watcher doesn't detect existing files on startup - exactly as the current watcher. All commented out log statements are being removed as well as unneeded Thread.sleeps in the test.

fommil:
This is so awesome! I'll try to review today unless @rorygraves gets there first. If this looks stone we should consider creating an actual release of the backend.

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
```diff
@@ -140,11 +167,13 @@ abstract class FileWatcherSpec extends EnsimeSpec
         try {
           withClassWatcher(dir) { watcher =>
             // would be better if this was atomic (not possible from JVM?)
-            parent.tree.reverse.foreach(_.delete())
+            waitForBaseRegistered(tk)
 
+            parent.tree.reverse.foreach(_.delete())
```
rorygraves: I've seen this code multiple times - utility method?
Originally there was FileUtils.deleteDirectory - maybe something similar?

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -43,10 +46,33 @@ class ClassfileWatcher(
...
+          if (target.isJar) {
+            log.debug(s"creating a Java 7 jar watcher for ${target}")
+            jarJava7WatcherBuilder.build(target, listeners)
+          } else {
+            log.debug(s"creating a Java 7 class watcher for ${target}")
+            classJava7WatcherBuilder.build(target, listeners)
+          }
+        }
+      }
+      if (Properties.javaVersion.startsWith("1.6")) {
```
rorygraves: I think I would make this a utility method on Properties as well 'isJava6' or similar

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -68,13 +94,35 @@ class SourceWatcher(
...
+        val sourceJava7WatcherBuilder = new SourceJava7WatcherBuilder()
+        for {
+          module <- config.modules.values
+          root <- module.sourceRoots
+        } yield {
+          log.debug(s"creating a Java 7 source watcher for $root")
+          sourceJava7WatcherBuilder.build(root, listeners)
+        }
+      }
+      if (Properties.javaVersion.startsWith("1.6"))
```
rorygraves: Ditto - see comment above.

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -43,10 +46,33 @@ class ClassfileWatcher(
...
+          log.debug(s"creating a ApachePollingFileWatcher watcher for $dir")
+          new ApachePollingFileWatcher(dir, selector, rec, listeners).asInstanceOf[Watcher]
+        }
+      }
+      def buildJava7Watcher() = {
+        val jarJava7WatcherBuilder = new JarJava7WatcherBuilder()
+        val classJava7WatcherBuilder = new ClassJava7WatcherBuilder()
+        config.targetClasspath.map { target =>
+          if (target.isJar) {
+            log.debug(s"creating a Java 7 jar watcher for ${target}")
```
rorygraves: I normally add log.ifDebugEnabled around these - somebody should write a macro that does that automatically, otherwise you are at a minimum creating a thunk for the log message.  (Nitpicking)

rorygraves:
No issue tracker on the java7 watcher so will put comment here:

https://github.com/ensime/java7-file-watcher/blob/master/src/main/scala/org/ensime/filewatcher/FileWatchService.scala#L301

The ```monitorThread match``` statement has no effect - its result is ignored and only the second if block is used.

rorygraves:
https://github.com/ensime/java7-file-watcher/blob/master/src/main/scala/org/ensime/filewatcher/FileWatchService.scala#L63
The match statement seems odd - if a thread exists call start on it - the only point of creation is this method which calls start if not set.   Maybe fail with a warning 'Already started'?

rorygraves:
https://github.com/ensime/java7-file-watcher/blob/master/src/main/scala/org/ensime/filewatcher/FileWatchService.scala#L195
This could be ```while(continueMonitoring)``` rather than ```while(true)``` and ```break```

rorygraves:
https://github.com/ensime/java7-file-watcher/blob/master/src/main/scala/org/ensime/filewatcher/FileWatchService.scala#L324
If you need the thread to stop before calling watchService.shutdown you should use a latch rather than a 20ms sleep

rorygraves:
Ok last comment - https://github.com/ensime/java7-file-watcher/blob/master/src/main/scala/org/ensime/filewatcher/FileWatchService.scala

You have a slightly strange (to me at least) layout style - which is fine.  My personal nitpick is that you are inconsistent about brace usage - match statement cases for example where you mix and match - I personally think you don't need them at all there. 

Please don't get me wrong, overall its good clean code - I'm very picky ;) I can add more comments if you like, but it comes more down to personal preference and taste rather than correctness.

fommil:
we really need https://github.com/ensime/ensime-server/issues/1402

lxol:
@rorygraves  styling was the least priority so far. I will update the code later today.

fommil:
oh, it's scala, I was thinking of the older code. Yeah, there are a lot more `{ ... }` braces than there needs to be.

Something to be aware of is that this kind of code can have problems

```
os.exists { _.recursive }
```

always use `(...)` (round brackets) when using the syntax sugar syntax to avoid hitting corner cases (probably lots of them in Scala puzzlers)

```
os.exists(_.recursive)
```

fommil:
this kind of code

```
    dir.listFiles.filter(_.isFile)
      .foreach { file =>
        {
          listeners filter { _.isWatched(file) } foreach { _.existingFile(file) }
        }
      }
```

could probably be rewritten to use for-comprehensions (more idiomatic). But also it would be a lot more efficient if it used the Java 7 NIO APIs because every call to the OS incurs a stackframe which is really expensive (relatively) on Windows.

fommil:
@lxol did you mean to close?

lxol:
That's strange, I rebased and force pushed half an hour ago. No intention to close...

lxol:
I haven't rebased properly, will fix it now

fommil:
You might have deleted your branch temporarily, github closes when it sees that.

lxol:
It is so easy to choose the wrong base in magit, and it looks like if there is nothing to merge in the pull request, github closes it automatically.

fommil:
@lxol sorry, looks like you need to rebase. Ping us when you think it's ready for merge.

lxol:
I rebased the branch. It looks good to me apart from several known issues.

Default WatchService on OSX has  10 sec polling interval. I haven't played  yet with `SencitivityWatchEventModifier` and native barbaryywatchservice that  @cvogt mentioned on gitter a couple of weeks ago.

Linux watcher could potentially  misses the event during the registration because of the bug in `inotify` based WatchService implementation. ensime massively registers directories during startup but as far as I understand it doesn't use watcher events until startup is finished. So I think we won't be affected by the bug.

fommil:
LGTM

fommil:
This is an epic piece of work, thank you! No doubt it isn't finished yet but it's good enough for me so far and I think we should release it. After a couple of weeks I'll cut a stable version of the watcher lib and put it on maven central.

rorygraves:
Ditto - echoing Sam's comments - this is great stuff LGTM

PR #1360, opened 2016-03-25
Update to scala-refactoring 0.10.0-SNAPSHOT

sschaef:
Just a small behavior change: scala-refactoring now creates (as it did
in the 0.8 era) empty lines between import lists and class declarations.


fommil:
thanks, but we already did this earlier in the day ;-)

I don't know why github doesn't complain about a merge conflict.

sschaef:
Because there are no merge conflicts. My changes are equal to yours and therefore result in a commit with simply no changes. ;)

Issue #1222, opened 2016-01-12
use sbt-catalysts


fommil:
there is lots of stuff that could be ripped out or simplified from our build definitions because its now available in https://github.com/InTheNow/catalysts

InTheNow:
And  https://github.com/InTheNow/sbt-catalysts

fommil:
oh! I'm reading the wrong one.

fommil:
If this is a prerequisite for the Typelevel community build then we need to do it. But it's really non trivial (not Low Hanging Fruit at all) because our build also preps the integration tests, and requires all our scala dependencies to be using sbt-catalysts, so that means we're blocked on spray-json-shapeless and #1398 

I also want to have minimal disruption to my own workflow, so that means SensibleSettings needs to become an AutoPlugin and all the bits that catalysts provides to be removed.

fommil:
No longer exists.

PR #1361, opened 2016-03-26
Use sbt-catalysts with a little cleanup of sbt build files.

Valdaer:
Closes: https://github.com/ensime/ensime-server/issues/1222

-added sbt-catalysts with uses in some placec
-removed some redundant settings
-created `EnsimeDeps` with a dependency map required by sbt-catalysts and moved all common dependency settings there from `EnsimeBuild` and `SensibleSettings`

Unfortunetely for a lot of repeated occurrences of dependencies we can't use sbt-catalysts helper functions because of lack of functionalities in sbt-catalysts:
https://github.com/InTheNow/sbt-catalysts/issues/28
https://github.com/InTheNow/sbt-catalysts/issues/29
https://github.com/InTheNow/sbt-catalysts/issues/30

@fommil please review it and if you will find some time tell me what you think about sbt-catalysts and these aformentioned functionalities

@InTheNow could you look at this too and tell me if maybe I missed something else in sbt-catalysts that we could use here? And please take a look at this issues that I proposed, if you want them and I will find some time I could try to do it.

### DIFF
--- a/project/EnsimeBuild.scala
+++ b/project/EnsimeBuild.scala
@@ -2,13 +2,15 @@ import SonatypeSupport._
 import com.typesafe.sbt.SbtScalariform._
 import java.io._
 import org.ensime.EnsimePlugin.JdkDir
+import org.ensime.ensimeserver.Dependencies._
 import org.ensime.Imports.EnsimeKeys
 import sbt.{ IntegrationTest => It, _ }
 import sbt.Keys._
 import sbtassembly.{ AssemblyKeys, MergeStrategy, PathList }
 import sbtassembly.AssemblyKeys._
 import scala.util.{ Properties, Try }
-import org.ensime.EnsimePlugin.JdkDir
+import org.typelevel.{Dependencies => typelevel}
+import sbtcatalysts.CatalystsPlugin.autoImport._
 import sbtbuildinfo.BuildInfoPlugin, BuildInfoPlugin.autoImport._
 
 object EnsimeBuild extends Build {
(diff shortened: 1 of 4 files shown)


fommil:
I'll be honest. I'm not too keen on this change. I know these build scripts intimately and I'd rather not change them unless there is a huge win for me. I reuse the SensibleSettings in multiple projects and this change limits my ability to reuse the files the way I'd organised them.

What extra do I get from this that we don't currently have, and is there a way to do that which minimises the changes?

fommil:
136 additions and 94 deletions, that's going the wrong way for a clean up :-P

fommil:
Also, we can't depend on any of these typelevel plugins because we're expected to be one of the first applications to ship for dev branches of scalac. We can't wait for all this middleware to be built. A strategic direction for us is to remove Scala dependencies, not add them - we already have far too many.

Sorry for the time you've put into this, but I'm going to close this one. We can keep discussing the idea in the PR comments if you'd like to try and convince me otherwise.

InTheNow:
@Valdaer 

I totally agree with your suggestions re sbt-catalysts and in fact I am re-working the dependency definitions to include the features you mention. I'm also moving to json so it is easier to pass the setting around.

Now that you have some experience in sbt-catalysts, perhaps you would be interested in giving some feedback when I PR the changes?

@fommil Right now, I can I imagine why you see no benefit. The big story starts really with the fact that most people/projects do not have a "sensible settings" file as you do, in part because they just do not have the sbt knowledge that you have.

So imagine sbt-catalysts as: 
- a "sensible settings for Typelevel projects". I'm sure then you can see a benefit for (other) TL projects, but perhaps not here as, well you already have one!
- Then consider that most of these setting were in a text file, such that they could be combined in a CSS fashion. That then means one could, for example, use the base TL settings but override with, say, Shapeless settings, then with ensime settings. ie my project can benefit from your settings, not in TL settings. The goal here is rather than directly set lib versions, to be able to 
 - "use the same as them", otherwise
 - "Use the latest" or even "use the lowest version that is binary compat with the latest", or
 - "Use the one that everyone else uses" or
 - "Use Fommils versions" or .....
- Finally, sbt-catalysts then just adds a load of helper functions that a user would normally copy and paste from some chat channel when asking "How do I...." etc

fommil:
@InTheNow @Valdaer how about starting with `spray-json-shapeless`? It's a much simpler build. The ensime build is quite complex and this breaks the integration tests in ways that you probably didn't expect.

It is also super important that no extra build or runtime dependencies are introduced by depending on this... we're very close to the compiler and we are trying our best to *remove* scala dependencies in order to make that easier. Ideally we'd only depend on shapeless/scalatest/scalacheck (you have no idea how much I want to remove akka).

InTheNow:
@fommil Sorry, I  wasn't clear - I am not suggesting we do the ensime build, rather that @Valdaer might want to look at the changes in sbt-catalysts that will be coming in soon.

I was only trying to explain the concept to you, not trying to convince you to use it :) But the offer to look at spray-json-shapeless is very kind and I would like to take you up on that!!

First step is some more work in sbt-catalysts!

PR #1394, opened 2016-03-31
Thread hygiene in the Debug support

cube2222:
Adding a stopExecution and cleanup after shutdown for MonitorOutput.

### DIFF
--- a/core/src/main/scala/org/ensime/core/debug/MonitorOutput.scala
+++ b/core/src/main/scala/org/ensime/core/debug/MonitorOutput.scala
@@ -15,9 +15,13 @@ private class MonitorOutput(val inStream: InputStream, broadcaster: ActorRef) ex
 
   @volatile var finished = false
 
+  def stopExecution(): Unit = {
+    finished = true
+    this.interrupt()
+  }
+
   override def run(): Unit = {
     val buf = new Array[Char](512)
-
     try {
       var i = in.read(buf, 0, buf.length)
       while (!finished && i >= 0) {
@@ -25,8 +29,11 @@ private class MonitorOutput(val inStream: InputStream, broadcaster: ActorRef) ex
         i = in.read(buf, 0, buf.length)
       }
     } catch {
+      case t: InterruptedException =>
       case t: Throwable =>
         log.info("Exception during execution", t)
+    } finally {
+      in.close()
     }
   }
 }


core/src/main/scala/org/ensime/core/debug/MonitorOutput.scala
```diff
@@ -15,20 +15,24 @@ private class MonitorOutput(val inStream: InputStream, broadcaster: ActorRef) ex
...
       var i = in.read(buf, 0, buf.length)
       while (!finished && i >= 0) {
         broadcaster ! DebugOutputEvent(new String(buf, 0, i))
         i = in.read(buf, 0, buf.length)
       }
+      in.close()
     } catch {
       case t: Throwable =>
         log.info("Exception during execution", t)
+        in.close()
```
fommil: perhaps better in a `finally`?

core/src/main/scala/org/ensime/core/debug/MonitorOutput.scala
```diff
@@ -15,20 +15,24 @@ private class MonitorOutput(val inStream: InputStream, broadcaster: ActorRef) ex
 
   @volatile var finished = false
 
-  // TODO This should have a stop method
+  def stopExecution(): Unit = {
+    finished = true
```
rorygraves: You need to interrupt the thread as well - otherwise it can be blocked not receiving a byte forever.
I then normally add 'interruptedException to the catch block and silently ignore it.

core/src/main/scala/org/ensime/core/debug/MonitorOutput.scala
```diff
@@ -15,11 +15,13 @@ private class MonitorOutput(val inStream: InputStream, broadcaster: ActorRef) ex
 
   @volatile var finished = false
 
-  // TODO This should have a stop method
+  def stopExecution(): Unit = {
+    this.interrupt()
+    finished = true
```
fommil: you probably want to flag `finished = true` before interrupting.

rorygraves: Yep, you definitely do....

cube2222: Done.

fommil:
can you please rebase?

cube2222:
Done the rebase.

fommil:
thanks!

Issue #1373, opened 2016-03-30
Add nice stop method to VMEventManager


asznajder:
`VMEventManager` is a `Thread` that doesn't have any gentle stop method.

rorygraves:
https://github.com/ensime/ensime-server/blob/master/core/src/main/scala/org/ensime/core/debug/VMEventManager.scala#L14

See - https://github.com/ensime/ensime-server/issues/1374 also

PR #1395, opened 2016-03-31
Add nice stop method to VMEventManager

Ma1eusz:
Adding gentle stop for VMEventManager as well as making this class not to extend Thread class. 
See issue #1373

### DIFF
--- a/core/src/main/scala/org/ensime/core/debug/VMEventManager.scala
+++ b/core/src/main/scala/org/ensime/core/debug/VMEventManager.scala
@@ -7,35 +7,47 @@ import com.sun.jdi.VMDisconnectedException
 import com.sun.jdi.event.{ ClassPrepareEvent, VMDisconnectEvent, EventQueue }
 import org.slf4j.LoggerFactory
 
-class VMEventManager(val eventQueue: EventQueue, debugManager: ActorRef) extends Thread {
+class VMEventManager(val eventQueue: EventQueue, debugManager: ActorRef) {
 
   val log = LoggerFactory.getLogger("VMEventManager")
 
-  @volatile var finished = false
-  override def run(): Unit = {
-    while (!finished) {
-      try {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/ensime/core/debug/VMEventManager.scala
```diff
@@ -7,37 +7,43 @@ import com.sun.jdi.VMDisconnectedException
...
-        val it = eventSet.eventIterator()
-        while (it.hasNext) {
-          val evt = it.nextEvent()
-          evt match {
-            case e: VMDisconnectEvent =>
-              finished = true
-            case e: ClassPrepareEvent =>
-              debugManager ! DMClassPrepareEvent(e, eventSet)
-            case _ =>
+  val thread = new Thread() {
```
rorygraves: I would make this private

core/src/main/scala/org/ensime/core/debug/VMEventManager.scala
```diff
@@ -7,37 +7,43 @@ import com.sun.jdi.VMDisconnectedException
...
+    override def run(): Unit = {
+      while (!this.isInterrupted) {
```
rorygraves: I think I would actually keep the finished flag - and then the stop method sets the flag and interrupts.  - that means the other catches just set log something and set the flag.  Feels cleaner to me - but it is stylistic.

rorygraves:
Thanks for the PR.  Can you please update the commit message to include the brief change description?

fommil:
can you please rebase?

Ma1eusz:
@fommil Done

chipsenkbeil:
@fommil shouldn't this have a debugger label on it since it's touching the existing debugger code?

fommil:
I don't tend to add labels to PRs, it's too much effort

fommil:
Thank you!

PR #1400, opened 2016-04-01
Removed TODOs and FIXMEs. Adding new TODO/FIXME will fail CI.

asznajder:
Removed all TODOs and FIXMEs from code. 

I was unsure about two things connected with failing builds that are have TODO/FIXME.

1) Is so simple approach fine? Or maybe should our search be case insensitive? If so, the filter will have to be a little bit smarter since for example `toDoc` phrase appears in code.

2) Where TODO/FIXME are allowed? I bet that `.drone.yml` but what else? Is there any `.git` or `target` directory? In my local directory I could find TODO there.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -7,6 +7,10 @@ build:
     - EMACS=/opt/emacs-24.5/bin/emacs
     - AKKA_TEST_TIMEFACTOR=10
   commands:
+    - if grep -rlE 'TODO|FIXME' . | grep -Ev 'target|.git|.drone.yml' > /dev/null; then
+        echo "Please remove TODO or FIXME. Create an issue at GitHub instead." ;
+        exit 1 ;
+      fi
     - git log | head -n 20
     - sbt ++$SCALA_VERSION gen-ensime ;
       sbt ++$SCALA_VERSION clean test:compile it:compile doc ;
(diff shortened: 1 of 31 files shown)


fommil:
Cool! Can you use git grep instead? Just exclude the drone defn itself.

.drone.yml
```diff
@@ -7,7 +7,7 @@ build:
     - EMACS=/opt/emacs-24.5/bin/emacs
     - AKKA_TEST_TIMEFACTOR=10
   commands:
-    - if grep -rlE 'TODO|FIXME' . | grep -Ev 'target|.git|.drone.yml' > /dev/null; then
+    - if git grep -E 'TODO|FIXME' . | grep -v '.drone.yml' > /dev/null; then
```
fommil: no need for the `.` in `git grep`. It knows where to search. Also, there is a more standard way to ignore files, see http://stackoverflow.com/a/14226610/1041691

fommil: no need to throw anything to `/dev/null` it might be useful to see the matches.

asznajder: Fixed, thanks!

asznajder:
CI builds failed due something not connected with this commit, right?

fommil:
thanks!

fommil:
can you please raise a trivial PR with a TODO in a random file? Just to convince ourselves that it definitely works :smile:

asznajder:
Sure!

PR #1403, opened 2016-04-03
Dummy TODO and FIXME to test CI

asznajder:
We're going to test whether #1400 works fine and PRs with TODO / FIXME will be blocked by CI. That's why, this PR contains both forbidden terms and it should not pass tests.
 
**This PR should never be merged!**


asznajder:
"fatal: Unimplemented pathspec magic '!' in ':!.drone.yml'"

What version of git do we have there?

asznajder:
_fatal: Unimplemented pathspec magic '!' in ':!.drone.yml'_

What version of git do we have there? I'm using 2.7.4 and it works fine on my computer.

asznajder:
Probably replacing `:!` with `:(exclude)` would work.

fommil:
the `git grep` line does look a it unusual. There is a flag to include all the default files so no need for the wildcard.

asznajder:
I can replace `./*` with `.`, but in my opinion it won't help. I checked in git sources when this message is printed and simply looks like it doesn't know that `!` is `:(exclude)`. I can replace these two things in one shot.

fommil:
I put this on gitter but I guess you missed it:

```
if $(git grep -qE "TODO|FIXME" -- `git ls-files | grep -v .drone.yml`) ; then echo YES ; else echo NO ; fi
```

PR #1404, opened 2016-04-03
Fixed not working git grep in .drone.yml

asznajder:
Unfortunately, drone doesn't understand magic paths (`!` for `:(exclude)`) in `git grep`. That's why #1403 showed us that #1400 doesn't really work and needs a fix. That's why, we changed _.drone.yml_ to use the unix `grep` instead of path exclusion. 

Extra: Prettier `git diff` in `.drone_yml` (no more lines counting to find diffs).

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -7,7 +7,7 @@ build:
     - EMACS=/opt/emacs-24.5/bin/emacs
     - AKKA_TEST_TIMEFACTOR=10
   commands:
-    - if git grep -E 'TODO|FIXME' -- './*' ':!.drone.yml'; then
+    - if $(git grep -qE "TODO|FIXME" -- `git ls-files | grep -v .drone.yml`) ; then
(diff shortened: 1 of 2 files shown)


fommil:
ok cool, so we'll see if this goes green and then add a commit which demonstrates TODO breaking the build, then back it out and add another commit showing code formatting also breaking the build, then back that out, then I'll squash merge.

fommil:
@asznajder ok, that was green. Now add a commit with a TODO in it and we'll see it go RED.

asznajder:
Yay! It's red ;)

fommil:
nice! OK, now add a commit that reverts this change and introduces some alternative formatting in a random scala file. It should be red also.

asznajder:
Looks like working

fommil:
cool! Add a commit that reverts the formatting change and I think this is good for merge.

fommil:
oh I just realised you created a fresh PR for this. You could have continued on the original PR.

Thanks!

PR #1406, opened 2016-04-04
Make case classes final

dwijnand:
Note the 1 added TODO to ClassName in core/indexer.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/domain.scala
+++ b/core/src/main/scala/org/ensime/indexer/domain.scala
@@ -35,7 +35,10 @@ case class PackageName(path: List[String]) extends FullyQualifiedName {
   def parent = PackageName(path.init)
 }
 
-case class ClassName(pack: PackageName, name: String) extends FullyQualifiedName with DescriptorType {
+sealed abstract class ClassName extends FullyQualifiedName with DescriptorType {
+  def pack: PackageName
+  def name: String
+
   def contains(o: FullyQualifiedName) = o match {
     case ClassName(op, on) if pack == op && on.startsWith(name) =>
       (on == name) || on.startsWith(name + "$")
@@ -46,28 +49,48 @@ case class ClassName(pack: PackageName, name: String) extends FullyQualifiedName
   def fqnString =
     if (pack.path.isEmpty) name
     else ClassName.cleanupPackage(pack.fqnString + "." + name)
-
-  def internalString =
-    "L" + (if (pack.path.isEmpty) name else pack.path.mkString("/") + "/" + name) + ";"
 }
 
 object ClassName {
-  private val Root = PackageName(Nil)
-  // we consider Primitives to be ClassNames
-  private def Primitive(name: String, desc: String): ClassName = new ClassName(Root, name) {
-    override def fqnString = name
-    override def internalString = desc
+  def apply(pack: PackageName, name: String): ClassName = Impl(pack, name, internalString(pack, name))
+
+  def unapply(c: ClassName): Option[(PackageName, String)] = Some((c.pack, c.name))
+
+  private final case class Impl(pack: PackageName, name: String, internalString: String) extends ClassName {
+    override def toString = s"ClassName($pack,$name)"
   }
 
-  val PrimitiveBoolean = Primitive("boolean", "Z")
-  val PrimitiveByte = Primitive("byte", "B")
-  val PrimitiveChar = Primitive("char", "C")
-  val PrimitiveShort = Primitive("short", "S")
-  val PrimitiveInt = Primitive("int", "I")
-  val PrimitiveLong = Primitive("long", "J")
-  val PrimitiveFloat = Primitive("float", "F")
-  val PrimitiveDouble = Primitive("double", "D")
-  val PrimitiveVoid = Primitive("void", "V")
+  private def internalString(pack: PackageName, name: String): String = {
+    def fallback = "L" + (if (pack.path.isEmpty) name else pack.path.mkString("/") + "/" + name) + ";"
+    if (pack.path.isEmpty)
+      name match {
+        case "boolean" => "Z"
+        case "byte" => "B"
+        case "char" => "C"
+        case "short" => "S"
+        case "int" => "I"
+        case "long" => "J"
+        case "float" => "F"
+        case "double" => "D"
+        case "void" => "V"
+        case _ => fallback
+      }
+    else fallback
+  }
+
+  private val Root = PackageName(Nil)
+  // we consider Primitives to be ClassNames
+  private def Primitive(name: String): ClassName = ClassName(Root, name)
+
+  val PrimitiveBoolean = Primitive("boolean")
+  val PrimitiveByte = Primitive("byte")
+  val PrimitiveChar = Primitive("char")
+  val PrimitiveShort = Primitive("short")
+  val PrimitiveInt = Primitive("int")
+  val PrimitiveLong = Primitive("long")
+  val PrimitiveFloat = Primitive("float")
+  val PrimitiveDouble = Primitive("double")
+  val PrimitiveVoid = Primitive("void")
 
   // must be a single type descriptor
   // strips array reification
--- a/core/src/main/scala/org/ensime/indexer/domain.scala
+++ b/core/src/main/scala/org/ensime/indexer/domain.scala
@@ -25,7 +25,7 @@ sealed trait FullyQualifiedName {
   def fqnString: String
 }
 
-case class PackageName(path: List[String]) extends FullyQualifiedName {
+final case class PackageName(path: List[String]) extends FullyQualifiedName {
   def contains(o: FullyQualifiedName) = o match {
     case PackageName(pn) => pn.startsWith(path)
     case ClassName(p, _) => contains(p)
@@ -114,7 +114,7 @@ object ClassName {
   }
 }
 
-case class MemberName(
+final case class MemberName(
     owner: ClassName,
     name: String
 ) extends FullyQualifiedName {
@@ -126,19 +126,19 @@ sealed trait DescriptorType {
   def internalString: String
 }
 
-case class ArrayDescriptor(fqn: DescriptorType) extends DescriptorType {
+final case class ArrayDescriptor(fqn: DescriptorType) extends DescriptorType {
   def reifier: ClassName = fqn match {
     case c: ClassName => c
     case a: ArrayDescriptor => a.reifier
   }
   def internalString = "[" + fqn.internalString
 }
-case class Descriptor(params: List[DescriptorType], ret: DescriptorType) {
+final case class Descriptor(params: List[DescriptorType], ret: DescriptorType) {
   def descriptorString =
     "(" + params.map(_.internalString).mkString("") + ")" + ret.internalString
 }
 
-case class RawClassfile(
+final case class RawClassfile(
   name: ClassName,
   generics: Option[String],
   superClass: Option[ClassName],
@@ -150,26 +150,26 @@ case class RawClassfile(
   source: RawSource
 )
 
-case class RawSource(
+final case class RawSource(
   filename: Option[String],
   line: Option[Int]
 )
 
-case class RawType(
+final case class RawType(
     fqn: String,
     access: Access
 ) {
   def fqnString = ClassName.cleanupPackage(fqn)
 }
 
-case class RawField(
+final case class RawField(
   name: MemberName,
   clazz: ClassName,
   generics: Option[String],
   access: Access
 )
 
-case class RawMethod(
+final case class RawMethod(
   name: MemberName,
   access: Access,
   descriptor: Descriptor,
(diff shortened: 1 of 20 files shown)


core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -35,6 +35,7 @@ case class PackageName(path: List[String]) extends FullyQualifiedName {
   def parent = PackageName(path.init)
 }
 
+// TODO: Should be final
 case class ClassName(pack: PackageName, name: String) extends FullyQualifiedName with DescriptorType {
```
dwijnand: Here's the added TODO (ran out of time for today).

dwijnand:
```
$ if $(git grep -qE "TODO|FIXME" -- `git ls-files | grep -v .drone.yml`) ; then echo "Please remove TODO or FIXME. Create an issue at GitHub instead." ; exit 1 ; fi
Please remove TODO or FIXME. Create an issue at GitHub instead.
```

zzz, thanks Drone..

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -47,27 +47,34 @@ case class ClassName(pack: PackageName, name: String) extends FullyQualifiedName
     if (pack.path.isEmpty) name
     else ClassName.cleanupPackage(pack.fqnString + "." + name)
 
-  def internalString =
```
dwijnand: Moved the "desc" for primitives into here, removing the need to extend ClassName.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -47,27 +47,34 @@ case class ClassName(pack: PackageName, name: String) extends FullyQualifiedName
...
+    case "void" => "V"
+    case _ => "L" + (if (pack.path.isEmpty) name else pack.path.mkString("/") + "/" + name) + ";"
+  }
 }
 
 object ClassName {
   private val Root = PackageName(Nil)
   // we consider Primitives to be ClassNames
-  private def Primitive(name: String, desc: String): ClassName = new ClassName(Root, name) {
-    override def fqnString = name
```
dwijnand: This was already accounted for.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -47,27 +47,34 @@ case class ClassName(pack: PackageName, name: String) extends FullyQualifiedName
     if (pack.path.isEmpty) name
     else ClassName.cleanupPackage(pack.fqnString + "." + name)
 
-  def internalString =
-    "L" + (if (pack.path.isEmpty) name else pack.path.mkString("/") + "/" + name) + ";"
+  def internalString = name match {
```
fommil: this may be functionally correct but it is going to involve a *lot* of unnecessary string comparison. This code is on the critical path during indexing.

Can we perhaps introduce a new `trait` with `ClassName` and `Primitive` as implementations? Might require refactoring elsewhere to use the new trait instead of `ClassName`.

dwijnand: I am trying to refactor the code, but I wonder: is it feasible to benchmark it, or is such a setup not currently available?

dwijnand: I'm not sure about the modelling.. It's very easy to create a ClassName that's actually a primitive with `ClassName(Root, "Z")`..

fommil: perhaps cleaner, you could introduce a new parameter `internalString` into `ClassName` and then provide an `apply` constructor for the `ClassName` which will construct the longer form `L...;` name.

dwijnand: The `ClassName` constructor and 3 parameter apply would still be exposed (can't be overloaded).

I found a way to split it, in a safe manner. I've pushed updates, have a look.

fommil: three parameter constructor is perfectly fine as a public API. It's actually a cleaner way to model it anyway, and avoids recalculating the `internal` String every time.

dwijnand: It's a derivative value, though - with a strict definition in terms of pack and name. I don't think it should be open to any kind of string.

dwijnand: We could make `internalString` a lazy val if you're concerned with performance.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -35,46 +35,73 @@ final case class PackageName(path: List[String]) extends FullyQualifiedName {
   def parent = PackageName(path.init)
 }
 
-final case class ClassName(pack: PackageName, name: String) extends FullyQualifiedName with DescriptorType {
+sealed abstract class ClassName extends FullyQualifiedName with DescriptorType {
```
fommil: I know this is what I originally suggested, but I think the three parameter case class is actually a lot easier to understand :smile: sorry for messing you around!

core/src/main/scala/org/ensime/indexer/domain.scala
fommil: I thought it was possible to reduce the visibility of the 3 parameter constructor.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -35,46 +35,73 @@ final case class PackageName(path: List[String]) extends FullyQualifiedName {
...
 object ClassName {
+  def apply(pack: PackageName, name: String): ClassName =
+    if (pack.path.isEmpty && primitiveMapping.isDefinedAt(name))
+      Primitive(name, primitiveMapping(name))
+    else
+      NonPrimitive(pack, name)
+
+  def unapply(c: ClassName): Option[(PackageName, String)] = Some((c.pack, c.name))
+
+  private final case class NonPrimitive(pack: PackageName, name: String) extends ClassName {
```
fommil: you could make these top level and make the constructor private to force people to go via the `ClassName` (trait companion).

core/src/main/scala/org/ensime/indexer/domain.scala
dwijnand: Nope, can't overload case class's synthetic apply..

core/src/main/scala/org/ensime/indexer/domain.scala
dwijnand: Making the constructor private, there's also the synthetic apply and the synthetic copy, where invariants can be broken...

core/src/main/scala/org/ensime/indexer/domain.scala
dwijnand: No problem. Done.

dwijnand:
Btw @fommil, if/when this ends up looking good to you, please allow me to rewrite the history: rather than 1 squashed commit I think it's best to separate the concerns of making ~200 case classes final and reworking `ClassName`..

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -46,28 +49,43 @@ case class ClassName(pack: PackageName, name: String) extends FullyQualifiedName
...
-  private val Root = PackageName(Nil)
-  // we consider Primitives to be ClassNames
-  private def Primitive(name: String, desc: String): ClassName = new ClassName(Root, name) {
-    override def fqnString = name
-    override def internalString = desc
+  def apply(pack: PackageName, name: String): ClassName = Impl(pack, name, internalString(pack, name))
+
+  def unapply(c: ClassName): Option[(PackageName, String)] = Some((c.pack, c.name))
+
+  private final case class Impl(pack: PackageName, name: String, internalString: String) extends ClassName {
```
fommil: this is looking good, but I'm not sure we need `Impl`. Can you just make `ClassName` have a private constructor?, e.g. something like

`case class ClassName private(foo: String, ...)  { ... }`

and that forces everybody to go through the two-parameter `ClassName.apply`. You'll still need the two parameter `unapply` (assuming people use it).

rorygraves: Short answer, no - case class constructor cannot be tweaked like that.

dwijnand: You can't redefine the case class apply, and you also have copy to worry about.

dwijnand:
```scala
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.6.0_65).
Type in expressions for evaluation. Or try :help.

scala> :pa
// Entering paste mode (ctrl-D to finish)

final case class Foo private(foo: String, i: Int)

object Foo extends ((String, Int) => Foo) {
  def apply(s: String, i: Int) = if (i == 0) ??? else new Foo(s, i)
}

// Exiting paste mode, now interpreting.

<console>:11: error: method apply is defined twice
  conflicting symbols both originated in file '<console>'
       final case class Foo private(foo: String, i: Int)
                        ^
```

fommil: this works for me

```
case class Foo private (foo: String)
 object Foo {
   def apply(): Foo = new Foo("foo")
 }

Foo()
```

surely you could use that to avoid the need to create `Impl`

dwijnand: But you can't hide the other apply method or the copy method:

```scala
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.6.0_65).
Type in expressions for evaluation. Or try :help.

scala> :pa
// Entering paste mode (ctrl-D to finish)

final case class Foo private(foo: String, i: Int, u: String)

object Foo {
  def apply(s: String, i: Int) = new Foo(s, i, s.toUpperCase)
}

// Exiting paste mode, now interpreting.

defined class Foo
defined object Foo

scala> Foo("a", 1)
res0: Foo = Foo(a,1,A)

scala> Foo("a", 1, "b")
res1: Foo = Foo(a,1,b)

scala> res0.copy(u = "b")
res2: Foo = Foo(a,1,b)
```

dwijnand: Case classes should be thought of a product of its types, with a name, labels for the values it wraps, and the possibility of defining methods (which some argue is wrong). So if its values are derivative, the whole type should be hidden IMO.

fommil: Right, ok, yeah the copy thing makes sense because we always want the value to be recalculated.

It's funny, the original implementation was actually much cleaner even though it involved extending a case class!

dwijnand: While there's definitely some additional complication due to limitations in Scala, I'm pretty sure this fixes `equals` (possibly even `hashCode`, not sure) and `internalString`.

dwijnand: Before:

```scala
scala> org.ensime.indexer.ClassName.fromInternal("boolean").internalString
res0: String = Lboolean;
```

After:

```scala
scala> org.ensime.indexer.ClassName.fromInternal("boolean").internalString
res0: String = Z
```

dwijnand: oh and it (possibly - to be proved!) made `internalString` faster by pre-computing it on construction instead of on use.

dwijnand:
AppVeyor's failure seems spurious:

```
[info] DebugTest:
[info] - should list/clear *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: expected DebugVmSuccess(success), found DebugVmError(1,com.sun.jdi.VMDisconnectedException: connection is closed,error)
[info]   at scala.Predef$.assert(Predef.scala:170)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:339)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:315)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:718)
[info]   at org.ensime.intg.DebugTestUtils$class.withDebugSession(DebugTest.scala:281)
[info]   at org.ensime.intg.DebugTest.withDebugSession(DebugTest.scala:22)
[info]   at org.ensime.intg.DebugTest$$anonfun$3$$anonfun$apply$mcV$sp$3$$anonfun$apply$22$$anonfun$apply$23.apply(DebugTest.scala:133)
[info]   at org.ensime.intg.DebugTest$$anonfun$3$$anonfun$apply$mcV$sp$3$$anonfun$apply$22$$anonfun$apply$23.apply(DebugTest.scala:127)
```

fommil:
ok cool, do your history rewriting if you'd like to and I'll try to remember not to squash.

rorygraves:
I stand corrected on the case class private thing.. ;)

dwijnand:
@fommil done

dwijnand:
@rorygraves Corrected? You were right.

rorygraves:
lol, even better - that will teach me for trying to follow the conversation on my phone ;)

dwijnand:
Hey GitHub fixed the topological ordering of commits that are reverse date ordering.

Shame it comes at the price of losing the whitespace-ignoring feature at the pull request level... :(

dwijnand:
Oh it actually might've just switched to commit date rather than author date (which is always in topological order, at least on one branch).

dwijnand:
appveyor, plz retest

PR #1407, opened 2016-04-05
Simplify ClassName, by using lazy val..

dwijnand:
Not sure why we didn't pursue this avenue..

-11 lines of code look good to me, WDTY?


rorygraves:
Looks good to me too

fommil:
This does seem simpler than the other alternatives. Nice one.

dwijnand:
Updated. Merge-squash away!

fommil:
what? No GPG verification?

Issue #1148, opened 2015-11-13
add the acyclic plugin


fommil:
improve the quality of things: https://github.com/lihaoyi/acyclic

lukasz-golebiewski:
@fommil Do you want to add the `acyclic.file` import to all files/packages?

fommil:
I don't really know how the plugin works but removing acyclic dependencies sounds like a good idea.

fommil:
knowing what is acyclic would be a good first move.

lukasz-golebiewski:
But this requires adding an explicit import to all classes / package objects :/

PR #1410, opened 2016-04-09
Add explicit implicits

dwijnand:


rorygraves:
Cool stuff, but drone build is going to fail:
@dwijnand There are failures - http://fommil.com/ensime/ensime-server/381

```
[error] /drone/src/github.com/ensime/ensime-server/protocol-jerky/src/main/scala/org/ensime/jerk/JerkFormats.scala:46: ambiguous implicit values:
[error]  both object SymbolJsonFormat in trait BasicFormats of type org.ensime.jerk.JerkConversions.SymbolJsonFormat.type
[error]  and value symbolFormat in object JerkConversions of type => spray.json.JsonFormat[Symbol]
[error]  match expected type spray.json.JsonFormat[Symbol]
[error]   val EnsimeServerMessageFormat: RootJsonFormat[EnsimeServerMessage] = cachedImplicit
[error]       ^
```

dwijnand:
Given the types of errors (NPEs -.-) and the number of changes, I think the way forward here is keep the current diff as the target, and rebuild the changes incrementally, testing along the way (unless someone can figure out what broke the build). It'll take me some time, so feel free to close this if it's in the way.

fommil:
I'll have a look, just need to find the time. My super high top priority is https://github.com/fommil/class-monkey so expect me to go to ground until it's finished. Maybe @rorygraves can find some time to have a quick look?

fommil:
but yes, I agree that if you can get as much stuff into an initial PR as possible that would be good, then we can investigate what is causing the problem and beg Miles to help.

fommil:
how very bizarre. I'm seeing a reasonably minimal failure with this

```
s-express/testOnly *BasicFormatsSpec -- -z Long
```

which should help localise the issue. I don't think shapeless can even be blamed for this one.

fommil:
@dwijnand woohoo! Fixed it! I'll send you a PR that you can just merge into here or cherry pick. I'm happy to squash everything into one big commit with your name on it.

dwijnand:
Cool. Thanks for last mileing it so it can go in.

fommil:
@dwijnand well, **almost** fixed it. Still getting problems in the SwankFormatsSpec but I think I know what the problem is now.

fommil:
@dwijnand can you please cherry pick the commit from https://github.com/fommil/ensime-server/commits/explicit-implicits ?

I just ran the unit tests, still getting a failure, and it's probably to do with my quick and dirty `lazyImplicit` method entering into infinite recursion. It really needs the equivalent of the `cachedImplicit` macro but for `Lazy`.

fommil:
asked here about the `lazyImplicit` https://gitter.im/milessabin/shapeless?at=57100bce5cd40114649b654c I think it's bedtime for me.

protocol-jerky/src/main/scala/org/ensime/jerk/JerkFormats.scala
```diff
@@ -19,7 +19,7 @@ private object JerkConversions extends DefaultJsonProtocol with FamilyFormats {
   implicit override def eitherFormat[A: JsonFormat, B: JsonFormat]: JsonFormat[Either[A, B]] = super.eitherFormat[A, B]
   // Note that its not possible to override an object in scala, so we
   // just define a new one that wins the race.
-  implicit val symbolFormat = SymbolJsonFormat
+  implicit val symbolFormat: SymbolJsonFormat.type = SymbolJsonFormat
```
fommil: this is so weird

dwijnand: Weird-wrong-looking or weird-bizarre?

fommil: Weird bizarre. I understand why it's needed though but I wouldn't have realised it before you added it.

dwijnand: It's the most specific type possible. I'm not sure it's what is normally inferred, but this seemed to do the trick.

fommil:
I'm surprised the CI is going green, I got red locally.

fommil:
I think my laptop is just not fast enough to run the 8 tests in parallel :smile: this is great, thanks @dwijnand !

dwijnand:
Oh sweet. Thanks again for helping me get this done.

fommil:
can we expect more rules enforced with scapegoat or whatever it is you used? You might also like #1148

dwijnand:
Yeah I'll probably cycle (no pun intended) back to these kind of improvements over time :)

Issue #1281, opened 2016-02-07
Documentation lookup for cats.data.Xor$$Right


d6y:
I'm using Atom with ENSIME, and not having any joy getting `DocUriAtPointReq` to work with `Xor.Right` (from cats).  

Example project: https://github.com/d6y/xor-scaldoc-lookup

Code:

``` scala
import cats.data.Xor._

object O {

  val r = Right(1)

}
```

And if I try to lookup the doc for `Right` like this...

```
 "req": {
    "typehint":"DocUriAtPointReq",
    "file":"/Users/richard/Developer/tmp/xor-scaldoc-lookup/src/main/scala/example.scala",
    "point":{"from":61,"to":61}},
    "callId": 28
  }
```

...the result is:  `{"callId":28,"payload":{"typehint":"FalseResponse"}}`

Expecting to resolve `Right` to:

```
/docs/cats-core_2.11-0.4.1-javadoc.jar/index.html#cats.data.Xor$$Right
```

This via: https://github.com/ensime/ensime-atom/pull/152#issuecomment-181031838

fommil:
ok, cool. It should be really easy to get an integration test case for this.

1. add whatever jars you want to the integration test project https://github.com/ensime/ensime-server/blob/e0ab76b8c8bbe09f33795d33f924f8ddfd9fc8d8/project/EnsimeBuild.scala#L264-L271
2. confirm what type is being inferred in https://github.com/ensime/ensime-server/blob/e0ab76b8c8bbe09f33795d33f924f8ddfd9fc8d8/core/src/it/scala/org/ensime/core/DocFindingSpec.scala (in my dev branch this is becoming a more general FQN discovery feature)
3. add a test to https://github.com/ensime/ensime-server/blob/e0ab76b8c8bbe09f33795d33f924f8ddfd9fc8d8/core/src/it/scala/org/ensime/core/DocResolverSpec.scala

d6y:
Thanks - are there any pointers on how to run the integration test to see if I have step 2 right? `sbt it:test` runs but I'm missing something else running (presumably the server itself).

fommil:
Check the .drone.yml for what the ci runs. You'd need to run gen-ensime and create the cache dir. We could probably handle the latter case the be honest.

d6y:
[...]

PR #1411, opened 2016-04-10
Fix for #1281 - Documentation lookup for cats.data.Xor$$Right

VlachJosef:
Fix for #1281

In some cases lookup of type return Object instead of Class, for example in
pattern matching:

`case Some(s) => ???`

type of `Some` is object, but there is no documentation for
`object Some`, only for `class Some`, so if we don't find documentation we
are looking for we try to find alternative documentation for class


fommil:
Sorry I thought I'd merged this already!! Great stuff, @d6y will be pleased

d6y:
Yes! Thank you @VlachJosef

Issue #993, opened 2015-06-14
load all files in module


fommil:
We can either have "load all in project" or "load just the open files".

When working on projects with a broken build, it would be useful to be able to "load all files in this module" as an alternative workflow.

fommil:
This would involve creating a new API endpoint for the feature (taking in the name of a module), handled by the Analyser, which would then load all the source files in the current module. Perhaps also the inverse (unload all files from a module).

ChrisCoffey:
This is good to close now that #1413 has landed.

fommil:
oh, yeah. The PR needs the comment "closes #xxx" before github can do it automatically.

PR #1413, opened 2016-04-13
Add Load/Unload Module

ChrisCoffey:
In #993 @fommil suggested taking the module name in with the request & loading the files from there. I couldn't find a straightforward way to get the ensime module name from a given source file, so ended I up adding code to the Analyzer (planning on factoring it out into the `config` package object though) that derives the module for a given file.

### Todo
- Better test coverage
- Factor out duplicate code in the Analyzer

Please let me know if you think this approach has legs.

### DIFF
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -27,6 +27,16 @@ class BasicWorkflow extends EnsimeSpec
           val fooFilePath = fooFile.getAbsolutePath
           val barFile = sourceRoot / "org/example/Bar.scala"
 
+          // typeCheck module
+
+          project ! TypecheckModule("testingSimple")
+          expectMsg(VoidResponse)
+          asyncHelper.expectMsgType[NewScalaNotesEvent]
+          asyncHelper.expectMsgType[FullTypeCheckCompleteEvent.type]
+
+          project ! UnloadModuleReq("testingSimple")
+          expectMsg(VoidResponse)
+
           // trigger typeCheck
           project ! TypecheckFilesReq(List(Left(fooFile)))
           expectMsg(VoidResponse)
@@ -183,7 +193,7 @@ class BasicWorkflow extends EnsimeSpec
               BasicTypeInfo("Test1$", DeclaredAs.Object, "org.example.Test1$", List(), List(), None),
               BasicTypeInfo("Test2", DeclaredAs.Class, "org.example.Test2", List(), List(), None),
               BasicTypeInfo("Test2$", DeclaredAs.Object, "org.example.Test2$", List(), List(), None),
-              BasicTypeInfo("package$", DeclaredAs.Object, "org.example.package$", List(), List(), None),
+              //BasicTypeInfo("package$", DeclaredAs.Object, "org.example.package$", List(), List(), None),
               BasicTypeInfo("package$", DeclaredAs.Object, "org.example.package$", List(), List(), None))) =>
           }
 
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -34,8 +34,14 @@ class BasicWorkflow extends EnsimeSpec
           asyncHelper.expectMsgType[NewScalaNotesEvent]
           asyncHelper.expectMsgType[FullTypeCheckCompleteEvent.type]
 
+          project ! TypeByNameReq("org.example.Bloo")
+          expectMsgType[BasicTypeInfo]
+
           project ! UnloadModuleReq("testingSimple")
           expectMsg(VoidResponse)
+
+          project ! TypeByNameReq("org.example.Bloo")
+          expectMsg(FalseResponse)
 
           // trigger typeCheck
           project ! TypecheckFilesReq(List(Left(fooFile)))
--- a/core/src/main/scala/org/ensime/config/package.scala
+++ b/core/src/main/scala/org/ensime/config/package.scala
@@ -7,11 +7,13 @@ import Predef.{ any2stringadd => _, _ => _ }
 import org.ensime.api._
 import org.ensime.util.file._
 
+import scala.collection.breakOut
+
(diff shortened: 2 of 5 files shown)


fommil:
Cool! This may be useful for many people.

Why can't you just search for the file as a child of one of the source roots on the client side? All the information you need is in the .ensime files.

In fact, I don't know how you've been able to support ensime without this feature client side. The first thing the client needs to do when opening a file is resolve which .ensime file to use for a given file, and that involves guessing and checking against the source roots. Has it been blindly using the first .ensime it finds?

Fyi i need to implement something like this in ensime-sbt for the compileOnly feature as well.

I definitely prefer if the api request had the module name instead of an example file.

ChrisCoffey:
After thinking through your comments, I think I need to do a bit of refactoring in the Vim plugin to make accessing the config easier (we do have this but it's only accessed during the startup process). 

I'm perfectly happy with the API taking a name instead of the file, it certainly simplifies the server code :)

I'll push a cleaned up version of the api w/ some tests this evening.

fommil:
FWIW the Emacs code also needs to be cleaned up in this area.

fommil:
@ChrisCoffey have you found any more time to look at this? :smile:

ChrisCoffey:
@fommil Sorry this sat for so long, was on vacation 😄 . I'll get this cleaned up by tomorrow night.

fommil:
It has conflicts the be merged and you were going to change the protocol to take the name of the module, not a file. N also needs tests ;-)

ChrisCoffey:
@fommil I've rebased

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -183,7 +193,7 @@ class BasicWorkflow extends EnsimeSpec
               BasicTypeInfo("Test1$", DeclaredAs.Object, "org.example.Test1$", List(), List(), None),
               BasicTypeInfo("Test2", DeclaredAs.Class, "org.example.Test2", List(), List(), None),
               BasicTypeInfo("Test2$", DeclaredAs.Object, "org.example.Test2$", List(), List(), None),
-              BasicTypeInfo("package$", DeclaredAs.Object, "org.example.package$", List(), List(), None),
+              //BasicTypeInfo("package$", DeclaredAs.Object, "org.example.package$", List(), List(), None),
```
fommil: is there any reason for this change?

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -27,6 +27,16 @@ class BasicWorkflow extends EnsimeSpec
           val fooFilePath = fooFile.getAbsolutePath
           val barFile = sourceRoot / "org/example/Bar.scala"
 
+          // typeCheck module
+
+          project ! TypecheckModule("testingSimple")
+          expectMsg(VoidResponse)
+          asyncHelper.expectMsgType[NewScalaNotesEvent]
+          asyncHelper.expectMsgType[FullTypeCheckCompleteEvent.type]
```
fommil: you can `expectMsg` on the instance here

core/src/main/scala/org/ensime/config/package.scala
```diff
@@ -10,12 +10,19 @@ import org.ensime.util.file._
...
 
   implicit class RichEnsimeConfig(val c: EnsimeConfig) extends AnyVal {
-    def scalaSourceFiles: Set[File] = for {
-      module: EnsimeModule <- c.modules.values.toSet
-      root <- module.sourceRoots
-      file <- root.tree
-      if file.isFile && file.getName.endsWith(".scala")
-    } yield file
+    def scalaSourceFiles: Set[File] =
+      c.modules.values.toSet.flatMap((m: EnsimeModule) => m.scalaSourceFiles)
```
fommil: you might be able to use `breakOut` here instead of having to call explicit `.toSet`

core/src/main/scala/org/ensime/config/package.scala
```diff
@@ -10,12 +10,19 @@ import org.ensime.util.file._
...
   }
 
+  implicit class RichEnsimeModule(val m: EnsimeModule) extends AnyVal {
+    def scalaSourceFiles: Set[File] = {
+      val s = for {
+        root <- m.sourceRoots
+        file <- root.tree
+        if file.isFile && file.getName.endsWith(".scala")
```
fommil: We have an `isScala` method in `org.ensime.file._` that can be used (missed it when it was added)

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -174,6 +175,20 @@ class Analyzer(
         restartCompiler(keepLoaded = false)
       }
       sender ! VoidResponse
+    case TypecheckModule(moduleName) =>
+      //consider the case of a project with no modules
+      config.modules get (moduleName) foreach {
+        case module =>
+          val files = module.scalaSourceFiles.map(SourceFileInfo(_, None, None)).toList
```
fommil: using `breakOut` here would also be beneficial.

fommil:
LGTM

fommil:
Only a few small comments, if you want to address them just push commits with the changes, otherwise let me know when you're happy.

The only real comment I would have on this is that the testing is a bit thin. Just testing the protocol level isn't really enough. Can you perhaps put in a query that would only work if everything really is loaded?

I'm a bit concerned about the change in the test assertion. Any ideas what is going on there?

Sorry for taking so long to get to this :blush:

ChrisCoffey:
no worries @fommil. I believe the commented out line occurs because we load it twice, and my unload call was removing the initial instance. I'd have to do a deep dive to confirm that assumption however. 

Let me know if you'd like me to.

fommil:
This is looking good. If convenient, can't you push a chance to remove the commented line? It's either case I'll merge when it goes green (I just hate commented code because it always asks a question of why it is there).

Thanks! Nighttime here so it might be 8 hours before the button is pressed.

ChrisCoffey:
Doh, I meant to remove that... I'll push it tonight.

Issue #1414, opened 2016-04-13
Support scala 2.12


fommil:
It is non-trivial to release a Scala 2.12 build of ENSIME because we depend on middleware that must also be available. This is a list of subtasks that are blocking the release:

**If you came here to check on the status of 2.12, but don't have time to dedicate to help us all achieve these goals, please consider [sponsorship](http://ensime.github.io/sponsor/)**

- [x] [java 8](https://github.com/ensime/ensime-server/pull/1639) not as trivial as you'd like to think... thanks Windows
- [x] akka - does not (and will not) have a cross build that works for both scala-2.10/scala-2.11/scala-2.12 at the same time.
- [ ] merge [the graphpocalypse](https://github.com/ensime/ensime-server/issues/1509) which removes a dependency on slick
- [x] scala-refactoring - is not necessarily available for new versions of scala. Ideally we'd make the refactoring logic optional on a per scala version basis (using the `BackCompat` pattern, maybe flag tests with tags).
- [x] scala-debugging - is closely tied to the binary format of scala. Ideally we'd make the debugging logic optional on a per scala version basis (using the `BackCompat` pattern, maybe flag tests with tags)..
- [x] spire / cats / macro-compat - are used in our tests and must be upgraded to a version that is cross build to 2.10/2.11/2.12, which will also involve updating the test assertions and perhaps addressing new bugs that arise
- [x] addressing general bugs that arise as a result of the binary format changes.
- [ ] scalamock could be removed and is only used in two places (although it will still be needed transitively by some of our dependencies).
- [x] we [don't need scalariform anymore](https://github.com/ensime/ensime-server/issues/1622) but it requires some work to reimplement one thing

We're waiting for a binary release of the following (links for non-trivial updates): 
- [x] [akka / testkit](https://github.com/akka/akka/issues/21763)
- [x] scalatest / scalactic
- [x] scalacheck
- [x] [scalamock](https://github.com/paulbutcher/ScalaMock/issues/119) @paulbutcher (unless removed)
- [x] shapeless
- [x] spray-json
- [x] parboiled2
- [x] spray-json-shapeless
- [x] [slick](https://github.com/slick/slick/issues/1572) (unless we remove)
- [x] [scala-refactoring](https://github.com/scala-ide/scala-refactoring/issues/168) @dragos / @sschaef  / @mlangc
- [x] [scalariform](https://github.com/scala-ide/scalariform/issues/232) @sschaef  / @SethTisue
[...]

PR #1416, opened 2016-04-15
akka http upgrade to 2.0.4

bartekkalinka:
Needed in #1414
Caveats:
1. Webserver.scala: in 2.0.4, ContentType.apply method has no version with optional HttpCharset, so I used HttpCharsets.UTF-8.
2. WebSocketBoilerplate.scala: in 2.0.4 Sink.asPublisher takes fanout: Boolean parameter, which decides if publisher supports multiple subscribers or not.  I've put false, not sure if that's correct.

### DIFF
--- a/project/SensibleSettings.scala
+++ b/project/SensibleSettings.scala
@@ -103,7 +103,6 @@ object Sensible {
 
   val scalaModulesVersion = "1.0.4"
   val akkaVersion = "2.3.14"
-  val streamsVersion = "1.0"
   val scalatestVersion = "2.2.6"
   val logbackVersion = "1.7.19"
   val quasiquotesVersion = "2.0.1"
--- a/project/EnsimeBuild.scala
+++ b/project/EnsimeBuild.scala
@@ -167,7 +167,7 @@ object EnsimeBuild extends Build {
       )
 
   val luceneVersion = "4.7.2"
-  val streamsVersion = "1.0"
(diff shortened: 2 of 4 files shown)


project/SensibleSettings.scala
```diff
@@ -103,7 +103,6 @@ object Sensible {
 
   val scalaModulesVersion = "1.0.4"
   val akkaVersion = "2.3.14"
```
fommil: do we also need to bump the akka version?

fommil:
@bartekkalinka you've got some unused imports. We have warnings are fatal enabled in CI (it's disabled locally so it doesn't get in the way).

project/SensibleSettings.scala
bartekkalinka: Actually I chose 2.0.4 version of akka-http/streams for now, because I think it's the highest version that doesn't need akka 2.4.x.  Bumping akka to 2.4.x results for start with following error:
`[error] com.typesafe#config;1.2.1 (needed by [com.typesafe.slick#slick_2.11;3.1.1]) conflicts with com.typesafe#config;1.3.0 (needed by [com.typesafe.akka#akka-actor_2.11;2.4.3])`

fommil:
btw, no need to force push anymore, now that github supports squashing on merge :smile:

fommil:
ok, I'll merge this as it brings us incrementally closer to a version of akka/streams that we need.

PR #1417, opened 2016-04-15
upgrade to recent akka / akka-http

jrudolph:
Here's a first approximation to update to a recent akka/akka-http version 2.4.4 (preliminary to #1414). It comes with all caveats that Akka 2.4 comes with, probably most importantly that it needs a JDK > 6. Is that a problem?

### DIFF
--- a/project/SensibleSettings.scala
+++ b/project/SensibleSettings.scala
@@ -64,7 +64,8 @@ object Sensible {
       "org.scala-lang.modules" %% "scala-xml" % scalaModulesVersion,
       "org.scala-lang.modules" %% "scala-parser-combinators" % scalaModulesVersion,
       "org.scalamacros" %% "quasiquotes" % quasiquotesVersion,
-      "org.scalatest" %% "scalatest" % scalatestVersion
+      "org.scalatest" %% "scalatest" % scalatestVersion,
+      "com.typesafe" % "config" % configVersion
     ) ++ logback ++ guava ++ shapeless(scalaVersion.value)
   ) ++ inConfig(Test)(testSettings) ++ scalariformSettings
 
@@ -102,12 +103,12 @@ object Sensible {
   )
 
   val scalaModulesVersion = "1.0.4"
-  val akkaVersion = "2.3.14"
-  val streamsVersion = "1.0"
+  val akkaVersion = "2.4.4"
   val scalatestVersion = "2.2.6"
   val logbackVersion = "1.7.19"
   val quasiquotesVersion = "2.0.1"
   val guavaVersion = "19.0"
+  val configVersion = "1.3.0"
 
   val macroParadise = Seq(
     compilerPlugin("org.scalamacros" % "paradise" % "2.0.1" cross CrossVersion.full)
--- a/project/EnsimeBuild.scala
+++ b/project/EnsimeBuild.scala
@@ -167,7 +167,7 @@ object EnsimeBuild extends Build {
       )
 
   val luceneVersion = "4.7.2"
-  val streamsVersion = "1.0"
+  val streamsVersion = Sensible.akkaVersion
   lazy val server = Project("server", file("server")).dependsOn(
     core, swanky, jerky,
     s_express % "test->test",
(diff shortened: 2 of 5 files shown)


project/SensibleSettings.scala
```diff
@@ -13,7 +13,8 @@ object Sensible {
 
   lazy val settings = Seq(
     ivyLoggingLevel := UpdateLogging.Quiet,
-    conflictManager := ConflictManager.strict,
+    // FIXME: how to be less strict for typesafe config? slick and akka pull in different versions
+    //conflictManager := ConflictManager.strict,
```
jrudolph: How do you deal with conflicts like this when it's likely or known that the versions are actually compatible?

fommil: you mean incompatible? Well there is nothing you can do when they are incompatible, we have to find a combination of all deps that actually work well together.

fommil:
ah, fantastic! :smile: We also have https://github.com/ensime/ensime-server/pull/1416 which is doing the same thing :blush: @bartekkalinka how would you both like to proceed?

project/SensibleSettings.scala
fommil: if you did mean "compatible" then you use `dependencyOverrides` to specify the version you want.

jrudolph: No, they probably are compatible but `update` fails with this when the line is active:

```
[info] Updating {file:/home/johannes/git/opensource/ensime/}core...
[error] com.typesafe#config;1.2.1 (needed by [com.typesafe.slick#slick_2.11;3.1.1]) conflicts with com.typesafe#config;1.3.0 (needed by [com.typesafe.akka#akka-actor_2.11;2.4.4])
[trace] Stack trace suppressed: run last core/*:update for the full output.
[error] (core/*:update) org.apache.ivy.plugins.conflict.StrictConflictException: com.typesafe#config;1.2.1 (needed by [com.typesafe.slick#slick_2.11;3.1.1]) conflicts with com.typesafe#config;1.3.0 (needed by [com.typesafe.akka#akka-actor_2.11;2.4.4])
[error] Total time: 0 s, completed Apr 15, 2016 12:40:50 PM
```

jrudolph: When I put config in `dependencyOverrides` it seems to work.

bartekkalinka:
I thought in #1416 I'll do a temporary upgrade of akka-http/streams only to 2.0.4, and upgrade of whole akka to 2.4.x may be done later.

jrudolph:
I think you need to decide which akka version ensime can run on. For Scala 2.12, running on Akka 2.4.x may be required anyways, though.

project/SensibleSettings.scala
fommil: cool, yes that's intentional so we know when we have two things asking for different versions and then we can investigate further.

jrudolph:
@bartekkalinka in case you want to proceed with this, please feel free to incorporate some of my changes as well.

fommil:
is everything here still compatible with Java 6 or will we have to cut a Java 7+ branch just for these upgrades? In which case splitting into two parts may be necessary: everything that ENSIME can currently take (on Java 6) and everything that needs Java 7. I can create a branch if the latter is needed, but it sounds like it could be a real pain if we need to do that.

jrudolph:
Scala 2.12 needs Java 8, so what was the matrix of versions you wanted to support? I think it may turn out unrealistic to require cross-built versions of all dependencies that still run on Java 6 for the Scala 2.11 variant if they can rely on Java 8 safely for 2.12, no?

fommil:
@jrudolph ENSIME is still supporting Java 6 and Scala 2.10 / 2.11. But I'm planning on cutting a release, just from my desktop, for Java 8 / Scala 2.12. We can't merge any akka code into ensime-server `master` if it requires Java 7+.

jrudolph:
Ok, so it cannot work against one Akka version because 2.3.x AFAIK won't be published for 2.12 and 2.4.x needs Java 8. I guess then you could still try to switch versions by scalaVersion in your build and hope that no source changes are needed.

fommil:
uugh, that's a mess. Well since I'm publishing locally, I can swap the version myself so long as they are nice enough to make it source compatible.

fommil:
will I close this safe in the knowledge that all the good stuff in here will be ported to #1416 ?

fommil:
since the other PR only brings us part way to here, I'm re-opening this PR but if @bartekkalinka wants to take this and start a fresh PR we can close it again.

fommil:
JDK8 is a blocker.

Issue #1418, opened 2016-04-15
use sftp drone plugin


fommil:
I've asked @larsrh if he could add our drone keys to the typelevel upload space. Once (if) that's in place, we can use this awesome plugin to simply the artefact uploading

  https://github.com/drone-plugins/drone-sftp/issues/8

which I think should be easy to install in drone.

fommil:
possibly blocked on https://github.com/drone-plugins/drone-sftp/issues/9 we'll see

PR #1420, opened 2016-04-15
CI improvements

fommil:


fommil:
closes #1421 I'm such an idiot

fommil:
might close #1418

fommil:
CI is actually OK I cancelled it to free up resource after squashing a green build.

Issue #1412, opened 2016-04-11
Remove netbeans


fommil:
I've just realised that the netbeans jars we use are GPLv2 (not GPLv3) and the alternative CDDL licence is not GPL compatible so we are probably in a gray area.

fommil:
actually, I wonder if we are ok with using them at all... I'm not sure we count as an "independent module". @aemoncannon how much value are we getting out of the netbeans libraries anyway? I've been pretty disappointed by the lack of engagement from that community to be honest. How much effort would it be to implement what we're using?

BTW, GPL2 is incompatible with Apache 2.0 so we'd be pretty scuppered even if we weren't GPLv3.

fommil:
("downloading the netbeans universe" is becoming a bit of a common complaint about ENSIME these days, I've noticed... and I do tend to agree)

fommil:
FYI, we can't swap to using Eclipse either, because their licence is also incompatible with us.

fommil:
Damn, it really is incompatible. Totally my bad, I should have seen that.

fommil:
I had a quick look. We barely use netbeans at all, we can totally rip it out, much to the pleasure of all our users.

aemoncannon:
re: how much value are we getting out of the netbeans libraries anyway?
Good question. It actually boils down to a couple helpers in TreeUtilities.java:
pathFor, scopeFor, and maybe one other thing. If we clean-room reimplement those utilities then I _think_ we could drop all of netbeans (hallelujah).

fommil:
this is the last time I ever used the phrase "well it's better than Eclipse" to justify *any* technical decision...

fommil:
@mwielocha I've noticed you're keen on Java support. It's this ticket something we could tempt you into picking up?

mwielocha:
Sure, on it.

lukasz-golebiewski:
I might have a look into this as well. Can this task be somehow split in two parts?

fommil:
@mwielocha @lukasz-golebiewski probably bests if the both of you find some time to chat on the `ensime-server` gitter chat room (or here) and come up with a plan :smile: One possibility is that one person (or both with a screen sharing session) creates a work-in-progress branch that removes NetBeans and creates a stub class (which can be written in Scala) with the features that we need... then you could separate the workload and independently write tests / implementations for the methods.

fommil:
or if we expand the scope of work to be done, how about implementing something like Scalariform for Java code using the Java AST?

[...]

Issue #1405, opened 2016-04-03
javacOptions for java analyser


fommil:
We currently ignore the user's `javacOptions` but we should really be passing them to the java analyser.

Will require an extra field added to the `EnsimeConfig` and support in ensime-sbt to add the flag.

mwielocha:
Hi, requested pull for sbt part of this, also I have the server part but don't really have an idea how to test it. Plus while experimenting with common javac options like -source 1.6, I'm getting NPE from com.sun.tools.javac.util.ClientCodeException. Might need some help on this.

fommil:
Cool! I'll have a detailed look in the next day or too.

mwielocha:
All you need to do to replicate the NPE is to add -source 1.6 in JavaCompiler line 65: 
```scala
compiler.getTask(null, fileManager, listener, List(
      "-cp", cp, "-Xlint:" + lint, "-proc:none", "-source", "1.6"
    ).asJava, null, files).asInstanceOf[JavacTask]
```
And run JavaCompilerSpec.

fommil:
This might have something to do with us using a version of netbeans libraries that targets the Java 7 api, but still runs on Java 6. Thanks for the repro. Please ping me if I forget to look at this, I'm under a lot of stuff right now.

mwielocha:
Seems that removing netbeans didn't help on this one. I did some digging and I think this might have something to do with [https://blogs.oracle.com/darcy/entry/bootclasspath_older_source](url), anyway adding `-bootclasspath` as na options partially fixes the problem (still throws an exception on `JavaCompilerSpec` - unfortunately). Don't know where to go from here.

jozi-k:
I think this is already solved.

fommil:
No it's not implemented yet.

PR #1424, opened 2016-04-18
Remove netbeans dependencies

mwielocha:
Request for #1412 
The basic idea is that the `scopeFor` and `pathFor` of `TreeUtilities` are now objects with corresponding apply methods. That way I could easily introduce a overloaded `TreeScanner` in the object's companion class for the `PathFor` functionality. That nasty hack with throwing an error to break out of the scanner is still there since I kinda assumed that this is the best way to go.
I've also removed the code that was dependant on source code tokenisation (`TokenHierarchy` and `TokenSequence` related) - the fact that `JavaCompilerSpec` is passing smoothly tells me that what we are left with is quite sufficient.


core/src/main/scala/org/ensime/core/javac/PathFor.scala
```diff
@@ -0,0 +1,55 @@
...
+import com.sun.source.util.SourcePositions
+import com.sun.source.util.TreePath
+import com.sun.source.util.TreePathScanner
+import scala.util.Try
+import scala.util.Failure
+import scala.util.Success
+
+import java.lang.Void
+
+case class Hit(path: TreePath) extends Error
```
fommil: is Error maybe a bit harsh?

core/src/main/scala/org/ensime/core/javac/PathFor.scala
```diff
@@ -0,0 +1,55 @@
...
+    startPosition <= position && endPosition >= position
+  }
+
+  override def scan(tree: Tree, p: Void): Void = {
+
+    Option(tree) match {
+
+      case Some(t) if isAtPosition(t) && t.getKind == Tree.Kind.ERRONEOUS =>
+        t.accept(this, p)
+        throw Hit(getCurrentPath)
```
fommil: this is where I go back and read your cover notes... :smile:

project/EnsimeBuild.scala
```diff
@@ -234,7 +231,6 @@ object EnsimeBuild extends Build {
       aggregate in assembly := false,
       assemblyMergeStrategy in assembly := {
         case PathList("META-INF", "namedservices", xs @ _*) => MergeStrategy.filterDistinctLines
```
fommil: a bunch of this other stuff is probably netbeans to. If you remove it, and `sbt ensime/assembly` still works, then it is ok to remove it.

fommil:
wow, I'm seriously impressed!

Since this could have a big impact on people using Java support (i.e. @aemoncannon ) let's leave this open for a week or so while they test it out.

I'm actually doing a lot of Java coding this week, so I'll check it out straight away!

mwielocha:
Plus the unused imports, I think ensime blew up while trying to optimise them and I haven't noticed - the irony.

fommil:
btw, have you got ENSIME logo stickers? If not, please email me your postal address and I'll make sure you get a few.

mwielocha:
There, now its an `Exception`, doesn't really feel less harsh, though ;P

fommil:
:laughing: I love the way we're pushing the boundaries of the Typelevel ethos: "pure, typeful, functional programming"

FWIW, "it works for me"! Just did a local assembly build and I'm using it to hack on another great example of monadic, referentially transparent, immutable programming: https://github.com/fommil/class-monkey :mask:

fommil:
well this has been working great for me all night, no problems at all. The ensime `.jar` file is also 15MB smaller. We'll merge in a week or when we hear back from Aemon.

core/src/main/scala/org/ensime/core/javac/Helpers.scala
```diff
@@ -47,32 +47,31 @@ trait Helpers extends UnsafeHelpers with SLF4JLogging {
...
-      kind == ElementKind.METHOD || kind == ElementKind.FIELD) {
-      Option(el.getEnclosingElement).flatMap(fqn(info, _)).map(_.copy(fieldOrMethod = Some(el.toString)))
-    } else {
-      parseFqnAsClass(el.toString)
+  private val isSimpleKind = Set(ElementKind.LOCAL_VARIABLE, ElementKind.PARAMETER)
+  private val isComplexKind = Set(ElementKind.CONSTRUCTOR, ElementKind.ENUM_CONSTANT, ElementKind.METHOD, ElementKind.FIELD)
+
+  def fqn(c: Compilation, el: Element): Option[JavaFqn] = {
+    el.getKind match {
+      case k if isSimpleKind(k) => Some(JavaFqn(None, None, Some(el.getSimpleName.toString)))
```
aemoncannon: nitpick: this looks nicer but is likely much slower

project/EnsimeBuild.scala
```diff
@@ -141,9 +141,6 @@ object EnsimeBuild extends Build {
         "com.h2database" % "h2" % "1.4.191",
         "com.typesafe.slick" %% "slick" % "3.1.1",
         "com.zaxxer" % "HikariCP-java6" % "2.3.13",
-        // Netbeans 7.4+ needs Java 7 (7.3 only needs it at runtime)
```
aemoncannon: Yay!

core/src/main/scala/org/ensime/core/javac/ScopeFor.scala
```diff
@@ -0,0 +1,45 @@
...
+          case Kind.METHOD => path.getLeaf.asInstanceOf[MethodTree].getParameters.toList
+          case otherwise => Nil
+        }
+
+        val isAtPosition: StatementTree => Boolean = {
+          sourcePositions.getStartPosition(root, _) <= position
+        }
+
+        val treePathOption = for {
+          tree <- statements.reverse.find(isAtPosition)
```
aemoncannon: Is there a nice way to do this without constructing the reversed list?

aemoncannon:
Looks great! I want to give it a try on my work project.

core/src/main/scala/org/ensime/core/javac/Helpers.scala
mwielocha: I have a feeling that the difference might not be measurable, but I might be wrong...

fommil: it might be more idiomatic scala to turn `isSimpleKind` into an `unapply`

aemoncannon: feel free to ignore my comment. i write Android java for a living so things like this jump out at me.

core/src/main/scala/org/ensime/core/javac/Helpers.scala
```diff
@@ -47,13 +47,24 @@ trait Helpers extends UnsafeHelpers with SLF4JLogging {
     Some(JavaFqn(front.mkString("."), back.mkString("."), None))
   }
 
-  private val isSimpleKind = Set(ElementKind.LOCAL_VARIABLE, ElementKind.PARAMETER)
-  private val isComplexKind = Set(ElementKind.CONSTRUCTOR, ElementKind.ENUM_CONSTANT, ElementKind.METHOD, ElementKind.FIELD)
+  private class isOneOf(options: ElementKind*) {
```
fommil: pattern matchers are typically uppercased.

That said, since this is basically a switch statement, we could just do

`case LOCAL_VARIABLE | Parameter | ... => `

and avoid any intermediates.

mwielocha:
And to think that I don't even have a strong opinion on that pattern match...

fommil:
@aemoncannon did it work ok?

fommil:
We should probably start creating a bunch of low hanging fruit feature requests for Java support.

aemoncannon:
lgtm. i haven't had a chance to use it, but this is too important to block. thanks!

fommil:
well if it breaks anything for you, you can always roll back and publish your own assembly jar :smile:

fommil:
thank you for this @mwielocha ! If you're keen on improving the Java support, feel free to create a bunch of small tickets for each missing feature (maybe list all the protocol messages that are supported by scala but not java?).

Two big ones that I always think about when doing Java:

1. add type to variable. We have this in scala also, but it's not syntactically necessary. Here it would be fixing an error.
2. formatting. There is no such thing as scalariform for Java. Netbeans and Eclipse have something, but then we're back to depending on 30MB of jars again. This would be an incredibly useful standalone library, as it would be brilliant to have an sbt plugin like scalariform for java code. Initially just supporting one style (configuration can come later) would be great.

fommil:
and don't forget to let me know where to send the logo stickers, if you want some for your laptop. We should probably get better swag, this kind of contribution is really what ENSIME is all about.

mwielocha:
I'll have a look at those, although number 2 seems like quite a long term task, especially when we want to have something completely independent. 
Just to say: I don't really have a strong felling about Java, #1405 was just the first low hanging fruit on the list 😅

fommil:
haha, ok, well there is plenty of other low hanging fruit, so just pick whatever you feel like. Also the "Needs Reproduction" tickets are well typically self contained.

lukasz-golebiewski:
This is really cool, thanks @mwielocha!

Issue #1198, opened 2015-12-31
Java completions shouldn't contain private functions


tanzoniteblack:
When using ensime to work on Java files (which I know is still in alpha), the completions list contain private functions for other objects.

This is not true when working with autocompletions for Java objects from within a scala file.

fommil:
yeah, we should probably filter by ability to see the symbol. Does our scala mode handle this correctly?

tanzoniteblack:
The scala mode handles it correctly.

aemoncannon:
Good catch! Should be a pretty easy change

aemoncannon:
@tanzoniteblack, want to take a crack at this?

tanzoniteblack:
Sure. I'll take a look when I get a chance. Probably sometime tomorrow.

tanzoniteblack:
@aemoncannon I have some tests sitting on a branch (https://github.com/ensime/ensime-server/compare/master...tanzoniteblack:java-private-completions?expand=1), but it's not immediately obvious to me how to actually filter out the privates.

PR #1426, opened 2016-04-19
filter out not accessible members

mwielocha:
Request for #1198 
I have a version of this for both master and #1412 branch.
Will merge when `no netbeans` gets approved.

### DIFF
--- a/core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
+++ b/core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
@@ -18,6 +18,7 @@ import com.sun.source.tree.Scope
 import scala.collection.mutable.ArrayBuffer;
 import scala.concurrent.Await
 import scala.concurrent.duration.Duration
+import scala.collection.breakOut
 
 trait JavaCompletion extends Helpers with SLF4JLogging {
 
@@ -209,7 +210,7 @@ trait JavaCompletion extends Helpers with SLF4JLogging {
     caseSense: Boolean
   ): List[CompletionInfo] = {
 
-    typeElement(info, target).map {
+    typeElement(info, target).toList.flatMap {
 
       case tel: TypeElement =>
 
@@ -221,13 +222,12 @@ trait JavaCompletion extends Helpers with SLF4JLogging {
 
         info.getElements.getAllMembers(tel).filter(isAccessible).flatMap { el =>
           filterElement(info, el, prefix, caseSense, importing, false)
-        }
+        }(breakOut)
 
       case e =>
         log.warn("Unrecognized type element " + e)
         List.empty
-
-    }.map(_.toList).getOrElse(List.empty)
+    }
   }
 
   private def methodInfo(e: ExecutableElement, relavence: Int): CompletionInfo = {
--- a/core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
+++ b/core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
@@ -214,9 +214,10 @@ trait JavaCompletion extends Helpers with SLF4JLogging {
 
       case tel: TypeElement =>
 
+        val path = info.getTrees.getPath(info.getCompilationUnit, target)
+        val scope = info.getTrees.getScope(path)
+         
         def isAccessible(el: Element): Boolean = {
-          val path = info.getTrees.getPath(info.getCompilationUnit, target)
-          val scope = info.getTrees.getScope(path)
           info.getTrees.isAccessible(scope, el, info.getTypes.getDeclaredType(tel))
         }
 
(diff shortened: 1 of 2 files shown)


fommil:
this is great stuff, I really noticed the private stuff when I was doing my recent Java hacking. There is a lot of stuff to filter out, it would appear :smile: (and almost everything I wanted access to turned out to be `private`... bloody stdlib...)

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
```diff
@@ -208,21 +208,26 @@ trait JavaCompletion extends Helpers with SLF4JLogging {
...
+
+        info.getElements.getAllMembers(tel).filter(isAccessible).flatMap { el =>
+          filterElement(info, el, prefix, caseSense, importing, false)
         }
-      }
-    }.getOrElse(List())
-    candidates.toList
+
+      case e =>
+        log.warn("Unrecognized type element " + e)
```
fommil: you might want to create a ticket to capture these, with detailed instructions of what you want people to do, and then  direct people to the ticket? You never know, somebody might add support for more trees.

mwielocha:
Yup, btw. the filter out `<init>` is for the `no netbeans` stuff, I missed that because of all the noise from private members.

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
```diff
@@ -208,21 +208,26 @@ trait JavaCompletion extends Helpers with SLF4JLogging {
...
+        List.empty
+
+    }.map(_.toList).getOrElse(List.empty)
```
fommil: do you know about `breakOut`? Highly recommended watching: https://skillsmatter.com/skillscasts/6939-optimising-scala-for-fun-and-profit (you'll never call `.toList`, `.toMap` or `.toSet` again...)

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
```diff
@@ -208,21 +208,26 @@ trait JavaCompletion extends Helpers with SLF4JLogging {
...
   ): List[CompletionInfo] = {
-    val candidates = typeElement(info, target).map { el =>
-      el match {
-        case tel: TypeElement => {
-          val elements: Elements = info.getElements()
-          elements.getAllMembers(tel).flatMap { e =>
-            filterElement(info, e, prefix, caseSense, importing, false)
-          }
+
+    typeElement(info, target).map {
```
fommil: is this maybe better as a `flatMap`? I realise this is just code you're changing but no reason not to clean it up at the same time.

mwielocha: I am now a bit more cautious when cleaning up, after the pattern match from  #1412 ;P

fommil: You'll soon learn to put one contentious style change in every PR, to distract us from the really hard bits :smile:

mwielocha:
So what is going on here? Isn't this auto-scalariformatted?

fommil:
scalariform is a bit weird, it formats source code as a pre compile phase, so you probably made changes without compiling.

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
```diff
@@ -209,7 +210,7 @@ trait JavaCompletion extends Helpers with SLF4JLogging {
     caseSense: Boolean
   ): List[CompletionInfo] = {
 
-    typeElement(info, target).map {
+    typeElement(info, target).toList.flatMap {
```
fommil: there shouldn't be any need for the `toList` here

mwielocha: I must admit, there is something here that I am missing: so you're saying that this: `Option(1).flatMap(x => List(x))` should compile, right?

fommil: yes, I think so (unless this is a gotcha)

mwielocha: Not a gotcha, I was honestly wondering how can this be done nicely, but unfortunately this only works one way: `List(1).flatMap(x => Option(x))`, either way I'm pretty sure that `Option.toList` is super-cheap, so shouldn't really be a problem here.

mwielocha:
Not sure if this: `java.lang.AssertionError: assertion failed: timeout (10 minutes) during expectMsgClass waiting for interface` is in any way connected with this pr...

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
fommil: oh sure, if it's on an Option then it's fine. I couldn't see the starting type.

fommil:
if it's in the `DebugTest` then it's just an annoying transient we've not tracked down yet.

fommil:
well I think this is a lot less contentious than the Netbeans removal, so I'll just accept this as is. Thank you!

PR #1427, opened 2016-04-20
Set akka.http.server.idle-timeout to infinite

2chilled:
To prevent the connection from being closed even if the ensime client does not send any messages to it's server for some time.

### DIFF
--- a/core/src/main/resources/application.conf
+++ b/core/src/main/resources/application.conf
@@ -35,4 +35,10 @@ akka {
       task-queue-size = -1
     }
   }
+  
+  http {
+    server {
+      idle-timeout = infinite
+    }
+  }
 }


fommil:
I'm not sure setting this to infinite is a good idea, it could have unintended consequences. @rorygraves thoughts?

rorygraves:
I think infinite is fine.  Resources cleared down on client disconnect - we are not a high volume website or something.  Without an active keepalive ping from clients (which is my prefered long term option), I'm not sure what we would set it to otherwise.

rorygraves:
@fommil  The appveyor error is another 10 minute timeout - thought you had tweaked those:

```
 - should trigger/continue *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (10 minutes) during expectMsgClass waiting for interface org.ensime.api.EnsimeServerMessage
[info]   at scala.Predef$.assert(Predef.scala:170)
[info]   at akka.testkit.TestKitBase$class.expectMsgClass_internal(TestKit.scala:423
```
Are you happy for me to merge otherwise?

fommil:
I fixed a timeout where we expected NO message for 10 mins. This is a real messed message, but if it's DebugTest I'm happy to merge.

rorygraves:
It is DebugTest - merging.

fommil:
you forget to choose squash :stuck_out_tongue:  (so we have a merge commit)

fommil:
thanks @2chilled this will hopefully make things more stable for Vim and Atom

PR #1433, opened 2016-04-29
Switch to Scala debugger API

### LINKED ISSUE #1223: Migrate to use Scala debugger api
Currently, the Ensime server uses the in-house debugging logic mixed with Akka message passing to handle debugging. Now that the [Scala debugger api](https://github.com/ensime/scala-debugger) is part of Ensime, we should look into migrating the server to using the api.

At this point, I don't know what the feature parity is between the debugger api and server; so, the debugger api might need a couple of updates along the way. Not sure how to break this one up, @fommil @rorygraves. Is there any way you can think of to tackle this in parts? Would make it more approachable for me and anyone else that wants to work on it.
###### Ensime Server Verifications
- [x] Achieve working integration tests based on old tests
- [ ] Add unit tests for all new classes/methods (*sob*)
- [x] Add new integration tests to cover other message types
- [ ] Manually test Ensime with Emacs/vim

###### Ensime Server Alterations
- [x] Create actor file that is purely geared towards processing input from the user (refactor existing actor) (DebugActor.scala)
- [x] Split off logic to create, use, and destroy virtual machine (VirtualMachineManager.scala)
- [x] Refactor logic transforming JDI information to Ensime-specific messages into separate class (StructureConverter.scala)
- [x] Clean up source manager, which currently uses raw source names rather than source paths
    - Still needs to allow looking up by source name rather than path to stay compatible
- [ ] Ensure proper refresh of source map (#1375)
- [ ] Fix duplicate breakpoints and traversal of sourcemap (#1330)
- [x] Start tests by starting a separate process and using attaching debugger instead of using launching debugger

###### Scala Debugger Improvements
- [ ] Even more issues, but I'm getting tired of listing them here...
- [x] https://github.com/ensime/scala-debugger/issues/235
- [x] https://github.com/ensime/scala-debugger/issues/233
- [x] https://github.com/ensime/scala-debugger/issues/231
- [x] https://github.com/ensime/scala-debugger/issues/229
- [x] https://github.com/ensime/scala-debugger/issues/219
- [x] https://github.com/ensime/scala-debugger/issues/218
- [x] https://github.com/ensime/scala-debugger/issues/217
- [x] https://github.com/ensime/scala-debugger/issues/213
- [ ] https://github.com/ensime/scala-debugger/issues/212
- [x] https://github.com/ensime/scala-debugger/issues/211
- [x] https://github.com/ensime/scala-debugger/issues/209
- [x] https://github.com/ensime/scala-debugger/issues/208
- [x] https://github.com/ensime/scala-debugger/issues/203
- [x] https://github.com/ensime/scala-debugger/issues/200
- [x] https://github.com/ensime/scala-debugger/issues/198
- [x] https://github.com/ensime/scala-debugger/issues/183
- [x] https://github.com/ensime/scala-debugger/issues/182
- [x] https://github.com/ensime/scala-debugger/issues/181
- [x] https://github.com/ensime/scala-debugger/issues/180
- [x] https://github.com/ensime/scala-debugger/issues/179
- [x] https://github.com/ensime/scala-debugger/issues/178
- [x] https://github.com/ensime/scala-debugger/issues/177
- [x] https://github.com/ensime/scala-debugger/issues/176
- [x] https://github.com/ensime/scala-debugger/issues/175
- [x] https://github.com/ensime/scala-debugger/issues/174
- [x] https://github.com/ensime/scala-debugger/issues/173
- [x] https://github.com/ensime/scala-debugger/issues/172
- [x] https://github.com/ensime/scala-debugger/issues/171
- [x] https://github.com/ensime/scala-debugger/issues/170
- [x] https://github.com/ensime/scala-debugger/issues/169
- [x] https://github.com/ensime/scala-debugger/issues/168
- [x] https://github.com/ensime/scala-debugger/issues/163
- [x] https://github.com/ensime/scala-debugger/issues/158
- [x] https://github.com/ensime/scala-debugger/issues/157
- [x] https://github.com/ensime/scala-debugger/issues/156
- [x] https://github.com/ensime/scala-debugger/issues/155
- [x] https://github.com/ensime/scala-debugger/issues/152
- [x] https://github.com/ensime/scala-debugger/issues/151
- [x] https://github.com/ensime/scala-debugger/issues/150
- [x] https://github.com/ensime/scala-debugger/issues/142
- [x] https://github.com/ensime/scala-debugger/issues/122

chipsenkbeil:
Resolves #1223. Still needs some refactoring of the `DebuggerActor` class since I don't want all of the logic directly inside an Akka actor. Some of the event handlers can be made more concise as well. Also want to add more integration tests and figure out why the existing integration tests are stalling occasionally.

- [x] Add additional integration tests for other message types
- [ ] Revamp integration test platform (for debugger) using Scala debugger API utilities such as `JDITools.spawn(...)` (see #1456)
- [ ] Refactor `DebuggerActor` command actions to separate classes that can be tested in isolation (see #1451)
- [x] Move from snapshot 1.1.0 version of Scala debugger to an M1 release (Ensime doesn't use snapshots)


chipsenkbeil:
Oh, interesting. TODOs aren't allowed. I'll have to switch them to... something else, I guess.

chipsenkbeil:
FYI, I'll be pushing quite a few junk commits to this WIP. Obviously, once it's ready to go, I'll rebase or whoever merges can use Github's new "squash and merge."

chipsenkbeil:
I wonder why I haven't seen any of these "warning to error" issues before. I've been building and testing locally after rebasing on the latest. Is there a flag that is being set on the build systems?

chipsenkbeil:
Mm, now it's failing for code formatting failing Ensime standards. Ha ha, I tend to have weird styling. I thought your formatter would have fixed that since it kept formatting my changes whenever I ran the tests, but maybe it doesn't cover everything.

chipsenkbeil:
Hm, breakpoint test failing. I'll have to double-check that rebasing didn't change something as it has not failed in that way locally (always fun).

fommil:
Hey that's cheating! :-P the point of banning TODO is to force use of the issue tracker so that open problems can be picked up by anybody, and ensure a proper description of the work is written. TODO style comments rot in the code.

This is great :-)

Because everything is async it means the assertions can wait for up to 10 minutes. So you're probably seeing real failures. You need to look in the integration test logs for further evidence of what went wrong.

fommil:
Hmm, I thought i turned off the LGTM plugin.

chipsenkbeil:
@fommil yeah, I'm not planning to leave those CHIP comments in the code, but I haven't decided what warrants an issue versus what was just musings of my own. :) Wanted to get something passing through all of the style checkers.

chipsenkbeil:
As of fc6eb35, changed snapshot of debugger api to suspend all threads for class prepare events. I think this might be causing a specific race condition when it was set to just the event thread, but I'm not sure.

chipsenkbeil:
Looks like removing the resume-on-attach fixed the issue for Mac/Linux. Windows is failing every time, but I don't know if the resume was related to that or not.

fommil:
just let me know when you're ready for me to take a final look by removing the `[WIP]` name from the PR

chipsenkbeil:
Will do. Still need to clean up the DebugActor. Was sidelined by the test inconsistency, but now can move forward with the refactoring. Shouldn't be too much longer, although I have a track record of being wrong. ;)

core/src/main/resources/logback.xml
```diff
@@ -14,6 +14,9 @@
     <appender-ref ref="STDOUT" />
   </root>
   <logger name="org.ensime" level="DEBUG" />
+  <logger name="org.ensime.core.debug" level="DEBUG" />
+  <logger name="org.scaladebugger" level="DEBUG" />
+  <logger name="org.scaladebugger.api.lowlevel" level="TRACE" />
```
fommil: eek, definitely to be backed out for local dev use only.

rorygraves: Doing a review of a WIP branch is a bit mean ;)

core/src/main/scala/org/ensime/core/debug/NoActiveVirtualMachineException.scala
```diff
@@ -0,0 +1,7 @@
+package org.ensime.core.debug
+
+/**
+ * Represents the scenario when a virtual machine is requested but there is
+ * no active virtual machine available.
+ */
+class NoActiveVirtualMachineException extends Exception
```
fommil: I'm not particularly keen on one line source files. Is there somewhere standard you could gather all your domain objects?

core/src/main/scala/org/ensime/core/debug/VirtualMachineManager.scala
```diff
@@ -0,0 +1,194 @@
...
+   *
+   * @param action The action to evaluate on top of the JVM
+   * @tparam T The expected return value from the action
+   * @return Some containing the result if successful, otherwise None
+   */
+  def withVM[T](action: (ScalaVirtualMachine => T)): Try[T] = synchronized {
+    if (!hasActiveVM) {
+      val error = new NoActiveVirtualMachineException
+      log.error("No VM active for debugging!", error)
+      return Failure(error)
```
fommil: `return` is not particularly idiomatic, can you perhaps reshape the function to avoid needing it?

fommil: if you're wanting a `Try` interface here, I don't think it makes sense to define a new domain exception for it. Just use an existing exception type, such as `IllegalStateException` with an appropriate message.

fommil:
only a few minor comments from me so far. It is good that you've been able to keep it localised to the debug code. I find that my hand wanders on big refactors and I end up touching everything.

core/src/main/resources/logback.xml
chipsenkbeil: 😆 Yeah, this will be backed out.

core/src/main/scala/org/ensime/core/debug/NoActiveVirtualMachineException.scala
chipsenkbeil: I'm sure I can, just need to look around.

core/src/main/scala/org/ensime/core/debug/VirtualMachineManager.scala
chipsenkbeil: Sure. Saves me needing to look around for a place to put the exception. Granted, I do prefer having more explicit exceptions.

fommil:
I'll have a more detailed look when the WIP gets removed :stuck_out_tongue:

chipsenkbeil:
Windows tests are passing now.

fommil:
wow! What did you do? What was the core problem?

fommil:
how many times did you sees success on windows though, it was more that the tests were flaky, failing maybe 20% of the time.

chipsenkbeil:
I had one place in the source map where I hard-coded a `/` instead of using `java.io.File.separator`.

chipsenkbeil:
As for times of success, I've seen it on one full run so far. I'll be refactoring more code around, so Windows will get a couple of additional runs. Given that the debugger API passes all of my integration tests (over 150 of them) except for the VMDeath event (over ten time now), I'm pretty confident that the core api is stable on Windows. We'll hopefully catch any ensime-server problems in the next couple of iterations.

chipsenkbeil:
Disregard the failures for d0c46db. Committed non-compiling code by accident. So far, 2/2 on all windows tests.

chipsenkbeil:
3045444 seems to hang near the end of all tests. All of the debugger tests pass, so not sure what's up with it that time.

chipsenkbeil:
Okay, total of 8/8 times where the windows tests pass, but this is the second time I've seen appveyor freeze and not finish, resulting in a failure. You can see all of the debug tests pass and additional tests after them pass as well, though. Have either of you seen this happen before? Only thought relating this problem to debug tests would be child processes from debug tests are not getting shut down properly and are causing appveyor to have problems.

fommil:
We could be breaching a memory limit.

chipsenkbeil:
10/10 Windows tests passing. Lowered the memory footprint of the debug processes, but still not sold that the Windows lockup is entirely the debug tests' fault.

Anyway, removing the WIP and created issues for refactoring-related work. Have at it, @fommil and @rorygraves.

core/src/it/scala/org/ensime/intg/DebugTest.scala
```diff
@@ -274,47 +488,50 @@ trait DebugTestUtils {
     expectMsg(TrueResponse)
 
     val vm = VMStarter(config, className)
```
rorygraves: Minor point - and it already existed - I'm becoming a big fan of simple case classes for use cases such as this so we have sensible names on _4 (a future), _3, etc.

rorygraves:
This LGTM !

fommil:
:shipit:

fommil:
:tada:

core/src/it/scala/org/ensime/intg/DebugTest.scala
chipsenkbeil: Yeah, I am as well. To be honest, I left it with the intention of removing that code since there is similar deployment code used in the debugger API under the `JDITools` object, but it needs to have a way for you to tie in to the standard output/error and redirect. The issue is #1456 with regard to that refactoring.

fommil: you could pattern match on the left hand side. It is weird that something that looks like a constructor is returning a tuple.

chipsenkbeil: Ha ha. I thought it was weird when I saw it, too.

Issue #1341, opened 2016-03-12
backpressure in the file indexing


fommil:
Currently, in the `SearchService`, we aggressively index the user's classfiles. However, this leads to memory usage explosions when the database write is too slow.

Make it use backpressure so we limit how many classes are being indexed at any one time. ensime server heap usage could be dramatically reduced.

fommil:
#1351 would probably make this trivial

I say Low Hanging Fruit, but only for somebody with experience of doing this sort of thing.

ChrisCoffey:
Is this any more complicated than using a bounded mailbox for the `IndexingQueueActor`?

fommil:
Yup. We're not using actors for this code. But I'm OK with blocking a thread to achieve backpressure, especially since I want to rewrite all this code to use something more like clojure SCP channels (and replace akka everywhere else with SCP)

fommil:
And also bounded queues don't achieve barkpressure, don't they just drop messages?

fommil:
Are you interested in looking at this? It would really help keep memory under control. I don't think it's too hard to get something working.

ChrisCoffey:
I believe you can make them block if you set the mailbox timeout to infinite. That being said, if the plan is to remove Akka anyways we're better off replacing the `IndexingQueueActor` entirely. 

From my reading of the code, I think we can get away with using a `java.util.concurrent.LinkedBlockingQueue` and drop the actor. I did something like this a few months ago at work.

ChrisCoffey:
Sure, I'll take a stab at it this weekend.

fommil:
That actor isnt the bit that needs the backpressure (well, it does, but it's not the priority), I don't think. It's the indexing during refresh.

fommil:
I'm going to be out all day on Saturday but I'll be hacking on ensime all day sunday/monday.

ChrisCoffey:
Cool, I'll be online then too, albeit EST.

ChrisCoffey:
on #1436 @fommil said:

> @ChrisCoffey before I start teaching granny to suck eggs, how much of the reactive manifesto and sequential concurrent programming are you familiar with?

[...]

Issue #1351, opened 2016-03-13
use fs2 to rewrite SearchService


fommil:
**UPDATE**: (the original ticket was to replace all of akka with something else, but reducing the scope to use a non-akka solution for the indexer seems a much more achieveable goal)

Depending on akka brings in a huge dependency tree and restricts our ability to work on bleeding edge scala releases.

Also, I've become increasingly unhappy with akka as a framework and we're in Typelevel so we should be using more a functional paradigm, e.g. when cats / fs2 becomes viable. @non may have thoughts on this.

This ticket is specifically about replacing our akka actors with a different framework / paradigm. Minimalism and backpressure (especially for the horrid `SearchService`) are important, but we also have to bare in mind that we are fundamentally wrapping a black box API around the compiler.

I'm pretty sure `core.async` in clojure could do what we want.

fommil:
https://github.com/pnf/scala-csp

dwijnand:
Perhaps Monix has the primitives that ensime-server requires?

@alexandru do you think that ensime-server could switch to use Monix instead of Akka?

alexandru:
Hello, just noticed this ticket :-)

A Monix `Observable` might be the right abstraction, having these characteristics:

Back-pressure and everything associated (e.g. overflow policies in case of buffering) is for free, versus Akka actors for which you have to implement your own back-pressure protocol (I can also give hints for Akka btw).

Monix works best if it models a unidirectional flow of messages. For a dialog (e.g. ping-pong), you can think of a client-server communication through WebSocket if that helps, where both can listen to receiving messages and send replies. In such a case, actors are easier to use.

I mean, you can model bidirectional communications with Monix, but in order to replace a bidirectional actor you'll work with an input channel (where you push messages) and an output channel (the Observable, on which you can subscribe for listening) and it gets awkward, because having a dialog means keeping shared state between these two channels. You've got basically the same problem with say, Iteratees in Play, where for websocket you receive an input/output pair and you can compare it with usage of actors, also in Play, which is actually much simpler.

That said, if you don't have dialogs, or lets say difficult dialogs, then Monix's Observable certainly leads to a cleaner model, as it's more decoupled and it gives you stuff for free. You guys think of Clojure's core.async and Monix has many of the same advantages.

Currently I do not have time to look at Ensime's implementation. I'm very busy with releasing a stable `2.0` release for Monix. But afterwards I could take a look if you want. Not sure what the time table is like. I hope you're not in a hurry :-P

[...]

PR #1436, opened 2016-05-02
WIP reduce memory pressure during SearchService.refresh

ChrisCoffey:
This is still under construction, but there are a couple of approaches I've considered for #1341

- Synchronize the refresh flow as much as possible. Certainly introduces "backpressure", but it also feels like we're loosing out on some parallelism here
- Use a bounded queue & `CallerRunsPolicy` once the queue is full. With a queue of size one, this means we'll begin using the calling thread for overflow work, which in turn slows down the caller, applying some back pressure
- Introduce a custom `RejectedExecutionHandler` policy that actually blocks the caller rather than delegate work to it. This seems like overkill for what we need.

I toyed around with the 3rd option, but ended up settling on the `CallerRunsPolicy` approach. I want to make sure this passes tests up here, and in the meantime I'll remove the old `val handler`.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -246,7 +246,7 @@ class SearchService(
     batchSize: Int = 1000
   ): Future[Int] = {
     val removing = files.grouped(batchSize).map(delete)
-    Future.sequence(removing).map(_.sum)(executor = ex)
+    Future.sequence(removing).map(_.sum)
   }
 
   // returns number of rows removed
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -37,7 +37,7 @@ class SearchService(
     with SLF4JLogging {
 
   //Create a custom execution context that blocks the calling thread if no worker is available
-  private implicit val ex = BoundedExecutor.callerBlockingExecutor(4)
+  private val ex = BoundedExecutor.callerBlockingExecutor(4)
   private[indexer] def isUserFile(file: FileName): Boolean = {
     (config.allTargets map (vfs.vfile)) exists (file isAncestor _.getName)
   }
@@ -156,7 +156,7 @@ class SearchService(
   def refreshResolver(): Unit = resolver.update()
 
   def persist(check: FileCheck, symbols: List[FqnSymbol], commitIndex: Boolean, boost: Boolean): Future[Option[Int]] = {
-    val iwork = Future { blocking { index.persist(check, symbols, commitIndex, boost) } }(ex)
+    val iwork = Future { blocking { index.persist(check, symbols, commitIndex, boost) } }
     val dwork = db.persist(check, symbols)
     iwork.flatMap { _ => dwork }
   }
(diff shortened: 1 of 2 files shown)


fommil:
There are a lot of options if you're prepared to perform major surgery. Have you also considered just a simple approach using mutexes? We don't really care about blocking threads because we know the full size of our app. This is terrible form in large applications, but for our server we can block the reduce memory overhead.

We must never stop work going to the writer, what we really need is a way for each write to report back that it has freed up space in a backfill buffer.

I would expect a good back pressure solution to be just as fast as the current impl.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -153,7 +156,7 @@ class SearchService(
   def refreshResolver(): Unit = resolver.update()
 
   def persist(check: FileCheck, symbols: List[FqnSymbol], commitIndex: Boolean, boost: Boolean): Future[Option[Int]] = {
-    val iwork = Future { blocking { index.persist(check, symbols, commitIndex, boost) } }
+    val iwork = Future { blocking { index.persist(check, symbols, commitIndex, boost) } }(ex)
```
fommil: this should not be bounded. The thing we need to back pressure is the indexing (which generates the memory overhead), not the persisting (which clears the heap).

ChrisCoffey: I seem to remember something about us having some profiler licenses kicking
around, is that still true? And if so, may I use one?

fommil: @rorygraves

fommil: lions-share might also be useful, but might be fiddly to set up.

rorygraves: I'll dig out the keys

ChrisCoffey:
Haha, true about all of the options. I'm hoping to avoid major surgery though 😄

rorygraves:
@ChrisCoffey @fommil  Yourkit license - I suggest you download and get a 1 month free trial and see how you get on with it.  The licenses we have are user locked once they are activated so I want use them only as a 'if needed'.

fommil:
I don't think this approach is going to work, you're just bounding the number of threads that can work on indexing. We can also do that trivially by having an `ExecutionContext` for indexing and another for writing, but that doesn't stop the indexer overtaking the writer... you might be able to tune it for your machine but it won't scale.

We really need communication back from the writer that the indexer can proceed to fill up some buffer of work.

fommil:
@ChrisCoffey before I start teaching granny to suck eggs, how much of the reactive manifesto and sequential concurrent programming are you familiar with?

There is so much buzzword bingo around the reactive manifesto that I wouldn't even consider sending you near it if you're new to it, but for SCP (and perhaps other paradigms that may be new to you) I strongly recommend Seven Concurrency Models in Seven Weeks by Paul Butcher. Clojure is used as the example but there is a scala implementation (not battle tested) in #1351 which could form the basis of a better model than akka going forward.

ChrisCoffey:
This was really just a test to see how drone handled the bounding since I was seeing strange behavior locally. Closing. Lets continue the conversation on the issue itself.

PR #1441, opened 2016-05-04
Updates the Java8Chars regex and the integration tests to expect

### LINKED ISSUE #1440: DocResolver creating incorrect URLs for java8 standard library functions
### Steps to reproduce

Attempt to look up documentation at point for any java 8 standard library function e.g. [static IntStream range(int startInclusive, int endExclusive)](http://docs.oracle.com/javase/8/docs/api/java/util/stream/IntStream.html#range-int-int-)

### Expected behaviour
It should open the documentation for the range function http://docs.oracle.com/javase/8/docs/api/java/util/stream/IntStream.html#range-int-int-

### Actual behaviour
It does not fully transform the anchor part of the URL, and leaves a comma between the arguments - causing the page to not open at the function being looked up.
e.g. http://docs.oracle.com/javase/8/docs/api/java/util/stream/IntStream.html#range-int,int-

halcyon:
Resolves #1440.

Doc lookups for java 8 stdlib functions were failing to generate correct URLs. The problem resulted from the Java8Chars regex expecting commas followed by a space.  The DocResolver was receiving signatures of the form func(arg1,arg2) instead of func(arg1, arg2).  This change updates the regex
and the integration tests to expect function signatures without the space between arguments.

### DIFF
--- a/core/src/it/scala/org/ensime/core/DocResolverSpec.scala
+++ b/core/src/it/scala/org/ensime/core/DocResolverSpec.scala
@@ -356,11 +356,11 @@ class DocResolverSpec extends EnsimeSpec
       ) shouldBe Some("http://docs.oracle.com/javase/8/docs/api/java/io/File.html#delete--")
 
       serv.resolve(
-        DocSig(DocFqn("java.lang", "Math"), Some("max(int, int)"))
+        DocSig(DocFqn("java.lang", "Math"), Some("max(int,int)"))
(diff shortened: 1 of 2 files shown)


aemoncannon:
Nice lgtm

fommil:
LGTM

fommil:
@aemoncannon btw our new LGTM plugin needs two comments beginning LGTM (case insensitive), it would appear it is really picky and doesn't like "nice, lgtm". We could probably change the regex to find lgtm anywhere in a comment.

fommil:
Ha, also works if one person says LGTM twice.

PR #1442, opened 2016-05-05
[WIP] Penalizes noisy inner classes with a negative boost

ChrisCoffey:
I need a deeper understanding of Lucene's internals to really know what's going on here, but the test case where I expect one of the less-noisy internal classes to appear second is non-deterministic.

Basically, only the initial entry `org.example2.Baz` has a unique score, all others have the same score. I'm unsure why this is, so any insights would be appreciated.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/IndexService.scala
+++ b/core/src/main/scala/org/ensime/indexer/IndexService.scala
@@ -16,6 +16,7 @@ import org.ensime.indexer.lucene._
 
 import org.ensime.util.list._
 
+import scala.annotation.tailrec
 import scala.collection.JavaConversions._
 
 object IndexService extends SLF4JLogging {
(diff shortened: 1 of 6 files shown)


testing/simple/src/main/scala/org/example2/Baz.scala
```diff
@@ -4,8 +4,8 @@ package org.example2
 
 class Baz {
   class Wibble {
-    class baz
-    object baz
+    class Baz
```
fommil: I'm not sure my example really reproduces the problem properly. The ticket has a better example using scalatest inner classes.

core/src/main/scala/org/ensime/indexer/IndexService.scala
```diff
@@ -84,15 +85,33 @@ class IndexService(path: File) {
 
   private val lucene = new SimpleLucene(path, analyzers)
 
+  private def computePenalty(fqn: String): Float = {
+    @tailrec
+    def compute(cs: List[Char], n: Int): Int =
```
fommil: ouch, that's going to create loads of object creation churn.

fommil: I'm sure there is a way to just count characters in a String

fommil:
@ChrisCoffey ! Ping, do you have any more thoughts on this? (it also needs a rebase)

ChrisCoffey:
still working through Lucene in Action, I'll rebase once I get the feature working correctly 😄

fommil:
@ChrisCoffey need any more help with this?

fommil:
I'm about to clean up branches and `master` is about to disappear, can you rebase against 2.0 and raise a new PR?

PR #1443, opened 2016-05-05
Simple semaphore-based backpressure

ChrisCoffey:
This should address the memory explosion in #1341. I've run a couple of tests and memory usage during indexing seems to be pretty constant (just using ), although I'd need to run a lot more tests to be positive. Because different files may have more or less symbols defined in them, the load on each thread is asymmetrical (think jar vs. classfile).

If we want to smooth this more I'm willing to change `extractSymbolsFromClassOrJar`, just let me know what you think.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -81,7 +81,8 @@ class DatabaseService(dir: File) extends SLF4JLogging {
 
   def persist(check: FileCheck, symbols: Seq[FqnSymbol])(implicit ec: ExecutionContext): Future[Option[Int]] =
     db.run(
-      fileChecks.filter(_.filename === check.filename).result.headOption.flatMap {
+      //Only try to insert the first time we see a check. Only called from refresh
+      fileChecks.filter(_.filename === check.filename).result.headOption.flatMap { 
         case Some(_) =>
           DBIO.successful(None)
         case _ =>
--- a/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
+++ b/core/src/main/scala/org/ensime/indexer/DatabaseService.scala
@@ -81,7 +81,12 @@ class DatabaseService(dir: File) extends SLF4JLogging {
 
   def persist(check: FileCheck, symbols: Seq[FqnSymbol])(implicit ec: ExecutionContext): Future[Option[Int]] =
     db.run(
-      (fileChecksCompiled += check) andThen (fqnSymbolsCompiled ++= symbols)
+      fileChecks.filter(_.filename === check.filename).result.headOption.flatMap {
+        case Some(_) =>
+          DBIO.successful(None)
+        case _ =>
+          fileChecksCompiled += check
+      } andThen (fqnSymbolsCompiled ++= symbols)
     )
 
   private val findCompiled = Compiled {
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -125,18 +125,14 @@ class SearchService(
         // Using an iterator so each list of symbols may be collected as soon as its used. 
         extractSymbolsFromClassOrJar(base).flatMap {
           case (symStream, vJar) =>
-            var x = 0
             val fs = symStream map (
-              persist(check, _, commitIndex = false, boost = boost) map {
-                case Some(n) => check.synchronized { x += n }
-                case _ =>
-              }
+              persist(check, _, commitIndex = false, boost = boost)
             )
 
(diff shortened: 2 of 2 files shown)


fommil:
Can you please use the actual Semaphore class from java's concurrent library?

fommil:
This doesn't seem to be the right place to apply the wait. It needs to be out the per class level, not a per-directory level

fommil:
It's also not the right place to release the lock either :smile: the line above you is a future see you need to chain onto the success or failure paths.

ChrisCoffey:
In addition to switching over to a real Semaphore 😄  I also changed the symbol extraction code to use an `Iterator` rather than a `Stream` to avoid caching all of the previously written symbols. 

This introduces a cached threadpool for the search service in order to avoid pool-related deadlocks that were occurring with the standard akka pool we were using.

I also needed to move the vfs jar cleanup until after the iterator has been full processed. This is pretty leaky, but I feel that it's alright because it doesn't leave the `refresh()` scope. In the leaky abstraction vein, I also found that I needed to update the `db.persist` call to only load a file check once. I'm not terribly concerned about this change because we remove stale records from the db prior to indexing/persisting new ones, but please correct me if I'm wrong here.

fommil:
I'll have to have a think about this. Stream only keeps refs alive if you have a ref to the head.

fommil:
There is far too much going on now, can we focus on the semaphore thing and save the other changes for later PRs? This code is already too complex and I think adding the semaphore is about the limit of how complex we should make it before a rewrite.

ChrisCoffey:
Agreed, this is now painfully complicated. I initially tried to just use a semaphore w/o a lazy collection, but that resulted in either a deadlock or buffering an entire jar's worth of symbols into memory depending upon how I implemented it. 

To apply backpressure we need to delay translating classfiles into symbols (or even delay pulling classfiles from a jar), which means the old code's approach of `scan(vJar) flatMap extractSymbols...` doesn't work. That led me to creating the `Iterator`-based approach I'm using here (you're right that this could be a `Stream` instead) along with the hideous return type from `extractSymbolsFromClassOrJar`. The ugliness in the return came about because of the vfs cleanup logic; because we don't want to buffer everything in `eSFCOJ` and instead return a lazy collection, we also need to return a reference to the jar so it may be cleaned up once the lazy collection has been consumed. We could create a `case class` specifically for this purpose which could encapsulate some of this logic...

Finally, there is the change to `DBService` so that we no longer persist files each time we see them. The old code worked great under the assumption that we would index an entire jar at once, but now that they've been chunked out into classfiles, the same jar name will be passed to the `DBService` many times, which frequently resulted in primary-key violations & subsequent indexer freezes during my testing.

To be honest, I really hope that I've missed something simple/ made an incorrect assumption here that can cut through all of this complexity, so if that's the case please let me know!

ChrisCoffey:
I'll also address the test failure when I have time this week.

fommil:
why not put the lock in the `scan` code?

ChrisCoffey:
gah, damn. forest for the trees... thanks

I'll look into that tonight/tomorrow morning, but it should work fine & cut out all that crap in `indexBase` 😄

ChrisCoffey:
Thought about this a bit & I'm don't think moving the semaphore to `scan` resolves the problem. If I move the lock to `scan` then it will delay when we pull a classfile from a jar, but with a strict collection that leads to a livelock if the jar has more classfiles than the semaphore has tokens because we need to persist a classfile before releasing the lock & persistence can't happen until a result is returned.

Alternatively, making `scan` lazy means we can hide the resource disposal code in the guts of some wrapper type, but I don't think that makes a lot of difference as we still need most of the other stuff. If you prefer this approach I'm more than happy to go with it as it makes the intent of `indexBase` clearer.

fommil:
I don't see the problem. We want the scan to block and it's a semaphore per class.

ChrisCoffey:
Assume the semaphore has 10 tokens & we have a jar w/ 11 classfiles. With a strict collection, all 10 tokens will be handed out by the semaphore before all 11 classfiles are read. Reading the 11th classfile will block & prevent `scan` from ever returning.

Alternatively, making `scan` lazy doesn't remove any complexity or jar buffering overhead, just moves it around. I say this b/c `vfs2` is array based, so we'll still need to buffer all of the contents w/in `scan` if we make it lazy (or implement our own `findFiles`) & we'll still need to nuke the vfs jar when we're done with it.

Does that help clarify my concerns?

ChrisCoffey:
Honestly, this has turned out to be far more complicated than I expected & I'm perfectly happy to close this PR and re-asses if you want.

fommil:
Hehe, no problem. I'll have a look at some point over the coming weeks.

timcharper:
"re-asses" is the best way to spell that. +100.

Issue #1438, opened 2016-05-04
don't have snapshot dependencies


fommil:
- [x] `scala-refactoring`
- [x] file watchers
- [x] `scala-debugger`

fommil:
file watcher is on its way to central...

rorygraves:
👍

PR #1446, opened 2016-05-08
use stable release of scala-refactoring

fommil:
and only require one LGTM

### DIFF
--- a/.lgtm
+++ b/.lgtm
@@ -1,0 +1,2 @@
+approvals = 1
+pattern = "(?i)LGTM"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,11 +22,11 @@ install:
   - cmd: SET AKKA_TEST_TIMEFACTOR=10
   - cmd: SET SBT_TASK_LIMIT=2
 build_script:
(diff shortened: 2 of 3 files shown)


rorygraves:
LGTM ;)

fommil:
blocked on http://search.maven.org/#search%7Cga%7C1%7Cscala-refactoring

fommil:
@sschaef can you please make a `_2.10` / `_2.11` style release of `scala-refactoring`?

Also, it needs to work on `2.11.6` because that's what I use at work.

fommil:
note this is blocking #1438 and therefore the 1.0 release.

fommil:
it looks like in https://github.com/scala-ide/scala-refactoring/commit/2128ef75a962d7337c8c678b460192fef7ba9c24 this was done as a workaround to scala publication process, so the 2.11.x jars should all be identical (they aren't actually making releases that depend on the full version).

So we can workaround it by now by basically hardcoding the version of scala-refactoring, but hopefully @sschaef can get back to publishing `binary` cross versions instead of full, as of the next release.

fommil:
LGTM

Issue #1445, opened 2016-05-08
build index and exit


fommil:
An enabler for #1208 it would be good (and easy) to be able to index a project from the CLI.

Use cases include:

- part of an automated workspace setup (I do this a lot at work) so that initial ENSIME indexing (and CTAG building etc) can all be triggered from a single command.
- simplifying our emacs integration tests

and could possibly even be used as a way to cache and speed up our own `it:test` configuration runs.

yyadavalli:
Hi, I'm trying to get started on this
If my understanding is correct we need something that's sort of a client (similar to vim client) that will start the ensime server, wait till the index is built/updated and then close it.
Please correct me if this approach is wrong

fommil:
no, even simpler. It can just be a main method that we invoke directly, waits for the index refresh to finish, then exits cleanly. In fact, no reason we can't just pass a flag to the existing main.

I **did** say Low Hanging Fruit :grinning:

yyadavalli:
Hmm, so something like

```
val ensimeFlag = propOrNone("ensime.flag").getOrElse("")

val mainActor = system.actorOf(Props(new ServerActor(config, protocol)), "ensime-main")

val indexingComplete = // check if indexing is completed

if(ensimeFlag == "indexOnly" && indexingComplete)
  mainActor ! ShutdownRequest("Completed indexing", isError = false)
```

in the following location

https://github.com/ensime/ensime-server/blob/master/server/src/main/scala/org/ensime/server/Server.scala#L136

Then, I need to figure out how to check indexing is complete

fommil:
Probably even simpler than that, just stick a cleanup/shutdown hook onto the Future returned by refresh.

PR #1447, opened 2016-05-10
Trigger shutdown after indexing is complete

yyadavalli:
Using system property `ensime.exitAfterIndex=true` will trigger a shutdown after the indexing is complete
- Send a new message type `CloseOnIndexUpdate` back to server to trigger the shutdown

Solves #1445

### DIFF
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -17,7 +17,7 @@ import scala.util._
 import org.ensime.util.file._
 import org.ensime.util.FileUtils
 
-case object CloseOnIndexUpdate
+final case class ShutdownRequest(reason: String, isError: Boolean = false)
 
 /**
  * The Project actor simply forwards messages coming from the user to
@@ -74,7 +74,7 @@ class Project(
         log.debug(s"created $inserts and removed $deletes searchable rows")
         val exitAfterIndex = propOrFalse("ensime.exitAfterIndex")
         if (exitAfterIndex)
-          context.parent ! CloseOnIndexUpdate
+          context.parent ! ShutdownRequest("Index only run", isError = false)
       case Failure(problem) =>
         log.warning(problem.toString)
         throw problem
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -72,8 +72,7 @@ class Project(
         // we could also just blindly send this on each connection.
         broadcaster ! Broadcaster.Persist(IndexerReadyEvent)
         log.debug(s"created $inserts and removed $deletes searchable rows")
-        val exitAfterIndex = propOrFalse("ensime.exitAfterIndex")
-        if (exitAfterIndex)
+        if (propOrFalse("ensime.exitAfterIndex"))
           context.parent ! ShutdownRequest("Index only run", isError = false)
       case Failure(problem) =>
         log.warning(problem.toString)
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -72,8 +72,8 @@ class Project(
         // we could also just blindly send this on each connection.
         broadcaster ! Broadcaster.Persist(IndexerReadyEvent)
         log.debug(s"created $inserts and removed $deletes searchable rows")
-        val ensimeFlag = propOrNone("ensime.flag").getOrElse("")
-        if (ensimeFlag == "indexOnly")
+        val exitAfterIndex = propOrFalse("ensime.exitAfterIndex")
+        if (exitAfterIndex)
           context.parent ! CloseOnIndexUpdate
       case Failure(problem) =>
         log.warning(problem.toString)
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -12,9 +12,12 @@ import org.ensime.indexer._
 
 import scala.collection.immutable.ListSet
 import scala.concurrent.duration._
+import scala.util.Properties._
 import scala.util._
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -69,6 +72,9 @@ class Project(
         // we could also just blindly send this on each connection.
         broadcaster ! Broadcaster.Persist(IndexerReadyEvent)
         log.debug(s"created $inserts and removed $deletes searchable rows")
+        val ensimeFlag = propOrNone("ensime.flag").getOrElse("")
```
fommil: it would be better if you used a boolean flag like `ensime.exitAfterIndex`

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -69,6 +72,9 @@ class Project(
...
+        if (ensimeFlag == "indexOnly")
+          context.parent ! CloseOnIndexUpdate
```
fommil: no need for `CloseOnIndexUpdate`, just send `ShutdownRequest`

fommil:
ok, looking good. The bigger piece of work with this idea is that it can be used to pre-index the `testing` projects for re-use in the integration tests. That would involve writing an `sbt` task to launch the server on the `testing` projects (this might actually make sense as a contribution to `sbt-ensime`!) and then copying over the indexes when we set up the integration tests. But let's do this one thing at a time and get this initial piece working, then head over to https://github.com/ensime/ensime-sbt to create a task for the programmatic indexing of a project.

core/src/main/scala/org/ensime/core/Project.scala
yyadavalli: `ShutdownRequest` is defined in the `server` module. I thought sending it would bring in a dependency for `core` on `server`. Please correct me if I am wrong.

yyadavalli:
Thanks, updated the variable, please check my comment on the message sent. I will check what should be done in `ensime-sbt`

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -12,10 +12,13 @@ import org.ensime.indexer._
 
 import scala.collection.immutable.ListSet
 import scala.concurrent.duration._
+import scala.util.Properties._
 import scala.util._
 import org.ensime.util.file._
 import org.ensime.util.FileUtils
 
+case object CloseOnIndexUpdate
```
fommil: you shouldn't need this case object

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -69,6 +72,9 @@ class Project(
         // we could also just blindly send this on each connection.
         broadcaster ! Broadcaster.Persist(IndexerReadyEvent)
         log.debug(s"created $inserts and removed $deletes searchable rows")
+        val exitAfterIndex = propOrNone("ensime.exitAfterIndex").map(_.toBoolean).getOrElse(false)
```
fommil: there must be a way to get a property with a fallback default in the stdlib

core/src/main/scala/org/ensime/core/Project.scala
yyadavalli: I cannot directly send `ShutdownRequest` as it is defined in server module and not visible in core. Without the case object what option do I have to send callback

fommil: move `ShutdownRequest` to the `ProjectActor`

fommil: or there might be an idiomatic akka way to tell an actor to shutdown and take the system with it.

fommil: @rorygraves thoughts?

fommil: actually, it's just easiest to move the message to `core`

rorygraves: There is no idomatic way as far as I'm aware - I don't think poison pilling the actor does not give the behaviour you want.

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -69,6 +72,9 @@ class Project(
         // we could also just blindly send this on each connection.
         broadcaster ! Broadcaster.Persist(IndexerReadyEvent)
         log.debug(s"created $inserts and removed $deletes searchable rows")
+        val exitAfterIndex = propOrFalse("ensime.exitAfterIndex")
```
rorygraves: A more idomatic was to do this would be to listen to the broadcaster event and triggering the shutdown - this code is effectively  a weird side effect on the behaviour here.

rorygraves:
Ok, my c cents - move the event so its visible (don't create an extra one - you currently have two bits of random logic in the code - the 'oh I'm in a special mode, send indexer ready and then send a shutdown command.  And a shutdown command which triggers the send of another shutdown command.

So I would go with moving the message class, which reduces that count by 1.  My deeper feeling on this is that it feels like it should be a separate main method spinning up just the indexer.  We are mixing concerns here  ( not that we don't everywhere else mind).

yyadavalli:
Ok, moved the message for now. I will wait for more feedback on writing a separate main method.

rorygraves:
I think leave that - it would add a huge amount of code.  I'm just becoming a bit of a pedant at the moment ;).

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -69,6 +72,9 @@ class Project(
...
+        if (exitAfterIndex)
```
fommil: I'd probably have not bothered introducing a `val` here, but its no big deal

fommil:
LGTM

fommil:
LGTM

fommil:
thank you! This is very useful.

PR #1448, opened 2016-05-12
Parse scalariform intents

jodersky:
Scalariform intents currently aren't correctly parsed from s-expressions. This PR adds explicit parsing.

Also note that `descriptor.preferenceType` actually has a parsing function that takes a string and returns a preference value. To stay consistent, and since data is read from s-expressions (not strings), this PR copies the behaviour of the parsing function.

### DIFF
--- a/core/src/main/scala/org/ensime/config/ScalariformFormat.scala
+++ b/core/src/main/scala/org/ensime/config/ScalariformFormat.scala
@@ -21,7 +21,12 @@ trait ScalariformFormat {
         descriptor.preferenceType match {
           case BooleanPreference => sexp.convertTo[Boolean]
           case IntegerPreference(_, _) => sexp.convertTo[Int]
-          case IntentPreference => sexp.convertTo[String]
+          case IntentPreference => sexp.convertTo[String].toLowerCase match {
+            case "preserve" => Preserve
+            case "force" => Force
+            case "prevent" => Prevent
+            case intent => throw new DeserializationException(s"Unknown intent: $intent")
+          }
         }
       }
 
@@ -45,7 +50,7 @@ trait ScalariformFormat {
     ): Sexp = descriptor.preferenceType match {
       case BooleanPreference => value.asInstanceOf[Boolean].toSexp
       case IntegerPreference(_, _) => value.asInstanceOf[Int].toSexp
-      case IntentPreference => value.asInstanceOf[Intent].getClass.getSimpleName.toSexp
+      case IntentPreference => value.getClass.getSimpleName.replaceAll("\\$", "").toLowerCase.toSexp
     }
 
     def write(f: FormattingPreferences) = {
--- a/core/src/test/scala/org/ensime/config/ScalariformFormatSpec.scala
+++ b/core/src/test/scala/org/ensime/config/ScalariformFormatSpec.scala
@@ -15,11 +15,13 @@ class ScalariformFormatSpec extends EnsimeSpec {
 
   val prefs = FormattingPreferences().
     setPreference(DoubleIndentClassDeclaration, true).
-    setPreference(IndentSpaces, 13)
+    setPreference(IndentSpaces, 13).
+    setPreference(DanglingCloseParenthesis, Preserve)
 
   "ScalariformFormat" should "parse some example config" in {
     val text = """(:doubleIndentClassDeclaration t
-                     :indentSpaces 13)"""
+                     :indentSpaces 13
+                     :danglingCloseParenthesis "preserve")"""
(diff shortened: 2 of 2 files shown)


fommil:
LGTM, needs a rebase

fommil:
@jodersky ! Ping

jodersky:
rebased

core/src/main/scala/org/ensime/config/ScalariformFormat.scala
```diff
@@ -21,7 +21,12 @@ trait ScalariformFormat {
         descriptor.preferenceType match {
           case BooleanPreference => sexp.convertTo[Boolean]
           case IntegerPreference(_, _) => sexp.convertTo[Int]
-          case IntentPreference => sexp.convertTo[String]
+          case IntentPreference => sexp.convertTo[String].toLowerCase match {
+            case "preserve" => Preserve
+            case "force" => Force
+            case "prevent" => Prevent
+            case intent => sys.error("Unknown scalariform intent: " + intent)
```
fommil: `error` is a bit harsh for this. There is already an exception type for s-express failures.

core/src/main/scala/org/ensime/config/ScalariformFormat.scala
```diff
@@ -45,7 +50,11 @@ trait ScalariformFormat {
     ): Sexp = descriptor.preferenceType match {
       case BooleanPreference => value.asInstanceOf[Boolean].toSexp
       case IntegerPreference(_, _) => value.asInstanceOf[Int].toSexp
-      case IntentPreference => value.asInstanceOf[Intent].getClass.getSimpleName.toSexp
```
fommil: we probably don't need the `asInstanceOf`, can we survive with `.getClass.getSimpleName.toLowerCase.toSexp`?

jodersky: that's correct, don't know why I wrote that

core/src/main/scala/org/ensime/config/ScalariformFormat.scala
jodersky: Would `org.ensime.sexp.DeserializationException` be more suited?

jodersky:
The latest commit should address the issues.

As a bonus, I also updated the `ScalariformFormatSpec` test suite. Btw, I just saw the same test suite exists in ensime-sbt. Would it be worth opening a PR with the added test? Or is that more work for you than it is useful (there is already a scripted test that checks intent serialization)?

fommil:
The ensime-server build is a subdirectory of sbt-ensime as it's used as a test project, but we don't run its tests.

jodersky:
ahh, I pressed github "update branch" button :( didn't know it would give me merge commit

fommil:
It's ok, we squash all PRs so it works out fine, no need to rewrite history.

jodersky:
ok, I thought I read somewhere in the contributor guidelines we should do
it ourselves

rorygraves:
@jodersky We changed the github config recently - so PRs are squashed by default.

jodersky:
the tests pass locally on my machine, could drone be having an issue?

rorygraves:
The fact that is passed on drone for 2.10 and not 2.11 makes me think it is a transient failure, but I think this one needs the judgement of @fommil

fommil:
Can you click through and see the message?

fommil:
looks like the emacs tests are failing since an earlier commit, might be an emacs side failure. I was going to do some CI tidying on the emacs side anyway tonight. I'll unblock.

jodersky:
from the error messages, it looks like some issues with emacs tests

jodersky:
as you said 1m ago :)

fommil:
oh, I think I know what it is. I had to change to using melpa instead of melpa-stable and now some third party lib's behaviour has changed causing our tests to fail.

fommil:
darn, I made a mistake when releasing `scala-mode` and the stable release is actually `scala-mode2` so I can't update this until it propagates (a few hours?). I might push out a quick fix.

jodersky:
no worries, this PR isn't at all critical and I still have a locally
published version for my purposes

fommil:
I'll push a quick fix, it was as I expected. yasnippet now indents after insertion.

fommil:
@jodersky I kicked off a build after fixing the tests client side, merge when you're happy

fommil:
actually, I'll merge. You've been patient enough :smile: (you should have permissions after we give an LGTM)

jodersky:
oh alright, I wasn't sure if I should wait for another core developer. Good to know for the future! thanks

Issue #1457, opened 2016-05-15
Ensime fails to boot if loading JAR files from paths containing URI-escaped-chars


timcharper:
Coursier dependency management system will encode unsafe characters, such as `:`, by URI encoding the characters. These are being treated as characters that should be decoded when passed to apache-commons vfs2 `resolveFile`, resulting in the following error:

```
Caused by: org.apache.commons.vfs2.FileSystemException: Could not replicate "file:///Users/timcharper/.coursier/cache/v1/http/artifactory.corporate.domain:8080/artifactory/libs-release/org/mortbay/jetty/jetty/6.1.26.cloudera.4/jetty-6.1.26.cloudera.4-sources.jar" as it does not exist.
    at org.apache.commons.vfs2.provider.AbstractFileSystem.replicateFile(AbstractFileSystem.java:418)
    at org.apache.commons.vfs2.provider.zip.ZipFileSystem.<init>(ZipFileSystem.java:61)
    at org.apache.commons.vfs2.provider.jar.JarFileSystem.<init>(JarFileSystem.java:50)
    at org.apache.commons.vfs2.provider.jar.JarFileProvider.doCreateFileSystem(JarFileProvider.java:82)
    at org.apache.commons.vfs2.provider.AbstractLayeredFileProvider.createFileSystem(AbstractLayeredFileProvider.java:89)
    at org.apache.commons.vfs2.provider.AbstractLayeredFileProvider.findFile(AbstractLayeredFileProvider.java:63)
    at org.apache.commons.vfs2.impl.DefaultFileSystemManager.resolveFile(DefaultFileSystemManager.java:693)
    at org.apache.commons.vfs2.impl.DefaultFileSystemManager.resolveFile(DefaultFileSystemManager.java:649)
    at org.apache.commons.vfs2.impl.DefaultFileSystemManager.resolveFile(DefaultFileSystemManager.java:605)
    at org.ensime.vfs.package$RichVFS$$anonfun$vjar$extension0$1.apply(EnsimeVFS.scala:66)
    at org.ensime.vfs.package$RichVFS$$anonfun$vjar$extension0$1.apply(EnsimeVFS.scala:66)
    at org.ensime.vfs.package$RichVFS$.org$ensime$vfs$RichVFS$$withContext$extension(EnsimeVFS.scala:52)
    ... 24 common frames omitted
21:25:28.588 INFO  None Server - Shutdown complete
```

[...]

PR #1458, opened 2016-05-15
properly handle files with uri-encoded components

timcharper:
vfs.resolveFile receives a String URI, and will parse and decode URI
components. This means if the file name contains URI-encoded components
(such as %3A, as courier will output to escape unsafe FS characters like
:), they will be decoded, and the wrong filename will be received.

fixes https://github.com/ensime/ensime-server/issues/1457


timcharper:
Sorry so sloppy, I didn't write tests for this.

timcharper:
Okay there is a test now. It's not the best test, it's a test and it beats all the rest.

timcharper:
The other tests in `util/src/test/scala/org/ensime/util` aren't using the mentioned `EnsimeSpec`. It appears that the class isn't available.

fommil:
oh right, `testutil` depends on `utils`

util/src/main/scala/org/ensime/vfs/EnsimeVFS.scala
```diff
@@ -46,6 +46,20 @@ object `package` {
   private val zipFileField = classOf[ZipFileSystem].getDeclaredField("zipFile")
   zipFileField.setAccessible(true)
 
+  object URLParamEncoder {
+    private val unsafe = " %"
+    def encode(input: String): String = {
+      val resultStr = new StringBuilder()
+      input.toCharArray.foreach { ch =>
+        if (unsafe contains ch)
```
fommil: I shudder to think how inefficient this is in terms of scala collections.

fommil:
LGTM merge when you're happy

fommil:
oh, looks like you need to rebase before you can do that

util/src/main/scala/org/ensime/vfs/EnsimeVFS.scala
timcharper: good point; I'll use a set instead

fommil: I'd just use an `if` statement to be honest. The object creation that happens when `String` is treated as a `Traversable` is pretty crazy.

util/src/main/scala/org/ensime/vfs/EnsimeVFS.scala
```diff
@@ -46,6 +46,22 @@ object `package` {
   private val zipFileField = classOf[ZipFileSystem].getDeclaredField("zipFile")
   zipFileField.setAccessible(true)
 
+  object URLParamEncoder {
+    def encode(input: String): String = {
```
lukasz-golebiewski: Won't `input.map{ c =>  if ((c == ' ') || (c == '%')) "%%%02x".format(c.toShort) else c }` do the same thing?

fommil:
heh, sorry, just merged another PR. Seems github is getting really fussy about that nowadays! It used to be happy with PRs that had a clean rebase.

util/src/main/scala/org/ensime/vfs/EnsimeVFS.scala
timcharper: Yes, but not as efficiently. I think?

timcharper: trying to reduce allocations since this is a bit of a hot-path in this code

fommil: isn't that returning `String` or `Char`? From an efficiency point of view we create a lot of `URI`s so the current approach could contribute to quite a bit of churn. To really tackle that we might want to drop the `foreach` and use a `while` loop. But no need to prematurely optimise, let's see if it's a problem.

timcharper:
```
scala> "asdf".map { _ => "B" }
res0: scala.collection.immutable.IndexedSeq[String] = Vector(B, B, B, B)
```

Yeah... definitely not as efficient.

timcharper: @fommil I've already dropped the `foreach`

fommil:
@rorygraves will be very proud :grinning: https://skillsmatter.com/skillscasts/6939-optimising-scala-for-fun-and-profit

fommil:
I really wish we had parallel matrix builds in drone...

util/src/main/scala/org/ensime/vfs/EnsimeVFS.scala
lukasz-golebiewski: What about `input.map{ c => if ((c == ' ') || (c == '%')) "%%%02x".format(c.toShort) else c }(scala.collection.breakOut)` ? ;-)

lukasz-golebiewski: LGTM! Didn't notice you were mapping to a `String`

lukasz-golebiewski: You could always do 
```
@ "asdb".flatMap(_ => "B")(scala.collection.breakOut)
res2: String = "BBBB" 
```

fommil: have you done an object allocation profile of that? I bet it's not pretty! :laughing:

fommil:
you managed to get out of date with master again

fommil:
ok, I removed the (new) restriction that branches have to be up to date. It might bite us occasionally, but I'm willing to live dangerously for the convenience.

fommil:
oh it's bedtime in your TZ, I'll merge.

util/src/main/scala/org/ensime/vfs/EnsimeVFS.scala
lukasz-golebiewski: You're right :) `for (x <- this) b ++= f(x).seq`

rorygraves:
@fommil  - Proud, but sad that this kind of stuff is still required.

fommil:
it would be interesting to use one of those automagic cleanup plugins you talked about, see if it makes a difference on a huge project.

PR #1459, opened 2016-05-15
Index method descriptors

### LINKED ISSUE #1190: correctly index methods by FQN and Descriptor
I'm currently working on this as part of the Graphpocalyse but it might make sense to split this out as a separate piece of work and support it in H2 first of all.

We can trivially get the FQN and Descriptor from the bytecode, but getting it from the scalac `Symbol` is tricky.

`Helpers.symbolIndexerName` and `DocFinding.javaFqn` should really be sharing the same logic, which could be implemented in the `FullyQualifiedName` domain model... and then translated into `Descriptor`s (logic exists) and `DocFqn`s.

In any case, my WIP is in this general area https://github.com/fommil/ensime-server/blob/graphpocalypse/core/src/main/scala/org/ensime/model/Helpers.scala#L128

And there are a bunch of other places in the code where we do something similar, so obviously there is a good opportunity for code reuse. I wonder how Scala IDE does this, if it does it at all. Surely `scalac` knows how to do this in the bytecode output layers.

fommil:
close #1190 

@rorygraves @sugakandrey please take a look


api/src/main/scala/org/ensime/api/outgoing.scala
```diff
@@ -475,11 +475,11 @@ final case class BasicTypeInfo(
 
 final case class ArrowTypeInfo(
     name: String,
+    fullName: String,
```
rorygraves: Is this a fully qualified name?  - is there a better variable name or type we can use?

fommil: nope, oh god yes, and nope... it's part of the API.

core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
```diff
@@ -422,17 +322,17 @@ class RichPresentationCompilerSpec extends EnsimeSpec
         val mem = sup.tpe.members.find(_.name == "banana").get.asInstanceOf[NamedTypeMemberInfo]
         val tpe = mem.tpe.asInstanceOf[ArrowTypeInfo]
 
-        tpe.resultType.name shouldBe "List"
+        tpe.resultType.name shouldBe "List[String]"
```
rorygraves: Yay!!!!!

rorygraves:
Like it - I think there are some big wins here.
As noted, I suspect all of the emacs tests are going to break, but that means mostly that the emacs tests are wrong and need to be fixed.

LGTM

fommil:
other than the fact that the emacs tests have been silently failing on a false positive for, weeks, months, years? It looks like there are a few subtle API interpretation changes.

The inspector ui test is sending this message and receiving `nil`, which is blowing up the url normaliser.

```elisp
(ensime-rpc-doc-uri-for-symbol "java.lang.String" "equalsIgnoreCase" "(x$1: String): Boolean")
```

but this looks suspect, especially the `member-sig` bit which is not a descriptor. If the server is returning this, it is not good, but I suspect the client might be constructing it which is also not good.

fommil:
FYI this is my emacs branch to fix this https://github.com/fommil/ensime-emacs/tree/fqn-fixes

fommil:
ok, looks like the server is sending a non-FQN string for `:signature-string`

fommil:
I expect the drone build to fail for 2.11 in the emacs tests. But if everything else is green, I'll merge the emacs update (temporarily red lights on the emacs side), then rebuild and merge when this turns green.

fommil:
emacs change merged. Rebuilding. Also had a transient on Windows, I think their servers are acting up.

fommil:
(braces for impact...)

PR #1461, opened 2016-05-15
Switch to sbt-coursier 1.0.0-M12

alexarchambault:
~~DON'T MERGE, this uses things from sonatype staging, and requires Java >= 7 as is.~~

This doesn't fix https://github.com/alexarchambault/coursier/issues/244... should be for the next milestone.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -24,4 +24,4 @@ ivyLoggingLevel := UpdateLogging.Quiet
 
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.0")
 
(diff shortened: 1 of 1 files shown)


fommil:
Does this have the cache for missing javadocs so it doesn't keep retrying?

alexarchambault:
1.0.0-M11 is already supposed to do that.

fommil:
Oh, I'm still seeing it trying to get some javadocs/src that aren't published.

alexarchambault:
It prints error messages, even if the cache says the javadoc or sources doesn't exist (and no request was made).

fommil:
Oh cool, printing is ok.

alexarchambault:
Not on Central yet...

Issue #635, opened 2014-09-28
show scala's String rep of a type to the user, not our custom rendering of the name


fommil:
i.e. rethink #629 including refined types and so on.

Without this support, type information in typelevel projects such as shapeless is mostly useless.

this means sanitising use of fqn/type/fullName (possibly dropping the latter most)

fommil:
I think this is now feasible. We might need to add shapeless to the tests to create refinement but as an example of the ensime code where we have a refinement, look at `common.scala` `DeclaredAs.allDeclarations` (which we might even want to replace with the shapeless enum pattern)

PR #1466, opened 2016-05-23
Pretty type-at-point

### LINKED ISSUE #1464: pretty print of Tuple and Function types
Fallout from recent refactor.

This is the type-at-point of the method call inside `FileEdit.compare`
`scala.Tuple4[java.io.File, scala.Int, scala.Int, java.lang.String] => scala.Int`

wouldn't it be nicer if it was using the syntax sugar?

Also, we should think about whether we really want the hack for single-parameter arrow types.

fommil:
closes #1464 #635 making type-at-point a lot prettier for tuples, `scala.FunctionX`, refined types and classes designed to be used infix. The test cases show exactly what it does.

Might break the emacs tests since this could affect string assertions.

### DIFF
--- a/core/src/it/scala/org/ensime/core/TypeToScalaNameSpec.scala
+++ b/core/src/it/scala/org/ensime/core/TypeToScalaNameSpec.scala
@@ -11,18 +11,22 @@ class TypeToScalaNameSpec extends EnsimeSpec
     with ReallyRichPresentationCompilerFixture {
   import ReallyRichPresentationCompilerFixture._
 
-  val original = EnsimeConfigFixture.EmptyTestProject
+  val original = EnsimeConfigFixture.ShapelessTestProject
 
   it should "calculate the TypeInfo at point" in withPresCompiler { (config, cc) =>
     runForPositionInCompiledSource(
       config, cc,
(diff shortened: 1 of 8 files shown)


fommil:
yup, emacs tests are failing. I'll need to change the assertions.

fommil:
ok, I think this has flagged up a genuine bug in the emacs side, possibly in `ensime--build-yasnippet-for-call`.

I think that needs a full evening to diagnose. If anybody desperately wants this support, that's where to start helping out. I think the brackets in the `doBlock` of "Test expanding parameter lists" is the trigger.

fommil:
a snippet is being created that looks like `(${1:block: Int)})` which has a stray `)` inside the curlys.

fommil:
yeah, this is all coming from our client-side `ensime-parse-type-info-from-scala-name`. That code definitely needs a clean up.

fommil:
adding `"(Char) => Boolean"` as a test case to `ensime-emacs-test-parse-type-info-from-scala-name` might be a possible way to temporarily workaround this.

rorygraves:
LGTM - awaiting signoff from CI

PR #1468, opened 2016-05-24
added TypeInfo to completion responses

### LINKED ISSUE #1467: add extra info to the completion API endpoint
emac has to perform some hacks to be able to do completions, it would be good to add the required info to the endpoint, in addition to what is already there (to not break the other clients).

fommil:
close #1467 

too easy


rorygraves:
LGTM - This is sending a load more data, but assuming everything does not grind to a halt should make clients simplier.

PR #1476, opened 2016-05-30
FunctionX are arrow types

fommil:
a.k.a. "I'm AM fixing the lightbulb"


rorygraves:
LTGM - I personally would not use apply here and prefer ```build``` or something similar to distinquish, but I think thats personal taste.

rorygraves:
LGTM even - I can type, promise (when nobody is looking obviously)

fommil:
LGTM looks like it's down

PR #1478, opened 2016-05-30
Backpressure

### LINKED ISSUE #1341: backpressure in the file indexing

fommil:
close #1341

This is a pretty evil hack, but the benefit to users is worth it.

We're introducing a semaphore during indexing to link backpressure between the (CPU and memory intensive) indexing and (very slow) persistence.

We can do it properly if we ever find a scala framework as good as `core.async` in clojure.

### DIFF
--- a/core/src/main/scala/org/ensime/core/Completion.scala
+++ b/core/src/main/scala/org/ensime/core/Completion.scala
@@ -125,7 +125,8 @@ trait CompletionControl {
 
       val contextOpt = x.get match {
         case Left(tree) =>
-          logger.debug("Completing at tree:" + tree.summaryString)
+          if (logger.isTraceEnabled())
+            logger.trace("Completing at tree:" + tree.summaryString)
           tree match {
             case Apply(fun, _) =>
               fun match {
@@ -237,8 +238,6 @@ trait CompletionControl {
       }
     } while (!x.isComplete)
 
-    logger.info("Found " + members.size + " members.")
-
     // Any interaction with the members (their types and symbols) must be done
     // on the compiler thread.
     askOption[Unit] {
@@ -246,7 +245,6 @@ trait CompletionControl {
         val s = m.sym.nameString
         matchesPrefix(s, context.prefix, matchEntire = false, caseSens = caseSens) && !s.contains("$")
       }
-      logger.info("Filtered down to " + filtered.size + ".")
       for (m <- filtered) {
         m match {
           case m @ ScopeMember(sym, tpe, accessible, viaView) =>
--- a/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
+++ b/core/src/main/scala/org/ensime/indexer/FileWatchers.scala
@@ -54,8 +54,8 @@ class ClassfileWatcher(
               (JarSelector, target.getParentFile, false)
             else
               (ClassfileSelector, target, true)
-          if (log.isDebugEnabled())
-            log.debug(s"creating a ApachePollingFileWatcher watcher for $dir")
+          if (log.isTraceEnabled())
+            log.trace(s"creating an ApachePollingFileWatcher watcher for $dir")
           new ApachePollingFileWatcher(dir, selector, rec, listeners).asInstanceOf[Watcher]
         }
       }
@@ -64,12 +64,12 @@ class ClassfileWatcher(
         val classJava7WatcherBuilder = new ClassJava7WatcherBuilder()
         config.targetClasspath.map { target =>
           if (target.isJar) {
-            if (log.isDebugEnabled())
-              log.debug(s"creating a Java 7 jar watcher for ${target}")
+            if (log.isTraceEnabled())
+              log.trace(s"creating a Java 7 jar watcher for ${target}")
             jarJava7WatcherBuilder.build(target, listeners)
           } else {
-            if (log.isDebugEnabled())
-              log.debug(s"creating a Java 7 class watcher for ${target}")
+            if (log.isTraceEnabled())
+              log.trace(s"creating a Java 7 class watcher for ${target}")
             classJava7WatcherBuilder.build(target, listeners)
           }
         }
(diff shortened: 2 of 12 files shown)


core/src/main/scala/org/ensime/core/Completion.scala
```diff
@@ -125,7 +125,7 @@ trait CompletionControl {
 
       val contextOpt = x.get match {
         case Left(tree) =>
-          logger.debug("Completing at tree:" + tree.summaryString)
+          logger.trace("Completing at tree:" + tree.summaryString)
```
rorygraves: if (log.isTraceEnabled())  ?
You managed it everywhere else.

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
```diff
@@ -151,28 +151,28 @@ private class ApachePollingFileWatcher(
...
-        if (log.isDebugEnabled())
-          log.debug(s"${event.getFile} was created")
+        if (log.isTraceEnabled())
+          log.trace(s"${event.getFile} was created")
         listeners foreach (_.fileAdded(event.getFile))
       }
     def fileDeleted(event: FileChangeEvent): Unit =
       if (base == event.getFile.getName.getURI) {
-        log.info(s"$base (a watched base) was deleted")
+        log.debug(s"$base (a watched base) was deleted")
```
rorygraves: debug guard?  I'm guessing this code is called a lot

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -61,7 +62,12 @@ class SearchService(
   private val index = new IndexService(config.cacheDir / ("index-" + version))
   private val db = new DatabaseService(config.cacheDir / ("sql-" + version))
 
-  implicit val workerEC: MessageDispatcher = actorSystem.dispatchers.lookup("akka.search-service-dispatcher")
+  import ExecutionContext.Implicits.global
```
rorygraves: Do we really want the global EC here?

core/src/main/scala/org/ensime/indexer/FileWatchers.scala
fommil: no, quite rare actually. I'm not so fussed about debug guards because it's the default. Trace definitely needs a guard though.

core/src/main/scala/org/ensime/indexer/SearchService.scala
fommil: yup, it's not used for any I/O.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -61,7 +62,12 @@ class SearchService(
...
+
+  // each jar / directory must acquire a permit, released when the
+  // data is persisted. This is to keep the heap usage down and is a
+  // poor man's backpressure.
+  val semaphore = new Semaphore(Properties.propOrElse("ensime.index.parallel", "1").toInt, true)
```
fommil: oops, meant to make this 10 by default.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -308,7 +321,7 @@ class IndexingQueueActor(searchService: SearchService) extends Actor with ActorL
       if (remaining.nonEmpty)
         debounce()
 
-      import searchService.workerEC
+      import ExecutionContext.Implicits.global
```
rorygraves: no no no...

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -163,26 +171,31 @@ class SearchService(
...
-      val check = FileCheck(jar)
-      Future {
-        blocking {
-          val vJar = vfs.vjar(jar)
-          try scan(vJar) flatMap (extractSymbols(jar, _))
-          finally vfs.nuke(vJar)
+  // this method leak semaphore on every call, which must be released
+  // when the List[FqnSymbol] has been processed (even if it is empty)
+  def extractSymbolsFromClassOrJar(file: FileObject): Future[List[FqnSymbol]] = {
+    def global: ExecutionContext = null
```
rorygraves: wha what???

core/src/main/scala/org/ensime/indexer/SearchService.scala
fommil: yes yes yes, it's not used for I/O.

core/src/main/scala/org/ensime/indexer/SearchService.scala
fommil: that disables the implicit `global: ExecutionContext` meaning we have to provide our own.

fommil: (it's not strictly necessary, but it means the compiler errors if I forget to pass an explicit EC to some of this I/O code)

rorygraves: Ouch - I would suggest a comment.

core/src/main/scala/org/ensime/indexer/SearchService.scala
rorygraves: I know - I'm pretty strongly against using the global EC for anything.

rorygraves:
LGTM - I don't really like the global EC usage, but this stuff will all be refactored out later so I probably should not stress about it.

fommil:
I'm ok with the global execution context for non-blocking work (best place for it), but I agree that all I/O should be isolated if it isn't backed by a native async implementation. Turns out there is some I/O happening on the global execution context here: the Lucene code is all synchronous. I'll create a 2.0 ticket.

fommil:
@ChrisCoffey FYI

Issue #1362, opened 2016-03-30
Add generics to RawClassfile


asznajder:
`RawClassfile` class represents a _.class_ file that's a result of Java/Scala code compilation. What's currently missing in its current representation is the generics part. It's implemented now as an `Option[String]` whereas it would be great to have a domain object instead.

fommil:
Possibly related to #472

fommil:
Needs some design work on what the domain objects would look like.

asznajder:
I will be happy to work on this next week.

fommil:
probably a `List[FullyQualifiedName]` would be my guess at something more sensible.

asznajder:
Since `generics` field in `RawClassfile` has really the following grammar:

_**TypeSignature**: Z | C | B | S | I | F | J | D | FieldTypeSignature
**FieldTypeSignature**: ClassTypeSignature | [ TypeSignature | TypeVar
**ClassTypeSignature**: L Id ( / Id )* TypeArgs? ( . Id TypeArgs? )* ;
**TypeArgs**: < TypeArg+ >
**TypeArg**: * | ( + | - )? FieldTypeSignature
**TypeVar**: T Id ;_

I would like to suggest creating domain objects that are consistent with it. 

`sealed abstract class TypeSignature`
 `case class PrimitiveTypeWrapper(className: ClassName) extends TypeSignature`
 `sealed abstract class FieldTypeSignature extends TypeSignature`
 `case class ArraySignature(arrayType: TypeSignature) extends TypeSignature`
 `case class TypeVar(name: String) extends TypeSignature`
 `case class ClassTypeSignature(className: ClassName, typeArgs: List<TypeArg>) 
 `case class TypeArg(boundType: Option[Char], typeSignature: FieldTypeSignature)`

`ClassTypeSignature` doesn't handle inner types in this form and something might be wrong, but I wanted only to quickly give some insight.

Anyway, `generics` field in `RawClassfile` would be then of type: 

`case GenericClass(typeParams: Option[TypeParam], signatures: List[ClassTypeSignature])`

[...]

PR #1479, opened 2016-05-30
Added generics extraction from class signatures

asznajder:
Resolving #1362


core/src/main/scala/org/ensime/indexer/DescriptorParser.scala
```diff
@@ -33,12 +67,11 @@ object DescriptorParser {
         throw new Exception("Failed to parse descriptor type: ", other)
     }
   }
-  val PackageNamePredicate = CharPredicate.All -- ";/ "
-  val ClassNameCharPredicate = CharPredicate.All -- ";/ "
+  val PackageNamePredicate = CharPredicate.All -- "<;/ "
```
fommil: these changes will interfere with the descriptor parser

core/src/main/scala/org/ensime/indexer/DescriptorParser.scala
```diff
@@ -2,13 +2,47 @@
...
 
+import org.ensime.indexer.ClassName._
 import org.parboiled2.Parser
 import org.parboiled2.Rule1
 import org.parboiled2._
 
 import scala.annotation.switch
 import scala.util.{ Failure, Success }
 
+trait ClassParser extends Parser {
```
fommil: some things in here could be shared, but not everything. Try to use an imported `object` in preference to a mix in trait if you can.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -31,12 +31,15 @@ final case class PackageName(path: List[String]) extends FullyQualifiedName {
     case ClassName(p, _) => contains(p)
     case FieldName(c, _) => contains(c)
     case MethodName(c, _, _) => contains(c)
+    case _ => false
```
fommil: we should never need to do this, this is a sealed family.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -31,12 +31,15 @@ final case class PackageName(path: List[String]) extends FullyQualifiedName {
...
   }
   def fqnString = path.mkString(".")
   def parent = PackageName(path.init)
 }
 
-final case class ClassName(pack: PackageName, name: String) extends FullyQualifiedName with DescriptorType {
+case class ClassName(pack: PackageName, name: String)
+    extends FullyQualifiedName with DescriptorType with SignatureType {
```
fommil: we don't need it to be `SignatureType`.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -125,6 +128,115 @@ final case class MethodName(
...
 }
 
+/**
+ * Generics signature
+ */
+
+trait GenericArg extends FullyQualifiedName
+trait SignatureType extends FullyQualifiedName
+
+case class GenericClass(
```
fommil: please make case classes `final`.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -125,6 +128,115 @@ final case class MethodName(
   def fqnString = owner.fqnString + "." + name + descriptor.descriptorString
 }
 
+/**
+ * Generics signature
+ */
+
+trait GenericArg extends FullyQualifiedName
```
fommil: these are not `FullyQualifiedName`s, no need to extend from it.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -143,7 +255,7 @@ final case class Descriptor(params: List[DescriptorType], ret: DescriptorType) {
 
 final case class RawClassfile(
   name: ClassName,
-  generics: Option[String],
+  generics: Option[GenericClass],
```
fommil: this is - correctly - the only place you need to add the new data structure... no need for any of the hierarchy changes above.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -125,6 +128,115 @@ final case class MethodName(
...
+    case sthElse => classNames.exists(_.contains(o))
+  }
+
+  override def fqnString: String = {
+    classNames.map(_.fqnString).mkString
+  }
+}
+
+case class GenericClassName(
+    className: ClassName,
```
fommil: instead of redefining `ClassName`, it would be better to add extra fields here. Everything in `FullyQualifiedName` is representative of the bytecode representation and generics are erased by that time, so all generic information needs to be captured here in the `GenericClass` hierarchy.

fommil:
I'm doing some branch reorganisation, so `master` might disappear. If that happens, just rebase against `2.0` and create a PR against that branch.

fommil:
sorry, I killed the windows / OS X jobs to free up some resources... doing a bit of release prep.

PR #1482, opened 2016-05-31
Reduce heap and don't go nuts when the compiler crashes

### LINKED ISSUE #809: Don’t highlight entire file as errline when presentation compiler crashes
Whenever the presentation compiler crashes, the entire file gets highlighted as errline with a “presntation compiler crashed” title text. This isn’t particularly helpful.

I would prefer if maybe it just added a special annotation to the mode-line lighter and let scala-mode2 highlight the file normally.

### LINKED ISSUE #1257: intern Strings in VFS
I did a quick heapdump of ENSIME server in quiesscent state and found about GB of duplicated strings (thanks to YourKit). We could remove a lot of that by using `.intern`, but a bunch of it is internal to VFS so we'd have to hack their code.

If we catch every string going into `EnsimeVFS` it should help. Might slow down Java 6 a bit but this is generally a good move on modern JVMs.

* https://issues.apache.org/jira/browse/VFS-593

fommil:
close #809 and I couldn't resist having a chance to close #1257


fommil:
and you can thank @dragos for that `println` that will forever show up in our integration tests

```
... trimmed 100 lines of stacktrace ...

1464719190940-0/testing/macros/src/main/scala/def.scala] marking unit as crashed (crashedFiles: Set(/tmp/1464719190940-0/testing/macros/src/main/scala/def.scala))
```

fommil:
doh! @sirthias did another 2.11 only release of parboiled :stuck_out_tongue: reverting the upgrade...

fommil:
ok, and a potential fix for #1321 which is basically just enabling @smootoo's index again now that it's a foreign key. I'll measure performance over the next week and if it doesn't negatively impact insertions, we can keep it. I'm assuming it still makes sense to add the index.

fommil:
damnit, my macro test only works in 2.11.

fommil:
travis is acting all weird, looks like an OS problem.

fommil:
yeah, the travis thing looks like a bad cache. I deleted all caches and kicked off a fresh build.

rorygraves:
LGTM

PR #1484, opened 2016-05-31
2.0 low hanging upgrades

fommil:
all the boring upgrades


fommil:
hmm, the emacs tests are still failing. I have no idea why.

fommil:
Reminder: langtools

rorygraves:
LGTM

fommil:
windows failure was

```
15:22:20.233 ERROR akka://default/system/IO-TCP/selectors/$a/0 akka.io.TcpListener - Bind failed for TCP channel on endpoint [/127.0.0.1:10004]
java.net.BindException: Address already in use: bind
```

When that happens it shouldn't fail the test because the whole point is that we want to have something on the port so the server can't start.

fommil:
The Windows error looks like a transient, the host was acting up and even failed to post the log files. Hopefully appveyor have the problem fixed.

Issue #1494, opened 2016-06-02
symbol at point is inconsistent for java methods


fommil:
symbol at point is currently returning a weird BasicTypeInfo that doesn't make sense for java methods, e.g. see `JavaCompilerSpec.scala` for anything on a method, e.g. 

```
          case "3" =>
            info.name shouldBe "java.io.PrintStream.println(java.lang.Object)"
            info.localName shouldBe "println"
            info.`type`.name shouldBe "(java.lang.Object)void"
```

the `info.type` should be returning a sensible `ArrowTypeInfo` instead of a fake `BasicTypeInfo`.

fommil:
Moving to 1.0 as this breaks java support.

fommil:
just checked, not as bad as I thought. It still works but we don't get a nice template (but I don't think we ever did)

mwielocha:
So if this is not breaking java support then what is?

aemoncannon:
@fommil Completion templates? Those work (or have been) for me.

fommil:
@aemoncannon they don't work so well as the scala method completions because we're returning the info as a blob in `BasicTypeInfo` instead of `ArrowTypeInfo`. But, yeah, it pretty much works right now for all practical purposes.

@mwielocha yeah, I'm not sure. It might just be on your machine... can you try to reproduce on the ensime-server project itself? We've got a couple of java files, mostly in the monkeys module.

mwielocha:
Not sure what the hell happen but it started working again, sorry for the confusion.

mwielocha:
Do we want this to be the same as in `scalaName` with the `=>` operator? Or more like Java for example `(java.lang.Object): java.lang.String` or something?

fommil:
let's use Java syntax. We probably shouldn't be using the scala API to return Java Compiler objects anyway, maybe give them their own protocol.

fommil:
(remember the user gets these responses back only when querying from a `.java` file)

mwielocha:
Yeah that was my thought as well. Oh well, At least I got a first bite on this one :)

mwielocha:
You think its a good time to re-implement both java incoming and outgoing?
Just found this:
```scala
    // HACK: to expedite initial dev, Java requests use the Scala API
    case m @ TypecheckFileReq(sfi) if sfi.file.isJava => javac forward m
    case m @ CompletionsReq(sfi, _, _, _, _) if sfi.file.isJava => javac forward m
[...]

PR #1495, opened 2016-06-02
Remove deprecated completions

### LINKED ISSUE #1493: remove deprecated completion response parts
trivial, I added `@deprecating` to a few fields in `outgoing.scala` last week. We can remove 'em! And hopefully speed up completions in emacs at the same time. Bonus.

fommil:
close #1493 and uncovered #1494

also renamed all `List()` to `Nil`... it's fun reading the source code of `List` and seeing just how many object allocations that saves us :smile:


soc:
@fommil scalac automatically replaces all `List()`s with Nil.

fommil:
@soc which version of the compiler? And is that true for `Seq()` and friends?

PR #1501, opened 2016-06-04
source resolver should only react to sources

### LINKED ISSUE #1488: constantly updating sources
I've noticed that if the source watcher is enabled the log spins on `updating sources`. I noticed this when using it with an sbt project using Java 7.

I wonder if the source watcher code should just be removed and made a client responsibility... maybe we make it false by default.

fommil:
close #1488

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/SourceResolver.scala
+++ b/core/src/main/scala/org/ensime/indexer/SourceResolver.scala
@@ -3,11 +3,13 @@
 package org.ensime.indexer
 
 import akka.event.slf4j.SLF4JLogging
+import java.io.File
 import org.apache.commons.vfs2._
 
 import org.ensime.api._
 import org.ensime.vfs._
 
+import org.ensime.util.file._
 import org.ensime.util.list._
 import org.ensime.util.map._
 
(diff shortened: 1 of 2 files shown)


fommil:
I ported the presentation compiler crash fix too.

core/src/main/scala/org/ensime/indexer/SourceResolver.scala
```diff
@@ -19,12 +21,15 @@ class SourceResolver(
...
-  // it's not worth doing incremental updates - this is cheap
-  // (but it would be nice to have a "debounce" throttler)
-  def fileAdded(f: FileObject) = update()
-  def fileRemoved(f: FileObject) = update()
+  def fileAdded(f: FileObject) = if (relevant(f)) update()
+  def fileRemoved(f: FileObject) = if (relevant(f)) update()
   def fileChanged(f: FileObject) = {}
 
+  def relevant(f: FileObject): Boolean = f.getName.isFile && {
+    val file = new File(f.getName.getURI)
```
rorygraves: Could we avoid the new File creation here - not sure how often this is called/whether it hurts performance at all - but thought I would mention it.

rorygraves: @fommil  says 'It's only called for changed files so it's not a churn hotspot'
LGTM

rorygraves:
LGTM

PR #1512, opened 2016-06-11
remove testutils from the parent

### LINKED ISSUE #1503: don't publish `testutil`
maybe we had plans to publish `testutil` at some point for code reuse but all it's doing is increasing the size of our jar.

fommil:
trivial close #1503


fommil:
LGTM trivial

PR #1514, opened 2016-06-13
batch queries in Slick to avoid OOM.

### LINKED ISSUE #1489: Stack overflow in Slick when indexing huge jars
### Steps to reproduce

1. Boot up ensime with a big (200mb) jar in compile deps.

### Expected behaviour

Doesn't crash : )

### Actual behaviour

Indexer fails with:

```
Exception in thread "Slick-5" java.lang.StackOverflowError
	at java.util.HashMap.hash(HashMap.java:362)
	at java.util.HashMap.getEntry(HashMap.java:462)
	at java.util.HashMap.containsKey(HashMap.java:449)
	at java.util.HashSet.contains(HashSet.java:201)
	at com.zaxxer.hikari.proxy.ConnectionProxy.checkException(ConnectionProxy.java:110)
	at com.zaxxer.hikari.proxy.ConnectionProxy.checkException(ConnectionProxy.java:117)
	at com.zaxxer.hikari.proxy.ConnectionProxy.checkException(ConnectionProxy.java:117)
        (above line repeated a bazillion times)
```

fommil:
close #1489

@aemoncannon can you please try this branch out at work on that mega jar?

@smootoo any chance of a quick look? Don't laugh.

I picked 10,000 arbitrarily. I'll need to check that I haven't introduced a performance regression at my own work.

Thanks to @raulraja for the suggestion of just mapping over the batches.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -120,9 +120,9 @@ class SearchService(
       }
     }.toSet ++ config.javaLibs.map(vfs.vfile)
 
-    def indexBase(base: FileObject, fileCheck: Option[FileCheck]): Future[Option[Int]] = {
+    def indexBase(base: FileObject, fileCheck: Option[FileCheck]): Future[Int] = {
       val outOfDate = fileCheck.map(_.changed).getOrElse(true)
-      if (!outOfDate) Future.successful(None)
+      if (!outOfDate) Future.successful(0)
       else {
         val boost = isUserFile(base.getName())
         val check = FileCheck(base)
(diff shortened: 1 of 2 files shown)


fommil:
LGTM, merging so we can get a build. @aemoncannon do you think you can test this in the next few days?

smootoo:
@fommil Sorry Sam. I missed this under a blizzard of emails I had when you pinged me.
Change itself likes fine. Won't break anything for sure. Would be interesting to see if it does fix the problem.
Batch inserts can be problematic though and can be driver specific in their behavior. https://github.com/slick/slick/issues/1272

fommil:
@smootoo very interesting, is there a simple way to check if we're getting bulk inserts with H2? I suspect we are, or it'd be insanely slow.

smootoo:
@fommil the only way I know for sure is performance. Maybe take your 10,000 here down to 1 and see what happens. Much slower means we are getting batching.

PR #1520, opened 2016-06-17
Add support for by-name parameters in ArrowTypeInfoBuilder.

### LINKED ISSUE #1477: TypeToScalaNameSpec should report ArrowType for by name parameters
1. add a test to `TypeToScalaNameSpec` for a by name method parameter
2. assert that it is an arrow type. By guess is that it reports as a `BasicType` right now

iravid:
This resolves #1477.


fommil:
can you raise this one against 2.0 please? There is special client side support for by-name methods that might break.

PR #1521, opened 2016-06-17
Add by-name parameter support to ArrowTypeInfoBuilder

### LINKED ISSUE #1477: TypeToScalaNameSpec should report ArrowType for by name parameters

iravid:
This resolves #1477


fommil:
This has broken the emacs integration tests.

It probably requires a small fix on the emacs side. Are you up to fixing that? grep the code for `<byname>` and you'll find it real easily.

iravid:
Yeah, figured that was the issue :-) Looking into it.

fommil:
remember the client has to support both protocols, so don't just change it... add support for both ways of writing a byname.

iravid:
Oh. Wasn't aware of that backwards compatibility requirement.

fommil:
I kicked off a build

fommil:
LGTM, what's next on your hit list?

fommil:
LGTM

PR #1522, opened 2016-06-18
Handling generic vars and arrays in class signature parameters

### LINKED ISSUE #1515: SignatureParser fails when type parameter's bound is also a type parameter
```
public class Java8Sig<U extends A, A> { }
```
Signature for this class is `<U:TA;A:Ljava/lang/Object;>Ljava/lang/Object;` which is currently not supported by `SignatureParser`.
This breaks java support in 2.0 on Java 8.

asznajder:
Fixes #1515


fommil:
LGTM

PR #1527, opened 2016-06-20
Fix debugger issues

### LINKED ISSUE #1505: Not all scenarios covered in object caching
As discussed with @dickwall, I have missed a couple of locations where object references should be cached to allow correct value inspection.

Specifically, the following places need caching:
- https://github.com/ensime/ensime-server/blob/1.0/core/src/main/scala/org/ensime/core/debug/DebugActor.scala#L334

    ```scala
    val result = s.tryThread(threadId).map(_.cache()).flatMap(t =>
          suspendAndExecute(t, action(s, t)))
    ```

- https://github.com/ensime/ensime-server/blob/1.0/core/src/main/scala/org/ensime/core/debug/DebugActor.scala#L249

    ```scala
    s.tryThread(threadId.id).map(_.cache()) match { // ...
    ```

Furthermore, in the case of accessing a thread that has not been cached when issued a stack slot request, we should search for the thread if not found in the cache:

https://github.com/ensime/ensime-server/blob/1.0/core/src/main/scala/org/ensime/core/debug/DebugActor.scala#L391

```scala
objectCache.load(threadId.id)
  .orElse(scalaVirtualMachine.tryThread(threadId.id).toOption).flatMap { // ...
```

The above requires making the scala virtual machine instance available to the `lookupValue` function.

There may also be additional places that I missed with the caching invocation.

Furthermore, the cache method currently returns an object info wrapper. This means certain code will require a hard cast (such as to a thread info, etc). Debugger api 1.1.0-M2 should include a fix to return the actual type instead of the base type of object.

### LINKED ISSUE #1506: Support Scala-specific field names
Not sure if I'd really call this a bug, but @dickwall has already encountered a Scala-specific debugging issue.

```scala
package com.example

object Hello {
  def main(args: Array[String]): Unit = {
    val h = new Hello("Rory", 5)
    h.sayHello()
  }
}

class Hello(name: String, times: Int) {
  def sayHello(): Unit = {
    val actualTimes = times * 2

    for (i <- 1 to actualTimes) // Breakpoint on this line
      println(s"Hello $name")
  }
}
```

In the above example, when reaching a breakpoint on the line `for (i <- 1 to actualTimes)`, the following variables are listed: `actualTimes` as the only local variable, `times` as a field, and `com$example$Hello$$name` as another field. The issue is that it appears that object fields have an odd name when using the JDI of `<class>$$<field name>`. This results in lookup via `DebugObjectField(threadId, name)` failing to find the field.

To remedy this, it looks like it is time to add the first Scala-specific profile to the debugger api. All this profile needs to do is inherit from the pure profile and override the field info method that returns the name of the field to chop off everything but the actual name of the field (after the `$$`). I need to open a debugger API issue to link to this.

On a side note, there is also an issue that accessing a field containing a primitive value doesn't seem to be possible with the current ensime protocol messages. The only field-related message is `DebugObjectField`, which contains an object id and a field name. This means primitive fields do not have a location to send back upon request (returns `None` [here](https://github.com/ensime/ensime-server/blob/1.0/core/src/main/scala/org/ensime/core/debug/DebugActor.scala#L210)).

Depends on https://github.com/ensime/scala-debugger/issues/257 to be completed.

### LINKED ISSUE #1508: RequestLocateNameReq for fields sends incorrect result
Not sure how I let this one get passed me. The following is a snippet from the `DebugLocateNameReq`:

```scala
            tr.get.flatMap {
              case v: IndexedVariableInfoProfile =>
                println(s"Found local variable for $name")
                Some(DebugStackSlot(DebugThreadId(t.cache().uniqueId), v.frameIndex, v.offsetIndex))
              case v if v.isField => v.toValueInfo match {
                case o if o.isObject =>
                  println(s"Found object field for $name")
                  val oo = o.toObjectInfo
                  Some(DebugObjectField(DebugObjectId(oo.cache().uniqueId), v.name))
                case _ =>
                  println(s"Found primitive field for $name")
                  None
              }
```

The field logic is entirely wrong. What it should be returning is `Some(DebugObjectField(<THIS OBJECT CONTAINING FIELD>, v.name))`. Instead, it is returning the field if it is an object and `None` otherwise. Instead, the field needs to have a reference to its containing parent (it already does internally) and access that.

This is going to require a change to the debugger api to have a standard method to return the parent object for the field (and will require a specific trait to expose it). Not hard, but going to have to publish an M2 once this (and any other required changes for the 1.0 release) are finished.

chipsenkbeil:
Resolves #1505, resolves #1506, and resolves #1508.

Tentative PR based on fixes made from @dickwall's and my own discoveries. Only change remaining is to publish `1.1.0-M2` of the debugger API and switch this branch to use that version. The snapshot is the latest master code; so, if everything in the debugger api checks out, I can publish M2 shortly.

- Fixed field location retrieval to use parent (no static field support)

- Added missing caching calls

- Fixed array bug (pattern matching was not appropriate)

- Fixed potential thread bug (same pattern matching problem as array)

- Updated dummy vm to use Scala profile

- Updated start of new VMs to use Scala profile

### DIFF
--- a/core/src/it/scala/org/ensime/intg/DebugTest.scala
+++ b/core/src/it/scala/org/ensime/intg/DebugTest.scala
@@ -240,7 +240,7 @@ class DebugTest extends EnsimeSpec
                 summary should startWith("Instance of scala.collection.immutable.$colon$colon")
                 exactly(1, debugFields) should matchPattern {
                   case DebugClassField(_, head, "java.lang.Object", summary) if (
-                    (head == "head" || head == "scala$collection$immutable$$colon$colon$$hd") &&
+                    (head == "head" || head == "hd") &&
                     summary.startsWith("Instance of java.lang.Integer")
                   ) =>
                 }
--- a/core/src/it/scala/org/ensime/intg/DebugTest.scala
+++ b/core/src/it/scala/org/ensime/intg/DebugTest.scala
@@ -306,6 +306,33 @@ class DebugTest extends EnsimeSpec
     }
   }
 
+  they should "be retrievable for the BugFromGitter scenario" taggedAs Debugger in withEnsimeConfig { implicit config =>
+    withTestKit { implicit testkit =>
+      withProject { (project, asyncHelper) =>
+        implicit val p = (project, asyncHelper)
+        withDebugSession(
+          "variables.BugFromGitter",
+          "variables/BugFromGitter.scala",
+          31
+        ) { (threadId, variablesFile) =>
+            getVariableAsString(threadId, "actualTimes").text should be("10")
+
+            getVariableAsString(threadId, "name").text should be("\"Rory\"")
+
+            getVariableAsString(threadId, "times").text should be("5")
+
+            getVariableAsString(threadId, "arrayTest").text should be("Array(length = 4)[1,2,3,...]")
+
+            getVariableAsString(threadId, "listTest").text should
+              startWith("Instance of scala.collection.immutable.$colon$colon")
+
+            getVariableAsString(threadId, "person").text should
+              startWith("Instance of variables.Person")
+          }
+      }
+    }
+  }
+
   they should "set variable values" taggedAs Debugger in withEnsimeConfig { implicit config =>
     withTestKit { implicit testkit =>
       withProject { (project, asyncHelper) =>
--- a/core/src/main/scala/org/ensime/core/debug/DebugActor.scala
+++ b/core/src/main/scala/org/ensime/core/debug/DebugActor.scala
@@ -194,21 +194,25 @@ class DebugActor private (
       sender ! withThread(threadId.id, {
         case (s, t) =>
           if (name == "this") {
-            t.tryTopFrame.flatMap(_.tryThisObject).map {
-              case objectReference =>
-                DebugObjectReference(objectReference.cache().uniqueId)
+            t.tryTopFrame.flatMap(_.tryThisObject).map(_.cache()).map {
+              case objRef => DebugObjectReference(objRef.uniqueId)
             }.getOrElse(FalseResponse)
           } else {
-            t.findVariableByName(name).flatMap {
+            val result = t.findVariableByName(name).map(_.cache())
+            result.flatMap {
               case v: IndexedVariableInfoProfile =>
-                Some(DebugStackSlot(DebugThreadId(t.cache().uniqueId), v.frameIndex, v.offsetIndex))
-              case v if v.isField => v.toValueInfo match {
-                case o if o.isObject =>
-                  val oo = o.toObjectInfo
-                  Some(DebugObjectField(DebugObjectId(oo.cache().uniqueId), v.name))
-                case _ =>
+                Some(DebugStackSlot(DebugThreadId(t.uniqueId), v.frameIndex, v.offsetIndex))
+              case f: FieldVariableInfoProfile => f.parent match {
+                case Left(o) =>
+                  o.cache()
+                  Some(DebugObjectField(DebugObjectId(o.uniqueId), f.name))
+                case Right(r) =>
+                  log.error(s"Field $name is unable to be located because it is static!")
                   None
               }
+              case x =>
+                log.error(s"Unknown value: $x")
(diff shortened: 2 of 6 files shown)


chipsenkbeil:
Oh, are snapshot dependencies disabled now? >_>

chipsenkbeil:
That doesn't seem to be the case. Looks like something just failed to download as part of the snapshot.

rorygraves:
My local build agrees with CI

chipsenkbeil:
Hm, don't want what I've done incorrectly for the snapshot publish. I just ran `sbt +publishSigned` like I always do.

rorygraves:
I'm not sure if any flags were changed in the build as part of the 1.0 process - I'm having a poke around

rorygraves:
Maybe the removal of this line:
```  
resolvers += Resolver.sonatypeRepo("snapshots") 
```
?

chipsenkbeil:
Hm, drone failed for Scala 2.11 on inspecting variables, but it doesn't fail locally with 2.11 via `sbt ++2.11.8 "core/it:testOnly org.ensime.intg.DebugTest"`. Not sure if that's a rare failure due to some weirdness or if there is something different about the drone setup.

chipsenkbeil:
Nope, seems consistent. Must have broken something.

chipsenkbeil:
Windows test failed with the following:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  EXCEPTION_ACCESS_VIOLATION (0xc0000005) at pc=0x00007ffcb2032045, pid=2944, tid=4092
#
# JRE version: Java(TM) SE Runtime Environment (7.0_79-b15) (build 1.7.0_79-b15)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (24.79-b02 mixed mode windows-amd64 compressed oops)
# Problematic frame:
# C  [ntdll.dll+0x32045] 
#
# Core dump written. Default location: C:\projects\ensime-server-circ6\core\hs_err_pid2944.mdmp
#
# An error report file with more information is saved as:
# C:\projects\ensime-server-circ6\core\hs_err_pid2944.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

That's a new one. Not sure what I've done to cause that.

rorygraves:
@chipsenkbeil  I have just seen the windows failure on a different PR - the drone 2.10 failure looks valid.

chipsenkbeil:
Oh, wait, was it 2.10 that failed last time? Oh, yeah, it was. That explains why I couldn't see the error when running on 2.11. Hopefully I can reproduce it locally with 2.10.

chipsenkbeil:
Nope, doesn't fail locally via `sbt ++2.10.6 "core/it:testOnly org.ensime.intg.DebugTest"`. Argh.

chipsenkbeil:
Think I figured it out. There was an integration test that was checking for an old `package$class$$fieldName`, which is now just `fieldName` using the Scala profile. Why I wasn't running into the problem on my local box and Macs, I couldn't say.

fommil:
LGTM of course we'll need the final release before we can cut 1.0, but I'll be keen to hear with @dickwall has to say

fommil:
you need to rebase.

chipsenkbeil:
Went ahead and rebased and published an M2 release. If for some reason there's an issue with the debugger api, I'm fine releasing an M3. But, if everything checks out, this will be ready to go. Do want @dickwall's approval if he gets the chance, of course.

fommil:
LGTM

fommil:
great work, thanks! Yes it would be good to get confirmation from Dick.

Turns out there is one other last minute bug that affected me at work in the indexer, should be a simple fix, but that is now blocking 1.0. Almost there...

fommil:
please raise against 2.0

chipsenkbeil:
@fommil, did you mean to merge this in without hearing back from @dickwall?

fommil:
yeah I want him to check it on the release version.

Issue #1366, opened 2016-03-30
Create requests / responses classes just for Java usage


asznajder:
`receive` method in `JavaAnalyzer` dispatches the requests from user depending their type. Create a sealed family of requests / responses just for Java usage. Currently for Java we use the same requests / responses classes as for Scala.

rorygraves:
https://github.com/ensime/ensime-server/blob/master/core/src/main/scala/org/ensime/core/JavaAnalyzer.scala#L53

@aemoncannon  Not quite sure what you are after here, care to clarify - do you want to seperate out all of the Scala and Java calss?

fommil:
I think that was me. It was a suggestion that we eventually have a dedicated protocol for Java. This is huge, it's not suitable for a one evening hack.

PR #1529, opened 2016-06-21
symbol at point is now consistent for java methods

mwielocha:
The main update here is that now all 1st level type infos are of `ArrowTypeInfo` which fulfills #1494.
The extra stuff is around java method naming. I went with full human readable signatures, but honestly I think this should be a part of a wider discussion.
Basically what we have now is:
```scala
s"$modifiers $returns $name$params"
```
where short version has type with simple names and full version has type names with packages:
```java
private static int compute(int a, int b)
// and
public java.lang.String getName(int index)
// etc
```
All three endpoints are updates: `askSymbolAtPoint`, `askTypeAtPoint` and `askCompletionsAtPoint`.
Feel free to comment on the naming format.


core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
```diff
@@ -113,7 +113,7 @@ class JavaCompilerSpec extends EnsimeSpec
           case "3" =>
             info.name shouldBe "java.io.PrintStream.println(java.lang.Object)"
             info.localName shouldBe "println"
-            info.`type`.name shouldBe "(java.lang.Object)void"
+            info.`type`.name shouldBe "public void println(Object arg0)"
```
fommil: this is more like the symbol at point rather than the type at point. The type wouldn't have the `println` name in it.

Seems we need a symbol-at-point and a type-at-point distinction in java.

fommil: now that you have the richer object it would be better to assert on the entire thing and not just the `name`... that way you can assert on the return type and named parameters.

util/src/main/scala/org/ensime/util/Show.scala
```diff
@@ -0,0 +1,18 @@
+package org.ensime.util
+
+trait Show[T] {
```
fommil: if we're going to introduce a `Show` typeclass we should at least use the one (or its pattern) from `cats` instead of creating our own.

fommil: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Show.scala

mwielocha: Including a whole library for one type class seemed a bit extreme... plus you said something about not relying on scala libs?

mwielocha: But if thats not an issue then yeah, I also vote for cats! 🐱

fommil: hehe, yeah maybe it's a bit much just for one typeclass (especially something so trivial). Can we perhaps do what we want without the `Show` typeclass? If we're going to use `Show`, we should be consistent and use it everywhere.

mwielocha: Thats why it went into utils, but sure it can do without it, I'll simplify this.

core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
mwielocha: Ok, I was kinda trying to avoid distinctions, but if its ok then even better.

mwielocha: Btw, `JavaCompiler` will need a refactor soon, its getting nasty in there, perhaps somewhere around  #1366?

fommil: hehe, yeah sounds good!

mwielocha:
Actually I complicated things a bit. We now have a choose by import pattern to determine which name formatting to use for `atPoint` methods. Also `JavaCompilerSpec` now asserts on `ArrowTypeInfo`. The naming might still use some adjusting thou, @fommil now sure what you mean by removing the return type, this seems super-useful and emacs handles it correctly...

mwielocha:
Btw, how do I assign myself to a task?

fommil:
@mwielocha search for "assign yourself" on this page (it's on the right below Labels, Milestone and Assignees)

mwielocha:
Nope, nothing like that :(

fommil:
oh, maybe you didn't accept the invite to join the organisation?

mwielocha:
Thats probably it. Btw, the tests are not passing probably due to a different java version am planning to fix this asap.

fommil:
use `jenv` and you'll always use the same (major) version as our CI.

mwielocha:
Seems that the problem is much more complicated since `Trees.getElement` behaves differently in java 1.7. So I will have to go through this again unfortunately. For now what I see is that for a given path over a method `getElement` returns null where in 1.8 it was returning a valid `ExecutableElement`. Anyway @fommil you might consider removing the low hanging fruit from this one 😄

fommil:
this is the story of everything with low-hanging fruit :smile:

mwielocha:
Not again.

mwielocha:
Ok, so the solution for java 1.7 was to use  sub-interfaces of `TypeMirror`, `ExecutableType` to be exact. We lost parameter names with this approach for `symbolAt` and `typeAt` but I don't think we really need them. Anyway, it was impossible to make it abstract, so its not pretty, but it does what it suppose to. Still not sure about the naming thou.

fommil:
Sorry I've been slow to respond. Been depressed by British politics. I'll try to review today.

mwielocha:
No worries, I know how you feel, its quite depressing for the whole Europe imho 😞

fommil:
LGTM thanks! The TypeInfoFormat looks unused can you please follow up with a PR to remove it? (And also delete the commented code)

Issue #1526, opened 2016-06-20
could not read file: jar:file


fommil:
looks like we have this bug

http://groovyjavathoughts.blogspot.co.uk/2015/01/why-am-i-getting-warning-arguments-left.html

https://github.com/ensime/ensime-server/blob/2.0/core/src/main/scala/org/ensime/indexer/SearchService.scala#L337

which means there is probably another bug hiding under this.

rorygraves:
Looking at this.  For akka logging the correct form is:

```
log.warning(t, "msg") 
```
rather than
```
log.warning("msg", t)
```
Of course for direct use of SLF its the other way round (not a var-args)

fommil:
ok, the problem happens when a jar file is deleted and recreated. The file watcher is too fast or the OS doesn't write the file fast enough or something. I guess the only way we can recover from this is by having a debounce to give the OS some time to write out the file.

fommil:
more importantly, this happens when the jar is deleted and replaced within a very short timeframe.

fommil:
```
org.apache.commons.vfs2.FileSystemException: Could not read file "jar:file:///C:/FOO/scala-2.11/last/BAR.jar!/foo/Bar.class".
	at org.apache.commons.vfs2.provider.AbstractFileObject.getInputStream(AbstractFileObject.java:1317)
	at org.apache.commons.vfs2.provider.DefaultFileContent.getInputStream(DefaultFileContent.java:411)
	at org.ensime.indexer.ClassfileIndexer$class.indexClassfile(ClassfileIndexer.scala:24)
	at org.ensime.indexer.SearchService.indexClassfile(SearchService.scala:28)
	at org.ensime.indexer.SearchService.org$ensime$indexer$SearchService$$extractSymbols(SearchService.scala:212)
	at org.ensime.indexer.SearchService$$anonfun$extractSymbolsFromClassOrJar$1$$anonfun$apply$7$$anonfun$apply$8.apply(SearchService.scala:196)
	at org.ensime.indexer.SearchService$$anonfun$extractSymbolsFromClassOrJar$1$$anonfun$apply$7$$anonfun$apply$8.apply(SearchService.scala:196)
	at scala.collection.immutable.List.flatMap(List.scala:327)
	at org.ensime.indexer.SearchService$$anonfun$extractSymbolsFromClassOrJar$1$$anonfun$apply$7.apply(SearchService.scala:196)
[...]

PR #1531, opened 2016-06-21
plug a semaphore leak

fommil:
address some parts of #1526 but not fix it because I can't reproduce it as a test.


fommil:
LGTM, small change, if it goes green we should merge.

fommil:
interesting. Test failure. I wasn't expecting that.

fommil:
oh I see it.

fommil:
LGTM and merge if it goes green

Issue #761, opened 2015-01-03
don't unpack jars/zip files


fommil:
currently we unpack files from archives and then tell clients about the unpacked version.

Two steps:
1. update emacs client to handle read-only loads of archived files (I'm sure this is supported)
2. update server to send URLs to the compressed files

edani:
Third step: update emacs client (and protocol) to send URLs of compressed files when doing symbol lookups / sem highlight / etc.

aemoncannon:
For reference, here's how cider-mode opens paths within archives 
https://github.com/clojure-emacs/cider/blob/e871b4c669bc887e2cdfec4445a0115c602f8d0d/cider-interaction.el#L536

aemoncannon:
Interesting, looks like Emacs virtualizes file interactions with file-name-handler-alist: http://www.gnu.org/software/emacs/manual/html_node/elisp/Magic-File-Names.html

aemoncannon:
We would need to agree on a file name format for files-within-jars. Looks like cider does:
^(jar|zip):file:(.+)!/(.+)
Where I guess '!/' is the separator between jar path and compressed file path.

fommil:
if it agrees with the Apache VFS naming conventions, then we're sorted! https://github.com/ensime/ensime-server/blob/master/src/main/scala/org/ensime/indexer/package.scala#L15

aemoncannon:
So... why do we want to do this? : )  It occurs to me that it's kinda nice to keep the client ignorant of archives. I mean, what's the down side?

fommil:
we never clean up after ourselves when we unextract, plus we have to do the file handling on the server side which is work that we'd rather not have to do.

aemoncannon:
> clean up
Agree it's annoying.

> file handling
Do you mean the unpacking itself? It looks like we do it with minimal
effort (loc-wise). Even the elisp to support files in jars will probably be
longer.

I'm wondering if this is a place where we can take a little hit in
complexity to make clients live's easier.

fommil:
yeah, low priority

fommil:
I got stung by this today. Was looking in Gen.scala of scalacheck, then upgraded scalacheck and did a jump to source and was really confused why I could see methods that the compiler refused to see.

As a workaround, we should perhaps delete the dep-src directory on startup.

fommil:
dag namnit, this caught me today when I was swapping versions of a backend library and I couldn't understand why my use of the (changing) API wasn't working.

[...]

PR #1533, opened 2016-06-23
WIP:Changed SourceFileInfo to work with Path instead of file: #761

hardmettle:
Modified Source file info to user Path instead of File. This will fix the unpacking of archives for the client #761

### DIFF
--- a/api/src/main/scala/org/ensime/api/common.scala
+++ b/api/src/main/scala/org/ensime/api/common.scala
@@ -3,6 +3,8 @@
 package org.ensime.api
 
 import java.io.File
+import java.nio.file.Path
+
 import scala.annotation.StaticAnnotation
 
 /**
@@ -91,10 +93,10 @@ object RefactorType {
 }
 
 final case class SourceFileInfo(
-    file: File,
+    path: Path,
     contents: Option[String] = None,
     contentsIn: Option[File] = None
 ) {
   // keep the log file sane for unsaved files
-  override def toString = s"SourceFileInfo($file,${contents.map(_ => "...")},$contentsIn)"
+  override def toString = s"SourceFileInfo($path ,${contents.map(_ => "...")},$contentsIn)"
 }
--- a/api/src/main/scala/org/ensime/api/common.scala
+++ b/api/src/main/scala/org/ensime/api/common.scala
@@ -93,10 +93,20 @@ object RefactorType {
 }
 
 final case class SourceFileInfo(
-    path: Path,
+    file: Path,
     contents: Option[String] = None,
     contentsIn: Option[File] = None
 ) {
   // keep the log file sane for unsaved files
-  override def toString = s"SourceFileInfo($path ,${contents.map(_ => "...")},$contentsIn)"
+  override def toString = s"SourceFileInfo($file ,${contents.map(_ => "...")},$contentsIn)"
 }
+
+sealed trait EnsimeFile
+final case class RawFile(file: Path) extends EnsimeFile
+
+/**
+ * Represents a path with archive location and the entry being reffered to
+ * @param archive the container of entry
+ * @param entry is absolute
+ */
+final case class ArchiveFile(archive: Path, entry: Path) extends EnsimeFile
--- a/api/src/main/scala/org/ensime/api/common.scala
+++ b/api/src/main/scala/org/ensime/api/common.scala
@@ -3,7 +3,7 @@
 package org.ensime.api
 
 import java.io.File
-import java.nio.file.Path
+import java.nio.file.{ Path, Paths }
 
 import scala.annotation.StaticAnnotation
 
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -3,6 +3,7 @@
 package org.ensime.api
 
 import java.io.File
+import java.nio.file.Path
 
 /**
  * There should be exactly one `RpcResponseEnvelope` in response to an
@@ -98,7 +99,7 @@ final case class NewJavaNotesEvent(
 final case class DebugStepEvent(
   threadId: DebugThreadId,
   threadName: String,
-  file: File,
+  path: Path,
   line: Int
 ) extends DebugEvent
 
@@ -106,7 +107,7 @@ final case class DebugStepEvent(
 final case class DebugBreakEvent(
   threadId: DebugThreadId,
   threadName: String,
-  file: File,
+  path: Path,
   line: Int
 ) extends DebugEvent
 
@@ -121,7 +122,7 @@ final case class DebugExceptionEvent(
   exception: Long,
   threadId: DebugThreadId,
   threadName: String,
-  file: Option[File],
+  path: Option[Path],
   line: Option[Int]
 ) extends DebugEvent
 
@@ -234,7 +235,7 @@ case object PosNeededYes extends PosNeeded
 sealed trait SourcePosition extends RpcResponse
 final case class EmptySourcePosition() extends SourcePosition
 final case class OffsetSourcePosition(file: File, offset: Int) extends SourcePosition
-final case class LineSourcePosition(file: File, line: Int) extends SourcePosition
+final case class LineSourcePosition(path: Path, line: Int) extends SourcePosition
 
 final case class PackageInfo(
     name: String,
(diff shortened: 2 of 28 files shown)


fommil:
Sorry I've been slow to respond. Been depressed by British politics. I'll try to review today.

hardmettle:
It's absolutely fine.  Please take your own time to review. I can understand the way political situation can let us down.

fommil:
@hardmettle just when I was feeling like I could emerge from a cave, post Brexit, I have come down with a cold so I won't be able to help out for a few more days :sob:

hardmettle:
@fommil no problems. Get well soon .. And I will manage myself(upto some extent)

fommil:
looks like a rebase to 2.0 is needed in any case.

api/src/main/scala/org/ensime/api/common.scala
```diff
@@ -91,10 +93,10 @@ object RefactorType {
 }
 
 final case class SourceFileInfo(
-    file: File,
+    path: Path,
```
fommil: this is going to involve quite a lot of changes to the clients. Maybe keep the field name as `file` to reduce the impact.

api/src/main/scala/org/ensime/api/outgoing.scala
```diff
@@ -98,15 +99,15 @@ final case class NewJavaNotesEvent(
 final case class DebugStepEvent(
   threadId: DebugThreadId,
   threadName: String,
-  file: File,
+  path: Path,
```
fommil: ditto

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -219,35 +220,35 @@ trait RichCompilerControl extends CompilerControl with RefactoringControl with C
...
-    createSourceFile(SourceFileInfo(file, None, None))
+    createSourceFile(SourceFileInfo(file.toPath, None, None))
   def createSourceFile(path: String): BatchSourceFile =
     createSourceFile(new File(path))
   def createSourceFile(file: AbstractFile): BatchSourceFile =
     createSourceFile(file.file)
   def createSourceFile(file: SourceFileInfo): BatchSourceFile = file match {
     case SourceFileInfo(f, None, None) =>
       new BatchSourceFile(
-        new PlainFile(f.getPath),
```
fommil: we'd still want to create `PlainFile` if the `Path` points to a real file, not a path within an archive.

core/src/main/scala/org/ensime/core/debug/SourceMap.scala
```diff
@@ -55,10 +56,14 @@ class SourceMap(
    * @param location The location whose source file to find
    * @return Some file representing the local source, otherwise None
    */
-  def findFileByLocation(location: LocationInfoProfile): Option[File] = {
-    val path = location.trySourcePath.toOption
-
-    path.flatMap(sourceForFilePath)
+  def findFileByLocation(location: LocationInfoProfile): Option[Path] = {
+    location.trySourcePath.toOption match {
+      case Some(f) => Some(new File(f).toPath)
```
fommil: there is a way to create `Path`s from `String` without having to go via `File`

core/src/main/scala/org/ensime/core/debug/StructureConverter.scala
```diff
@@ -167,7 +167,7 @@ class StructureConverter(private val sourceMap: SourceMap) {
     import org.ensime.util.file._
     val pcLocation = sourceMap.newLineSourcePosition(frame.location).getOrElse(
       LineSourcePosition(
-        File(frame.location.sourcePath).canon,
+        File(frame.location.sourcePath).canon.toPath,
```
fommil: we might want to avoid going via `File` here also, but that means we need a way of canonicalising `Path`, not just `File`

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -69,7 +71,7 @@ class JavaCompiler(
 
   def internSource(sf: SourceFileInfo): JavaFileObject = {
     val jfo = getJavaFileObject(sf)
-    workingSet.put(sf.file.getAbsolutePath, jfo)
+    workingSet.put(generateMd5(sf.path.toString), jfo)
```
fommil: like with scala, we'd want to avoid creating virtual files unless we need to .

core/src/main/scala/org/ensime/core/javac/JavaCompletion.scala
```diff
@@ -259,7 +260,7 @@ trait JavaCompletion { this: JavaCompiler =>
   }
 
   private def contentsAsString(sf: SourceFileInfo, charset: Charset) = sf match {
-    case SourceFileInfo(f, None, None) => f.readString()
+    case SourceFileInfo(f, None, None) => f.readString
```
fommil: `readString()` implies a side effect, so we should keep the `()`

util/src/main/scala/org/ensime/util/path.scala
```diff
@@ -0,0 +1,29 @@
+package org.ensime.util
+
+import java.net.{ JarURLConnection, URL }
+import java.nio.charset.Charset
+import java.nio.file.{ Path => JPath }
+
+import com.google.common.io.Files
+/**
+ * Created by harsh on 15/06/16.
```
fommil: we try not to have these sorts of comments. Note that you'll need to run the sbt-header parser to add copyright/license info to the top of the file. See `.drone.yml` for the command.

util/src/main/scala/org/ensime/util/path.scala
```diff
@@ -0,0 +1,29 @@
...
+ */
+package object path {
+  type Path = JPath
+  implicit class EnrichPath(val path: Path) extends AnyVal {
+    def isJarUrl = path.toString.startsWith("jar:file:")
+    def isJava: Boolean = path.toString.endsWith(".java")
+    def isScala: Boolean = path.toString.endsWith(".scala")
+    def readString(implicit cs: Charset): String = {
+      if (!path.isJarUrl) Files.toString(path.toFile, cs) else {
+        val in = new URL(path.toString).openConnection().asInstanceOf[JarURLConnection].getInputStream
```
fommil: I'm surprised that we'd need to do it this way. I thought the `Path` API was able to handle any content. We might need to be careful about closing the jar after we've used it though.

Try to avoid using `scala.io.Source`, it's incredibly badly designed: both from a performance perspective and from a resource management / error handling perspective.

hardmettle:
Apologies for the repeated commits , kind of got caught up in the rebase(with 2.0) conflict again and again and got confused

fommil:
I always find it is easier to do a rebase than a merge. But remember that your changes are called `theirs` in git nomenclature.

core/src/main/scala/org/ensime/core/debug/StructureConverter.scala
hardmettle: You mean canonising both `jar:file:` pattern and the normal path differently ? Wouldn't `canon` do it for us for both the cases ?

hardmettle: @fommil Do you want an implementation of `resolve` here for path ?

util/src/main/scala/org/ensime/util/path.scala
hardmettle: I did a bit of findings and hit and trial and this is how it worked. Also isn't `in.close()` closing the jar stream? I will get rid of `scala.io.Source` though .

fommil:
@hardmettle I'm going to try to have a look at this today. My goal is to:

1. create marshallers for `Path`
2. create an equivalent of `Canon` for `Path`

which are things that I can commit to the 2.0 branch and you can make use of them.


With regards to the `path: String` going via `File`, my concern is that we have paths that are actually `jar:file:/foo/bar.jar!/baz/gaz/daz.scala` and so on, which will just blow up.

On the scala compiler, we should be using real file references (as we currently are) where possibly but creating virtual files when we have a jar reference.

I suspect we need to think about proper code coverage here too.

fommil:
*sigh* I'm not going to get around to doing anything solid on this. The SWANKY protocol stuff is taking much longer than expected.

fommil:
moving branch

PR #1534, opened 2016-06-23
Fix re-indexing of jars

### LINKED ISSUE #1526: could not read file: jar:file

fommil:
close #1526


fommil:
LGTM, for what that good will do us in our saddest hour...

PR #1535, opened 2016-06-23
politely close the jar file system when nuking it

fommil:
forward port #1534


fommil:
LGTM and god help us all

PR #1538, opened 2016-06-27
fix bad merge in EnsimeBuild

sugakandrey:


sugakandrey:
@fommil hm, so this does indeed fix duplicate classpath in `sbt ensime/assembly`, I'm looking into emacs tests right now to make this work.

fommil:
awesome stuff!

Issue #1504, opened 2016-06-06
ArrowTypeInfo should have a list of type parameters


ktonga:
Currently it's not possible to regenerate the list of type params for a generic method based on `typeInfo` content.

fommil:
This is a server bug.

fommil:
https://github.com/ensime/ensime-server/issues/1494

ktonga:
Hi @fommil, I don't think this problems is related to #1494. It's related to the `completions` endpoint not to `symbol at cursor`, and in this case the server is returning `ArrowTypeInfo` which is the proper type for methods.

The problem here is that `ArrowTypeInfo` does not have a field for listing the type params defined for the method.

So for instance, if you have `def genericMethod[A, B](a: A)(f: A => B): B` you have no way to figure out the `[A, B]` part of the signatures from the method `typeInfo`.

Hope it clarifies the problem better.

fommil:
Yes it's definitely that ticket. Happy to talk you through in more detail. Can you have a look at the referenced source files and also some example payloads for Scala (equiv Scala tests will do). All the information you want should be in TypeInfo.

fommil:
(i.e. we either need to add type parameters to arrow types or we need to specialise for java)

fommil:
incidentally, why do you need the type parameter list in particular? Isn't it enough that the types are on the parameters? We've never needed this for scala.

fommil:
oh, you're using "generic" method when you're talking about scala methods. That's typically only reserved for talking about java methods since it's an implementation detail. For scala methods, we'd say "method with type parameters".

What's your usecase? We'd prefer not to increase the size of these payloads unless absolutely necessary, as it can impact performance.

ktonga:
@fommil  Sorry if it sounded confusing, I referred all the time to them as 'type params', didn't realize about the 'generic' word connotation. I'm talking just about Scala.

To see my use case please take a look at [this comment](https://github.com/ensime/ensime-vim/issues/273#issuecomment-224019333) and to the issue in general.

Here you have a couple of comments I put in the gitter channel that better illustrate my problem:

> Hi. I'm trying to improve completions formatting in ensime-vim but I can't find an easy way to regenerate the type params list for a method from the typeInfo. Type params are embedded inside the paramSections, I could find them all and get the unique list but I won't be able to know in which order they were defined. Am I missing something? Thanks.

> I'm thinking in something like def foo\[A, B\](b: B)(f: A => B): B

[...]

Issue #1481, opened 2016-05-31
wrap lucene with an async API


fommil:
currently in `SearchService.scala` all calls to lucene are performing I/O. We should take a leaf out of slick's book and make all calls return a `Future` which is backed by a dedicated I/O thread pool (possibly even the same pool as slick).

asznajder:
I will be happy to work on this, if it's not very urgent :)

PR #1539, opened 2016-06-27
Add `typeParams` to `ArrowTypeInfo` (#1504)

ktonga:
I don't totally agree on some of the coding decisions but I tried to follow the style to not introduce unrelated changes. Also, I know there are a lot of tickets for improving the codebase in general, I hope to be able to help with that!

And obviously, @fommil,  I want my ENSIME sticker(s?) :-)

### DIFF
--- a/api/src/test/scala/org/ensime/api/EnsimeTestData.scala
+++ b/api/src/test/scala/org/ensime/api/EnsimeTestData.scala
@@ -14,7 +14,7 @@ trait EnsimeTestData {
     }
   }
 
-  val typeInfo = new BasicTypeInfo("type1", DeclaredAs.Method, "FOO.type1", List(), List(), None)
+  val typeInfo = BasicTypeInfo("type1", DeclaredAs.Method, "FOO.type1", Nil, Nil, None, Nil)
 
   val interfaceInfo = new InterfaceInfo(typeInfo, Some("DEF"))
   val typeInspectInfo = new TypeInspectInfo(typeInfo, List(interfaceInfo))
@@ -105,7 +105,11 @@ trait EnsimeTestData {
 
   val noteList = NewScalaNotesEvent(isFull = true, List(note1, note2))
 
-  val entityInfo: TypeInfo = new ArrowTypeInfo("Arrow1", "example.Arrow1", typeInfo, List(paramSectionInfo))
+  val entityInfo: TypeInfo = new ArrowTypeInfo("Arrow1", "example.Arrow1", typeInfo, List(paramSectionInfo), Nil)
+
+  val typeParamA = BasicTypeInfo("A", DeclaredAs.Nil, "example.Arrow1.A", Nil, Nil, None, Nil)
+  val typeParamB = BasicTypeInfo("B", DeclaredAs.Nil, "example.Arrow1.B", Nil, Nil, None, Nil)
+  val entityInfoTypeParams: TypeInfo = new ArrowTypeInfo("Arrow1", "example.Arrow1", typeInfo, List(paramSectionInfo), List(typeParamA, typeParamB))
 
   val completionInfo = CompletionInfo(Some(typeInfo), "name", CompletionSignature(List(List(("abc", "def"), ("hij", "lmn"))), "ABC", false), false, 90, Some("BAZ"))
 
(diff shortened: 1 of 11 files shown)


fommil:
Cool. Only a quick look but some comments: use List instead of Iterable. Don't add the companion, we try to keep these files as pure ADT with no code (have to break the rule where the JVM forces us for toString etc).

ktonga:
Ok, I used iterable just to follow the other fields. I can change it.
Ok with the companion, will remove it. But are you ok with having the shortcuts somewhere? it makes no sense to repeat the same four parameters over and over again in creation and pattern matching.

fommil:
Yeah that's fine, if our domain model is repetitive it probably hints that the model is broken.

fommil:
Changing Iterable to List would be nice :-P

ktonga:
It's broken in a sense that they are not truly ADTs 'cause fields are unnecessarily defined in a higher trait, when they could be in the concrete case class which needs the field (it seems like old fashioned OOP habit). I'd like to participate if you are planning on reworking the API/Model.

I've already changed to `List` the field I've introduced. You mean I should change all of them?

ktonga:
Done. I've moved the companion to the model package, near the Builders.

fommil:
yeah I agree that it's not a true ADT. Same is true of the `domain.scala` file in the indexer. There is definitely some work in both packages to move to a true ADT with typeclasses for all the functionality, but I've just always had higher priority things to work on. If you're looking for general clean-ups that you could get involved in, it would be very much welcomed... indeed I'd even consider moving the FQN domain objects into `api` (which would require a change of license but I think everybody will agree).

api/src/test/scala/org/ensime/api/EnsimeTestData.scala
```diff
@@ -14,7 +14,7 @@ trait EnsimeTestData {
     }
   }
 
-  val typeInfo = new BasicTypeInfo("type1", DeclaredAs.Method, "FOO.type1", List(), List(), None)
+  val typeInfo = BasicTypeInfo("type1", DeclaredAs.Method, "FOO.type1", Nil, Nil, None, Nil)
```
fommil: we use `new` in a whole load of places, it's because of the path that the code has taken... it wasn't always as clean an ADT as it is today and used to be combined with the modelbuilder file.

fommil:
LGTM, the failing tests look like transients. Travis has been acting up, not sure what's wrong with drone.

fommil:
you've found some nice code clean-ups here as you've gone along. It's well appreciated.

fommil:
emacs tests failed again, but looks unrelated to your changes. @chipsenkbeil how is your machine looking?

chipsenkbeil:
@fommil, just SSH'd in. Looks fine. I can see something running on it. Compiling fine so far.

ktonga:
Yeap, dunno what's wrong with Drone and Travis. FTR I've run `test` and `it:test` locally and they succeeded, including the new cases I've introduced.

Looking forward to keep contributing. Ensime is an amazing piece of Software. I'll give it a try to replace Akka with FS2 in SearchService, let's see how it goes.

fommil:
Monix is looking like a viable alternative too ;-)

ktonga:
I know, I was following the ticket closely. But my heart is with FS2, it brings such an elegant programming model and they are working hard to improve performance.

ktonga:
You need anything else from me for this PR?

fommil:
I'll have a look over the next few days.

fommil:
I still think these are transient errors. Travis looks like it picked up a bad cache at some point, so I'll clear that out and restart. No idea what's going on with drone.

fommil:
on the SearchService, I think it's blocked by #1481 ... but if you're keen on something streamy then looking into replacing akka-stream websockets with something else would be really good. Although we'd prefer Java libraries instead of anything too fancy, to make it easier to support scala 2.12.

fommil:
Try pushing a commit. I'm starting to think this isn't a transient on drone.

ktonga:
I've squashed. The sbt build in drone finishes ok, it's a timeout in the ensime-emacs tests I think.

ktonga:
It failed again with same error, it can't update Ensime server. I don't know what that means. Not an emacs expert precisely.

chipsenkbeil:
@fommil, looks like this problem has also happened on your machine several days ago; so, I don't think this is a machine-specific problem. Seems to be plaguing the 2.11.8 build.

fommil:
The emacs tests are passing for other prs so it makes me wonder.

fommil:
I'll investigate locally to see if I spot anything obvious

fommil:
btw, you raised the PR against 1.0, not 2.0. Can you please re-raise against the dev branch? 1.0 is in critical bugfixes only mode. This implies that we have a broken CI for the 1.0 branch that I'll need to fix.

Issue #1469, opened 2016-05-28
use one FQN representation


fommil:
our java analyzer code has an FQN class, but it would be a lot cleaner if it reused the `FullyQualifiedName` ADT in our indexer module..

mwielocha:
I'm taking this one, I think its gonna be relatively easy for me after the last one.

mwielocha:
Its closing time!

PR #1540, opened 2016-06-28
Javafqn to fullyqualifiedname

mwielocha:
This removes `JavaFqn` and solves #1469.
Some random thoughts:
`toDocSig` should be in `FullyQualifiedName` and all implementing classes, but I'm not sure about this cause its not clear to me how Scala Analyzer does the same thing so for now its in java helpers.
While on it, I've discovered `UnsafeHelpers` (yeah, better late than never) and that single method there solves the problem of `getElement` returning null in Java < 8. I think we should stick to a single pattern here especially that there is a simple reflection workaround to make this helper safe again.
Also, I have a really strong feeling that all name formatting in `AtPoint` mixins should utilise `FullyQualifiedName` which would simplify the code a quite a lot. Although I think this should go to a separate ticket.

### DIFF
--- a/core/src/main/scala/org/ensime/core/javac/Helpers.scala
+++ b/core/src/main/scala/org/ensime/core/javac/Helpers.scala
@@ -5,15 +5,19 @@ package org.ensime.core.javac
 import akka.event.slf4j.SLF4JLogging
 import com.sun.source.tree._
 import com.sun.source.util.TreePath
-import javax.lang.model.`type`.{ DeclaredType, PrimitiveType, TypeKind, TypeMirror }
+import javax.lang.model.`type`._
 import javax.lang.model.element._
+import javax.lang.model.element.ElementKind._
 import scala.collection.JavaConversions._
 
-import org.ensime.api.deprecating
 import org.ensime.core.{ DocFqn, DocSig }
 import org.ensime.indexer._
 
 trait Helpers extends UnsafeHelpers with SLF4JLogging {
+
+  private implicit class EnhachedElement(e: Element) {
+    def isOf(kinds: ElementKind*): Boolean = kinds.exists(_ == e.getKind)
+  }
 
   def typeMirror(c: Compilation, t: Tree): Option[TypeMirror] = {
     Option(c.trees.getTypeMirror(c.trees.getPath(c.compilationUnit, t)))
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/ensime/core/javac/Helpers.scala
```diff
@@ -3,31 +3,21 @@
...
+trait Helpers extends UnsafeHelpers with SLF4JLogging {
 
-object JavaFqn {
-  def apply(pack: String, tpe: String, fieldOrMethod: Option[String]): JavaFqn = {
-    JavaFqn(
-      if (pack.isEmpty) None else Some(pack),
-      if (tpe.isEmpty) None else Some(tpe),
-      fieldOrMethod
-    )
+  private implicit class EnhachedElement(e: Element) {
```
fommil: `Enhached`? :smile:

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -188,6 +189,7 @@ final case class ArrayDescriptor(fqn: DescriptorType) extends DescriptorType {
     case a: ArrayDescriptor => a.reifier
   }
   def internalString = "[" + fqn.internalString
+  def fqnString = fqn.fqnString
```
fommil: is this true? I'm not sure it is.

core/src/main/scala/org/ensime/core/javac/Helpers.scala
mwielocha: Yeah I just hate those blown up patter matchings. Plus I'm not too good in naming things (I guess nobody is).

core/src/main/scala/org/ensime/indexer/domain.scala
mwielocha: Not sure what you mean, fqn for descriptor doens't really mean anything (I thing) its just for convenience...

core/src/main/scala/org/ensime/core/javac/Helpers.scala
mwielocha: Oh right and its a typo, god...

core/src/main/scala/org/ensime/indexer/domain.scala
fommil: it's probably best we just take the long route then and call `.fqn.fqnString` instead of trying to pretend that the descriptors always have fqns.

fommil:
2.10 looks like it passes but is showing red. I have no idea what's going on there. I'll bounce it.

fommil:
LGTM

fommil:
I've seen this transient a few times on Travis. I think a bad jar got into the caches. I'm flushing them and restarting the build. Merge if it goes green. (I'd do it myself but it is such a joy to do it that I think the honour should be all yours :smile: )

fommil:
Grr travis

Issue #1287, opened 2016-02-10
exact match classnames being trumped by (often noise) inner classes


fommil:
e.g. when I search for `Foo` if there are loads of inner classes, they'll dominate the search results for me.

fommil:
should probably make each `$` have a penalty when lucene indexing, I also thought our search query had a boost for an exact match. I'll need to check that.

rorygraves:
Notes:  This will be in the SearchService class (the lucene indexer).  It is possible to create an integration test expressing the behaviour.

caoilte:
You would think https://github.com/ensime/ensime-server/commit/df0eb4a7ef9af07d14d79370dea9993574fa76a4 would demonstrate that this cannot happen, but there must be something coincidental about that naming which doesn't lead to the inner class coming top.

I was able to reproduce the exact problem with (and can be called in SearchServiceSpec conveniently)

`service.searchClasses("Matchers", 100)`

which returns the following with `org.scalatest.Matchers` right in the middle,

[...]

PR #1542, opened 2016-06-29
Penalise inner classes in lucene

sugakandrey:
Addresses #1287.  Every non-trailing `$` in fqn is penalised by 0.25 reduction in score weight down to 0. This means after 4 non-trailing `$` the order of documents is not well defined but I don't think we care about anything with 4 `$`'s, we might as well not index them at all.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/IndexService.scala
+++ b/core/src/main/scala/org/ensime/indexer/IndexService.scala
@@ -105,7 +105,7 @@ class IndexService(path: Path) {
     if (boost) {
       fqns foreach { fqn =>
         val currentBoost = fqn.boost("fqn")
-        fqn.boostText("fqn", currentBoost + 1f)
+        fqn.boostText("fqn", currentBoost + .5f)
       }
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/ensime/indexer/IndexService.scala
```diff
@@ -30,7 +30,7 @@ object IndexService extends SLF4JLogging {
       def spacey(s: String) = List(s, s.replace('.', ' '))
 
       val shortPkg = path.map(_.charAt(0)).mkString("", ".", "." + name)
-      val innerNames = name.split('$').filter(_.size > 3).flatMap(cases)
+      val innerNames = name.split('$').flatMap(cases)
```
sugakandrey: This is what was causing problems with @ChrisCoffey PR, I'm not entirely sure why `filter(_.size > 3)` is ever needed here. Perhaps @fommil could explain this.

fommil: I think Evil Sam wrote that... I don't have a clue!

fommil:
LGTM

sugakandrey:
Too big of a user boost messed with `BasicWorkflow` tests, should be more adequate now

PR #1543, opened 2016-06-30
use correct server version in emacs tests

fommil:
@ktonga this is why your PR was failing. the emacs tests were trying to install a version that no longer exists.


fommil:
*sigh* looks like the CLA service is broken.

LGTM

PR #1546, opened 2016-07-01
Add `typeParams` to `ArrowTypeInfo`. Closes #1504

### LINKED ISSUE #1504: ArrowTypeInfo should have a list of type parameters

ktonga:
Still want my stickers @fommil

### DIFF
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -451,6 +451,7 @@ sealed trait TypeInfo extends EntityInfo {
   def typeArgs: Iterable[TypeInfo]
   def members: Iterable[EntityInfo]
   def pos: Option[SourcePosition]
+  def typeParams: List[TypeInfo]
 
   final def declaredAs = declAs
   final def args = typeArgs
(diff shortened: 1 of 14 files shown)


ktonga:
Emacs tests failed, I guess it has something to do with a test asserting the exact response from the server. If this is the case, it's ok that it fails since a new field has been introduced, and as far as I could see, empty lists are included in the serialised body. Excluding `None`s and empty `List`s from the serialisation (or the assertion) could be a way for the tests to need less attention I think.

fommil:
Which country are you in?

fommil:
Hmm, might be cheaper to get a fresh batch made up.

ktonga:
I'm in Australia/Sydney. Just joking, did't want to bother you, I read it in the PR template.

fommil:
@ktonga no no, we need to get some stickers out there, but you'll be responsible for mailing them to other AU/NZ contributors  :stuck_out_tongue:

fommil:
ok, definitely something funky going on with the drone build. This is more or less the same thing we see in the graphpocalypse @sugakandrey I'm all out of ideas, this has come from nowhere.

fommil:
maybe the answer is to just not bail if we fail to delete that directory, and just log it instead.

fommil:
LGTM the honour is all yours to merge!

fommil:
:tada:

Issue #1507, opened 2016-06-09
toSymbol fails to resolve classnames with special characters


sugakandrey:
[This](https://github.com/ensime/ensime-server/blob/2.0/core/src/main/scala/org/ensime/core/FqnToSymbol.scala#L49-L58) will fail to produce a symbol for the `fqn` of a class with special characters in its name (e.g. `Predef.<:<` `ClassName(PackageName(List("Predef")), "$less$colon$less")`)

fommil:
ah, well spotted. I believe it works in the other direction though. Should be easy to add a unit test.

sugakandrey:
I added the test in my `scalap-converter` branch (15c03fc), will try to fix that.

sugakandrey:
Fixed in 7f8406f, I'll leave this open until `scalap-converter` is ready to be merged.

Issue #1472, opened 2016-05-29
Scalap symbol <=> FQN Converter 


sugakandrey:
The idea is to implement a way to convert between `scalasig.Symbol`/`scalasig.Type` instances and `FullyQualifiedName` domain objects. See #1459 for similar functionality for pres. compiler Symbols/Types.

PR #1548, opened 2016-07-03
scalasig.Symbol to FQN converter

### LINKED ISSUE #1547: FqnToSymbol doesn't handle overloaded symbols properly
Both `toSymbol(scalaName: String, rootSymbol: Symbol = RootClass): Symbol` and `toSymbol(fqn: FullyQualifiedName): Symbol` will fail on the following case:

```
package org.example
class Foo {
  object bar {
    class Baz
  }
  def bar(i: Int): Unit = ???
}
```
If we ask `FqnToSymbol` for the symbol of `class Baz` it will fail, because intermediate symbol for `bar` will be overloaded and calling `barSymbol.info.member(newTypeName("Baz")` will return `NoSymbol` [here](https://github.com/sugakandrey/ensime-server/blob/master/core/src/main/scala/org/ensime/core/FqnToSymbol.scala#L30) and [here](https://github.com/sugakandrey/ensime-server/blob/master/core/src/main/scala/org/ensime/core/FqnToSymbol.scala#L58). Fixing this as part of #1472

sugakandrey:
Closes #1547, #1507, #1472. I probably missed some bizarre edge cases but it works most of the time :)


sugakandrey:
2.10 build will sometimes fail with `StackOverflow` error, this is a bug in presentations compiler [see](https://issues.scala-lang.org/browse/SI-7221).

fommil:
Great! I'll try to find time to look at this tomorrow

fommil:
LGTM

PR #1550, opened 2016-07-04
Highlight class parameters as valFields.

marcsaegesser:
Class parameters are treated as val fields by the compiler but are
flagged as PARAMACCESSOR rather than ACCESSOR. The semantic highlighter
would previously highlight the definition of class parameters as
valFields but then not highlight their usage as anything.


fommil:
LGTM, do you have permission to merge?

fommil:
(prefer a squash commit to a merge)

PR #1552, opened 2016-07-09
Proactively start indexing on initialisation rather than first message - fixes #1551

### LINKED ISSUE #1551: Project actor init should happen on creation, not on connection info request
We're delaying indexing until the connection is made, but this means we can't make use of the `ensime.exitAfterIndex` feature.

@rorygraves thoughts? Any reason you can think of why we only init on a connection?

rorygraves:


core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -26,10 +25,10 @@ final case class ShutdownRequest(reason: String, isError: Boolean = false)
 class Project(
     broadcaster: ActorRef,
     implicit val config: EnsimeConfig
-) extends Actor with ActorLogging with Stash {
+) extends Actor with ActorLogging {
   import context.{ dispatcher, system }
 
-  import FileUtils._
+  import org.ensime.util.FileUtils._
```
fommil: this looks a bit gnarly

rorygraves: It was their already - I made it fully qualified rather than having FileUtils imported above and then _ imported here.

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -53,19 +52,9 @@ class Project(
   }
   private val classfileWatcher = context.actorOf(Props(new ClassfileWatcher(config, searchService :: reTypecheck :: Nil)), "classFileWatcher")
 
-  def receive: Receive = awaitingConnectionInfoReq
+  def receive: Receive = handleRequests
 
-  def awaitingConnectionInfoReq: Receive = withLabel("awaitingConnectionInfoReq") {
-    case ConnectionInfoReq =>
-      sender() ! ConnectionInfo()
```
fommil: it looks like we're not sending `ConnectionInfo` at all now. Should you maybe send it in the server connection-handler actor?

rorygraves: You misunderstand the change in logic - its handled in the handleRequests method
The original code was a hack to say 'handle this one, and trigger indexing, stash anything until we have seen it.

fommil:
LGTM

fommil:
emacs is getting caught up on this http://fommil.com/ensime/ensime-server/765

should we disable the emacs tests in the server and fix it client side?

fommil:
it would be good if the project sent a persisted connection-info message during startup. That simplifies the client code because it means you don't need to wait for the connection and then send a message and wait for a response... you can just fire off connection handling logic and wait async until the connection info returns.

rorygraves:
It can't - its a response to a ConnectionInfoRequest
The server will respond correctly to that message - any will send the aysnc messages as an when they aer available (possibly immediately on connection).  These feels like hack logic that should not exist on the server.

rorygraves:
You should just be able to connect and go full async - assume the server is up and wait for the ready messages to tell you when its available for work.
You probably don't need to send ConnectionInfoReq at all

rorygraves:
I suspect you don't check the result - so its just a ping-pong.

rorygraves:
Ok, the emacs expectations about startup behaviour are subtle and I'm going to have to introduce a new hack until such times as we fix the protocol (it assumes that the first thing it will ever see is a response to the a connectionInfoReq and not any other async message).

aemoncannon:
Some history: we were adjusting our requests based on the protocol version, but there was no deterministic ordering of startup messages, so we'd end up firing early RPCs not compatible with the current protocol. That's why we do the blocking handshake before anything else. Any plan to get rid of the handshake should probably include a revised plan for handling protocol evolution.

References:
https://github.com/ensime/ensime-server/pull/1142#issuecomment-154190104
https://github.com/ensime/ensime-emacs/pull/282/files
https://github.com/ensime/ensime-emacs/pull/284#issuecomment-154219528

aemoncannon:
The elisp tests are designed to ensure we get the protocol version before compiler ready is sent (because compiler ready kicks off a bunch of other rpcs).

rorygraves:
Yeah, I'm going to do this in two passes:
1) Tweak the above to ensure that the ConnectionInfo is handled first (i.e. implement the current behaviour, but not in Project)
2) Move to a welcome message (i.e. first message on channel will be a version info - you don't need to request it - and kill off that syncronous message.

fommil:
dead as a dodo, could form the basis of a revitalised attempt though, coming soon.

PR #1554, opened 2016-07-09
SWANKY

### LINKED ISSUE #1189: SWANKY (over WebSockets)
can't believe we didn't have a ticket for this.

fommil:
close #1189 

this is WIP because I haven't created all the test assertions yet, and that's expected to fail.

This enables the SWANKY protocol over WebSockets and also adds an intermediate SWANKI protocol over TCP.

@aemoncannon I decided to go with data style instead of alist style, as we've been doing so far. The names of classes and fields are raw (which means camel case)... what do you think to making them use emacs-lisp style `lower-case-hyphened`? It shouldn't be too much effort.


fommil:
@propensive since it never went anywhere, this is removing the RPC variant of the ENSIME API... we definitely want to encourage WebSockets.

fommil:
@ktonga FYI

fommil:
awesome, if we need to experiment, emacs has our backs https://www.emacswiki.org/emacs/CamelCase

fommil:
slight digression...

```elisp
(defun cycle-camel-dashes (name)
  "Cycle NAME between `CamelCase' and `dashed-notation'."
  ;; (cycle-camel-dashes ":FooBarBaz")
  ;; (cycle-camel-dashes ":foo-bar-baz")
  (let (case-fold-search)
    (cl-flet* ((split-camels (s) (replace-regexp-in-string "\\([a-z]\\)\\([A-Z]\\)" "\\1 \\2" s))
               (split-parts (s) (s-split "[- ]+" (s-downcase (split-camels s)))))
      (let* ((split (split-parts name))
             (replacement (cond
                           ((s-contains? "-" name) (s-join "" (-map #'capitalize split)))
                           (t                      (s-join "-" split)))))
        replacement))))

(defun symbol-at-point-cycle-camel-dashes ()
  "Cycle the symbol at point between `CamelCase' and `dashed-notation'."
  (interactive)
  (let* ((name (symbol-name (symbol-at-point)))
         (replacement (cycle-camel-dashes name)))
    (message "replacing %s with %s" name replacement)
    (search-forward name)
    (replace-match replacement t)))

(global-set-key (kbd "C-c D") 'symbol-at-point-cycle-camel-dashes)
````

**Comment by fommil** [comment]
*sigh* foiled by the fact that `scala-mode` doesn't return the correct `symbol-at-point` when there is a dash inside a name.

**Comment by fommil** [comment]
oooh yeah

```elisp
(defun symbol-at-point-cycle-camel-dashes ()
  "Cycle the symbol at point between `CamelCase' and `dashed-notation'."
  (interactive)
  (let* ((name (save-excursion
                 ;; workaround bad scala symbol-at-point logic
                 (search-forward-regexp "[^ ]+")
                 (match-string 0)))
         (replacement (cycle-camel-dashes name)))
    (message "replacing %s with %s" name replacement)
    (search-forward name)
    (replace-match replacement t)))
```

fommil:
emacs tests are failing due to missing `dash` dependency. I have no idea what's going on there, maybe melpa is down.

fommil:
same again, I think I'll have to make some changes to the docker image to get emacs 24.5.

fommil:
hmm, I upgraded to emacs 24.5 and it still fails.

fommil:
emacs fixed, rebuilding.

aemoncannon:
Not totally sure what you're after here. Can we do the camel conversions on
server? In and out? It's going to be a lot slower in emacs ( especially
responses)

fommil:
Yeah server side.

fommil:
I'll do the hyphenated cleanup in a follow-up PR, and also some small changes like custom formatters for debug/object ids.

PR #1558, opened 2016-07-10
SWANKY improvements

fommil:
some cleanups to the API.

A few outstanding thoughts:

- should we prefix all type hints with `ensime-` or `ensime-api-` ? It's really easy, and consistent on the emacs side, but it's really ugly.
- ~~the format for `case object`s is a bit ugly, like `(:Foo nil)` instead of `:Foo`. Changing it is a bit involved.~~ turned out to be easy, pushing a commit.

### DIFF
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -331,6 +331,7 @@ object DebugThreadId {
    * @param s A Long encoded as a string
    * @return A ThreadId
    */
+  @deprecating("no code in the API")
   def apply(s: String): DebugThreadId = {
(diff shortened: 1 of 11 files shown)


fommil:
@ensime/atom-panel @ensime/vim-panel this introduces a very small change to the API. The capitalisation of two debugging requests are `Vm` instead of `VM`.

fommil:
just a quick check to make sure that emacs is ok with keywords as values, and it is

```elisp
(cl-destructuring-bind
    (&key foo bar)
    '(:foo :hello :bar :hello)
  (message "%s %s" foo bar)) => :hello :hello
```

fommil:
LGTM

PR #1560, opened 2016-07-11
OrientDB schema change

sugakandrey:
A couple of OrientDB convenience methods, some typesafe wrappers here and there, `BigDataFormat` for coproducts, basic support for viewing type hierarchies via `searchService.getClassHierarchy(someTrait, HierarchyType.Subclasses)`.

### DIFF
--- a/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
+++ b/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
@@ -5,6 +5,7 @@ package org.ensime.indexer
 import scala.concurrent._
 import scala.concurrent.duration._
 import org.ensime.fixture._
+import org.ensime.indexer.graph.{ ClassDef, ClassHierarchy, HierarchyType }
 import org.ensime.util.EnsimeSpec
 import org.ensime.util.file._
 import org.scalatest.Matchers._
@@ -205,6 +206,43 @@ class SearchServiceSpec extends EnsimeSpec
 
   "exact searches" should "find type aliases" in withSearchService { implicit service =>
     service.findUnique("org.scalatest.fixture.ConfigMapFixture$FixtureParam") shouldBe defined
+  }
+
+  "class hierarchy viewer" should "find all classes implementing a trait" in withSearchService { implicit service =>
+    val someTrait = "org.hierarchy.SomeTrait"
+    val implementingClasses = service.getClassHierarchy(someTrait, HierarchyType.Subclasses)
+    implementingClasses shouldBe defined
+    inside(implementingClasses.get) {
+      case ClassHierarchy(classDef, refs) =>
+        classDef.fqn should ===(someTrait)
+        inside(refs) {
+          case Seq(ref1, ref2) =>
+            inside(ref1) {
+              case ClassHierarchy(aClass, Seq(subclass)) =>
+                aClass.fqn should ===("org.hierarchy.ExtendsTrait")
+                inside(subclass) {
+                  case cdef: ClassDef => cdef.fqn should ===("org.hierarchy.Subclass")
+                }
+            }
+            inside(ref2) {
+              case cdef: ClassDef => cdef.fqn should ===("org.hierarchy.ExtendsTraitToo")
+            }
+        }
+    }
+  }
+
+  it should "find all superclasses of a class" in withSearchService { implicit service =>
+    val hierarchy = service.getClassHierarchy("org.hierarchy.Qux", HierarchyType.Superclasses)
+    hierarchy shouldBe defined
+    hierarchy.get.toSet.map(_.fqn) should contain theSameElementsAs Set(
+      "org.hierarchy.Qux",
+      "scala.math.Ordered",
+      "org.hierarchy.Bar",
+      "org.hierarchy.NotBaz",
+      "java.lang.Runnable",
+      "java.lang.Comparable",
+      "java.lang.Object"
+    )
   }
 }
 
--- a/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
+++ b/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
@@ -60,6 +60,7 @@ class SearchServiceSpec extends EnsimeSpec
 
       classfile.delete()
       refresh() shouldBe ((1, 0))
+      searchExpectEmpty("org.example.Foo")
     }
   }
 
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -34,6 +34,7 @@ class SearchService(
 ) extends ClassfileIndexer
     with FileChangeListener
     with SLF4JLogging {
+  import SearchService._
 
   private[indexer] def isUserFile(file: FileName): Boolean = {
     (config.allTargets map (vfs.vfile)) exists (file isAncestor _.getName)
@@ -170,15 +171,15 @@ class SearchService(
 
   def refreshResolver(): Unit = resolver.update()
 
-  def persist(check: FileCheck, symbols: List[FqnSymbol], commitIndex: Boolean, boost: Boolean): Future[Int] = {
-    val iwork = Future { blocking { index.persist(check, symbols, commitIndex, boost) } }
+  def persist(check: FileCheck, symbols: List[(RawSymbol, SourceInfo)], commitIndex: Boolean, boost: Boolean): Future[Int] = {
+    val iwork = Future { blocking { index.persist(check, symbols.map(_._1), commitIndex, boost) } }
     val dwork = db.persist(check, symbols)
     iwork.flatMap { _ => dwork }
   }
 
   // this method leak semaphore on every call, which must be released
   // when the List[FqnSymbol] has been processed (even if it is empty)
-  def extractSymbolsFromClassOrJar(file: FileObject): Future[List[FqnSymbol]] = {
+  def extractSymbolsFromClassOrJar(file: FileObject): Future[List[(RawSymbol, SourceInfo)]] = {
     def global: ExecutionContext = null // detach the global implicit
     val ec = actorSystem.dispatchers.lookup("akka.search-service-dispatcher")
 
@@ -206,7 +207,7 @@ class SearchService(
   private val blacklist = Set("sun/", "sunw/", "com/sun/")
   private val ignore = Set("$$", "$worker$")
   import org.ensime.util.RichFileObject._
-  private def extractSymbols(container: FileObject, f: FileObject): List[FqnSymbol] = {
+  private def extractSymbols(container: FileObject, f: FileObject): List[(RawSymbol, SourceInfo)] = {
     f.pathWithinArchive match {
       case Some(relative) if blacklist.exists(relative.startsWith) => Nil
       case _ =>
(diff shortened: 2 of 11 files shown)


core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -205,6 +207,43 @@ class SearchServiceSpec extends EnsimeSpec
   "exact searches" should "find type aliases" in withSearchService { implicit service =>
     service.findUnique("org.scalatest.fixture.ConfigMapFixture$FixtureParam") shouldBe defined
   }
+
+  "class hierarchy viewer" should "find all classes implementing a trait" in withSearchService { implicit service =>
```
fommil: would this also find all implementing `abstract class`es and sub `trait`s?

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -219,22 +220,13 @@ class SearchService(
...
-              FqnSymbol(None, name, path, method.name.fqnString, None, sourceUri, method.line)
-            } ::: clazz.fields.toList.filter(_.access == Public).map { field =>
-              val internal = field.clazz.internalString
-              FqnSymbol(None, name, path, field.name.fqnString, Some(internal), sourceUri, clazz.source.line)
-            } ::: depickler.getTypeAliases.toList.filter(_.access == Public).filterNot(_.fqn.contains("<refinement>")).map { rawType =>
-              // this is a hack, we shouldn't be storing Scala names in the JVM name space
-              // in particular, it creates fqn names that clash with the above ones
-              FqnSymbol(None, name, path, rawType.fqn, None, sourceUri, None)
-            }
+          (clazz :: clazz.methods.toList ::: clazz.fields.toList ::: depickler.getTypeAliases.toList).map(_ -> sourceFileInfo)
```
fommil: we should make these underlying impls return `List` so we don't have to convert them.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -251,6 +243,8 @@ class SearchService(
   /** only for exact fqns */
   def findUnique(fqn: String): Option[FqnSymbol] = Await.result(db.find(fqn), QUERY_TIMEOUT)
 
+  def getClassHierarchy(fqn: String, hierarchyType: HierarchyType): Option[Hierarchy] = Await.result(db.getClassHierarchy(fqn, hierarchyType), QUERY_TIMEOUT)
```
fommil: the legacy API demanded `Option` instead of `Future`, if it's possible we should go for async here

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -289,6 +283,10 @@ class SearchService(
   }
 }
 
+object SearchService {
+  type SourceInfo = (String, String, Option[String])
```
fommil: perhaps better as a `case class` with named fields? Raw like this makes it hard to remember what the tuple entries are.

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -26,7 +27,63 @@ import org.ensime.util.file._
...
+  def source: Option[String]
+  def declAs: DeclaredAs
+  def access: Access
+
+  def sourceFileObject(implicit vfs: EnsimeVFS): Option[FileObject] = source.map(vfs.vfile)
+}
+
+sealed trait Hierarchy {
+  def toSet: Set[ClassDef] = this match {
+    case cdef @ ClassDef(_, _, _, _, _, _) => Set(cdef)
```
fommil: no need to match on the contents, just `case cdef: ClassDef` should work

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -26,7 +27,63 @@ import org.ensime.util.file._
...
+    case ClassHierarchy(cdef, refs) => Set(cdef) ++ refs.flatMap(_.toSet)
+  }
+}
+
+final case class ClassHierarchy(aClass: ClassDef, classRefs: Seq[Hierarchy]) extends Hierarchy
+
+sealed trait HierarchyType
+object HierarchyType {
+  case object Subclasses extends HierarchyType
```
fommil: no need to put this inside an `object`, that just confuses things

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -26,7 +27,63 @@ import org.ensime.util.file._
...
+sealed trait Hierarchy {
+  def toSet: Set[ClassDef] = this match {
+    case cdef @ ClassDef(_, _, _, _, _, _) => Set(cdef)
+    case ClassHierarchy(cdef, refs) => Set(cdef) ++ refs.flatMap(_.toSet)
+  }
+}
+
+final case class ClassHierarchy(aClass: ClassDef, classRefs: Seq[Hierarchy]) extends Hierarchy
+
+sealed trait HierarchyType
```
fommil: try not to put the word `Type` into a type :wink:

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -26,7 +27,63 @@ import org.ensime.util.file._
...
+  def line: Option[Int]
+  def source: Option[String]
+  def declAs: DeclaredAs
+  def access: Access
+
+  def sourceFileObject(implicit vfs: EnsimeVFS): Option[FileObject] = source.map(vfs.vfile)
+}
+
+sealed trait Hierarchy {
+  def toSet: Set[ClassDef] = this match {
```
fommil: the definition of `toSet` tells me that it is perhaps best as a typeclass or in a utility method somewhere else?

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -26,7 +27,63 @@ import org.ensime.util.file._
...
+sealed trait HierarchyType
+object HierarchyType {
+  case object Subclasses extends HierarchyType
+  case object Superclasses extends HierarchyType
+}
+
+final case class ClassDef(
+    fqn: String,
+    file: String,
+    path: String,
```
fommil: what's the difference between `file` and `path`? Would using `Path` make sense here?

fommil: ah, just spotting the comments on the original file.. definitely sensible to keep them around.

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -40,30 +97,66 @@ object FileCheck extends ((String, Timestamp) => FileCheck) {
...
-    if (fqn.contains("(")) DeclaredAs.Method
-    else if (internal.isDefined) DeclaredAs.Field
-    else DeclaredAs.Class
-}
 
 // core/it:test-only *Search* -- -z prestine
 class GraphService(dir: File) extends SLF4JLogging {
+  import org.ensime.indexer.graph.GraphService._
+
+  implicit object TimeStampSPrimitive extends SPrimitive[Timestamp] {
```
fommil: these `SPrimitive` impls can probably be moved into the stringymap package / files.

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -119,92 +212,69 @@ class GraphService(dir: File) extends SLF4JLogging {
...
+      symbols.foreach {
+        case (s: RawClassfile, (file, path, source)) =>
+          val classDef = ClassDef(s.name.fqnString, file, path, source, s.source.line, s.access)
+          val classV = RichGraph.upsertV[ClassDef, String](classDef)
+          RichGraph.insertE(classV, fileV, DefinedIn)
+          // nulls are used, because these fields are not really optional and are guaranteed to be
+          // set in the graph after the indexing is completed.
+          val superClass = s.superClass.map(name => ClassDef(name.fqnString, null, null, None, None, null))
+          val interfaces = s.interfaces.map(name => ClassDef(name.fqnString, null, null, None, None, null))
+          (superClass.toList ::: interfaces).foreach { cdef =>
```
fommil: same as `.toList` comment above

core/src/main/scala/org/ensime/indexer/orientdb/SimpleOrient.scala
```diff
@@ -32,9 +34,77 @@ package api {
   case class VertexT[T](underlying: Vertex)
 
   trait EdgeT[+Out, +In]
+
+  sealed trait IndexT {
+    def orientKey: String
+  }
+
+  object IndexT {
```
fommil: same comment as above about this `sealed trait` with implementations in the companion pattern, it's a bit unusual.

core/src/main/scala/org/ensime/indexer/orientdb/SimpleOrient.scala
```diff
@@ -32,9 +34,77 @@ package api {
...
+  }
 }
 
 package object syntax {
+  import org.ensime.indexer.orientdb.api.{ IndexT, Indexable }
+
+  implicit class RichOrientGraph(graph: OrientExtendedGraph) {
+    private val schema = graph.getRawGraph.getMetadata.getSchema
+
+    def createIndexOn[E <: Element, T, F](idx: IndexT = IndexT.NotUnique)(
```
fommil: this typesafe schema is pretty cool. I wonder if we should move it into a separate package. With the lucene helpers we could create a `database` package or something like that.

core/src/main/scala/org/ensime/indexer/orientdb/SimpleOrient.scala
```diff
@@ -157,9 +246,9 @@ package object syntax {
                 v.setProperty(key, value)
           }
 
-          (keys -- updates.keySet).foreach { nulled =>
-            v.removeProperty[AnyRef](nulled)
-          }
+          //          (keys -- updates.keySet).foreach { nulled =>
```
fommil: commented code

core/src/main/scala/org/ensime/indexer/stringymap/StringyMap.scala
```diff
@@ -29,21 +34,25 @@ package api {
   trait SPrimitive[T] {
     def toValue(v: T): AnyRef
     def fromValue(v: AnyRef): T
+    def getType: (OType, Boolean)
```
fommil: what is the `Boolean` for? Adding a line comment here would be good.

core/src/main/scala/org/ensime/indexer/stringymap/StringyMap.scala
```diff
@@ -68,6 +78,7 @@ package object impl {
     def label: String = ???
     def toProperties(t: HNil): StringyMap = new java.util.HashMap()
     def fromProperties(m: StringyMap) = Right(HNil)
+    def toSchema: Map[String, (OType, Boolean)] = Map()
```
fommil: `Map.empty` is more efficient. For fun, follow through the code that gets invoked when you call `Map()`.

core/src/main/scala/org/ensime/indexer/stringymap/StringyMap.scala
```diff
@@ -116,9 +133,49 @@ package object impl {
...
+  }
+
+  implicit def CNilBigDataFormat[T]: BigDataFormat[CNil] = new BigDataFormat[CNil] {
+    override def label: String = ???
+    override def toProperties(t: CNil): StringyMap = ???
+    override def toSchema: Map[String, (OType, Boolean)] = ???
+    override def fromProperties(m: StringyMap): BigResult[CNil] = ???
+  }
+
+  implicit def CoproductBigDataFormat[Key <: Symbol, Value, Tail <: Coproduct](
```
fommil: by adding the `label` onto the `HList` implementation I guess we don't really need a coproduct implementation because that makes it handle classes that aren't part of a sealed family. With that in mind, maybe we should do it properly and drop the `label` and only allow it to work with sealed families?

project/SensibleSettings.scala
```diff
@@ -51,7 +51,7 @@ object Sensible {
     javaOptions += "-Dfile.encoding=UTF8",
     //this is needed by OrientDB versions 2.2.0+, 512g is recommended on their github wiki
     //https://github.com/orientechnologies/orientdb-docs/blob/master/Release-2.2.0.md
-    javaOptions += "-XX:MaxDirectMemorySize=512g",
+    javaOptions += "-XX:MaxDirectMemorySize=4g",
```
fommil: I added this into the `EnsimeBuild.scala` in my branch. I think we should probably update the ensime 2.0 version of `sbt-ensime` to have this by default so that we don't have to tell people to do it.

fommil:
cool, this is a really good PR. I have just given it a quick once over and haven't done a deep study locally, but I don't see any major problems --- just the points above that would be good to address in a follow up PR.

LGTM

fommil:
the emacs tests are acting up, I can't see how this would cause a problem in `Ensime integration test compile-only.`

fommil:
oh crap, of course if the emacs tests are acting up then that means we won't be able to get any auto published releases. OK, that needs to be fixed. Feel free to comment out the emacs tests in your follow up PR with fixes to the above.

fommil:
here's the problem... reverting

```
Starting test: 'Test inspect type at point.'
ENSIME server starting...
Server connection closed unexpectedly: connection broken by remote peer

Connecting to Swank on port 49613..
Connected to ENSIME speaking protocol 1.9.0, please wait while the project is loaded.
WARNING more than one pending message
WARNING more than one pending message
WARNING more than one pending message
Initializ
ENSIME ready. CanBuildFrom[List[Dream], Reality, List[Reality]]
WARNING more than one pending message
WARNING more than one pending message
WARNING more than one pending message
WARNING more than one pending message
WARNING no response to messages: (5 4) . Waiting some more.
ERROR no response to messages: (5 4)
Server connection closed unexpectedly: connection broken by remote peer



Starting test: 'Test inspect type in range.'
ENSIME server starting...
Connecting to Swank on port 55865..
Connected to ENSIME speaking protocol 1.9.0, please wait while the project is loaded.
WARNING more than one pending message
WARNING more than one pending message
WARNING more than one pending message
Initializ
ENSIME ready. May the _ be with you.
WARNING more than one pending message
WARNING more than one pending message
WARNING more than one pending message
WARNING more than one pending message
WARNING no response to messages: (5 4) . Waiting some more.
ERROR no response to messages: (5 4)
Server connection closed unexpectedly: connection broken by remote peer
```

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
sugakandrey: `superClass` here is actually an `Option`, so I guess `.toList` is okay here

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
sugakandrey: I generally try to follow this rule, but I have no idea how to name this trait without `Type` in it, because it is a **type** that represents a **type** of hierarchy. :)

core/src/main/scala/org/ensime/indexer/SearchService.scala
sugakandrey: Why are these fields `Queue`s in the first place? We never seem to use FIFO nature of them anyways.

PR #1562, opened 2016-07-12
SWANKY: prefix with ensime-api- namespace

fommil:
many many many thanks to @bvenners for the implicit `Position` in scalatest, which now means that the emacs test cycle looks like

- `C-c c` (run test)
- `C-c e` (jump to exact line of test failure)

fix and repeat. It helped me just cut through this change.


fommil:
LGTM

PR #1565, opened 2016-07-16
Switch from BuildTime to build date

dwijnand:
This makes continuous compilation faster, because it doesn't keep
compiling the BuildInfo object with the updated build timestamp.

It redefines "builtAtString" format to "yyyy-MM-dd", where it was
previously "yyyy-MM-dd HH:mm:ss.SSS"; and it also drops "builtAtMillis".

### DIFF
--- a/project/EnsimeBuild.scala
+++ b/project/EnsimeBuild.scala
@@ -163,8 +163,14 @@ object EnsimeBuild extends Build {
     ) enablePlugins BuildInfoPlugin settings (
         buildInfoPackage := organization.value,
         buildInfoKeys += BuildInfoKey.action("gitSha")(Try("git rev-parse --verify HEAD".!! dropRight 1) getOrElse "n/a"),
-        buildInfoOptions += BuildInfoOption.BuildTime
+        buildInfoKeys += BuildInfoKey.action("builtAtString")(currentDateString())
       )
+
+  private def currentDateString() = {
+    val dtf = new java.text.SimpleDateFormat("yyyy-MM-dd")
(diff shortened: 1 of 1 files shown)


fommil:
transient failure in 2.10

LGTM

fommil:
weird, same transient twice in a row... I'm starting to think this could have triggered some weird scala 2.10 presentation compiler bug.

sugakandrey:
@fommil If I saw the failure correctly, this is indeed  a presentation compiler bug, which occurs in `ScalapSymbolToFqnSpec` test, I commented on this in #1548, maybe it makes sense to just disable these tests for 2.10?

fommil:
@sugakandrey yeah, I remember you saying this... but three times in a row seems a little weird. Let's disable the tests in 2.10. The way we can do this is by moving them into the `scala-2.11` test folder, there are already a few tests that need this.

sugakandrey:
@fommil It is just one test in the suite that fails, I think I saw something like this somewhere in `RichPresentationCompilerSpec`
```
    val scalaVersion = scala.util.Properties.versionNumberString
    val parts = scalaVersion.split("\\.").take(2).map { _.toInt }
    if (parts(0) > 2 || (parts(0) == 2 && parts(1) > 10)) {
      ...
    }
```
Should I move it to 2.11 folder or just skip it like this?

fommil:
@sugakandrey ideally let's move that one test into the 2.11-only suite

fommil:
another thing we could do is flag it as a flakey test like we do with the file watcher tests...

fommil:
but that does assume that it'll pass given a second chance.

PR #1573, opened 2016-07-25
Replace Akka HTTP with Netty and Akka Stream with... functions

ktonga:
Both message formats, `jerky` and `swanky`, are now handled at `/websocket`

The clients must specify the desired format using WS subprotocol negotiation, it's a standard mechanism to establish the format of the messages between the client and the server supported by most of the websocket libraries.

Here there is an example for Python:

```python
ws = websocket.create_connection("ws://localhost:1234/websocket", subprotocols=["jerky"])
```

Also:

- Implement HTTP and WebSocket services with Netty
- Remove Akka HTTP and Akka Stream
- Replace XML literals with [scalatags](https://github.com/lihaoyi/scalatags)
- `DocJarReading` is now a trait for better testability
- Rewrite the tests for adapt them to Netty
- Fix `sbt assembly` merge error for netty jars
- Poor man's stashing to sort out subprotocol negotiation waiting (all this hack should be removed soon. See netty/netty#5734)

~~This is a work in progress:~~

- [x] Get ride of XML literals
- [x] Pattern matching for HTTP routing
- [x] Improve error handling
- [x] Add some logs
- [x] Serve both formats, __jerky__ and __swanky__ at the same port.
- [x] Add unit tests for Netty stuff


fommil:
oh wow, I thought this had gone cold :smile_cat: I'll try to find some time tomorrow to look at this.

In the meantime, please run the commands in the `.drone.yml` to apply our standardised headers. We're using `sbt-header` and it just keeps everything nice and tidy.

project/EnsimeBuild.scala
```diff
@@ -187,11 +190,9 @@ object EnsimeBuild extends Build {
       commonSettings ++ commonItSettings
     ).settings(
         libraryDependencies ++= Seq(
-          "com.typesafe.akka" %% "akka-stream-experimental" % streamsVersion,
```
fommil: muuaahahahahahhaaha!!

server/src/main/scala/org/ensime/server/DocsHandler.scala
```diff
@@ -0,0 +1,98 @@
...
+import io.netty.handler.codec.http.HttpResponseStatus.{ BAD_REQUEST, FORBIDDEN, NOT_FOUND, OK }
+import io.netty.handler.codec.http.HttpVersion.HTTP_1_1
+import io.netty.handler.codec.http.{ DefaultFullHttpResponse, FullHttpRequest, HttpUtil }
+import io.netty.util.CharsetUtil
+
+import org.ensime.core.DocJarReading
+
+class DocsHandler(docs: DocJarReading) extends SimpleChannelInboundHandler[FullHttpRequest] {
+
+  val docRegex = """/docs/([^/]+\.jar)/(.+)""".r
```
fommil: aside: if you use a capital on a regex value, you can use it in the pattern matcher more easily

server/src/main/scala/org/ensime/server/DocsHandler.scala
```diff
@@ -0,0 +1,98 @@
...
+
+    // Send the response and close the connection if necessary.
+    val f: ChannelFuture = ctx.channel().writeAndFlush(res)
+    if (!HttpUtil.isKeepAlive(req) || res.status().code() != OK.code())
+      f.addListener(ChannelFutureListener.CLOSE)
+  }
+
+  private val docJarsPage: ByteBuf = {
+    val content =
+      <html>
```
fommil: it'd be nice to be rid of the XML literals... but that's a digression. Maybe another ticket for 2.0

fommil:
272 additions, 527 deletions... so much for reactive frameworks keeping down the boilerplate!

ktonga:
> 272 additions, 527 deletions... so much for reactive frameworks keeping down the boilerplate!

To be fair, I just deleted the tests, I didn't replace them with the new ones yet.

ktonga:
But indeed, Akka Stream boilerplate is insane.

server/src/main/scala/org/ensime/server/DocsHandler.scala
ktonga: I've been thinking about wrapping netty model objects into a case class so I can express the http mappings in a pattern match. Will try it to see if it looks nicer.

server/src/main/scala/org/ensime/server/DocsHandler.scala
ktonga: I don't like it either. What do you recommend? Plain string interpolation? A template engine maybe? Or case classes to xml with some marshaller?

fommil: There was a project to do xml interpolation but I'm not sure it went anywhere. There is also Li Haoyi's tags library.

fommil:
At yes, good point. I think we'll definitely be wanting those tests before merging.

fommil:
sorry I completely forgot to look at this. Keep pinging me if I forget. BTW, please email me your address so I can send a load of typelevel stickers your way.

fommil:
oh wow, green lights. I didn't realise you'd got that far. Would you like a review before you continue with the items or should I do that at the end?

ktonga:
@fommil early reviews are always welcome 😄 

I have the same level of test coverage, and all the functionality is already in place, so I guess it's a good time for you to take a look and object if you see something you really don't like. I'd like to go further and totally remove Akka from the server module but that seems to be out of scope for this very PR. And I think the old RPC service can't be killed yet.

Currently I'm kind of blocked by the impossibility to start both protocols at the same time, I already filled a Netty issue ( netty/netty#5696 ) to figure out how I could sort this out, but they didn't answer yet. And any other approach would need non-retrocompatible changes.

Any thoughts?

fommil:
Hmm. How about asking on stack overflow? Might be a bigger audience there.

ktonga:
I already check in OS (in everywhere actually) and [the question](http://stackoverflow.com/questions/24781886/create-multiple-websocket-server-with-one-port-number) already exist but I don't like the answer __AT ALL__. They suggest to put separate Netty servers on different ports behind a Nginx to do the routing/url-rewrite magic, what I find unacceptable for our usecase. I'm just hoping to get a better answer from a Netty member.

TBH I think we are doing it wrong, different URIs just to encode the messages differently doesn't sound like the right way to do it.

fommil:
Hmm. What an unexpected problem.

Maybe we could differentiate protocol with some custom header message. Can that run off the same endpoint as a regular GET/POST? Thinking / with /jerky for a while to let clients catch up.

ktonga:
I was thinking that the only real usecase for the server is to serve just one of the protocols at a time. So maybe we can fix the url to `/ws`, but hook the message enconding to either swanky or jerky at server launch time reusing the `ensime.protocol` argument.

ktonga:
It looks like simple content type negotiation, but I dunno how it works for WS. If you are ok with a non-backward-compatible change to solve this I can research further today.

fommil:
Yeah we could do it that way. Is there any way to initially forward `/jerky` to the `/ws` so that we don't break atom/vim/scalanator/etc?

It would be nicer if we had a header value to set the payload messages instead of a server option.

ktonga:
I would prefer that too. Will investigate tonight.

fommil:
cool! BTW, I made a small change to the CLA (`s-express` is LGPL) and now it wants everybody to sign it again, sorry. Ping me when you feel it's in a state to be merged. Since we don't have integration test coverage around the websockets protocol, let me know if it works for you in Vim.

server/src/main/scala/org/ensime/server/WebServer.scala
```diff
@@ -2,84 +2,69 @@
...
 
-import scala.concurrent.Future
+import org.ensime.api.{ RpcRequestEnvelope, RpcResponseEnvelope }
+import org.ensime.core.DocJarReading
 
-trait WebServer {
-  implicit def system: ActorSystem
-  implicit def timeout: Timeout
-  implicit def mat: Materializer
+object WebServer {
```
fommil: does this need to be a trait? Seems that having a singleton might not be good for testing.

server/src/main/scala/org/ensime/server/WebSocketFrameHandler.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  override protected def userEventTriggered(ctx: ChannelHandlerContext, evt: Object): Unit = {
+    if (HANDSHAKE_COMPLETE == evt) {
+      Future {
+        // Sorry about this tricky hack, it's the only way I found to access the handshaker
+        // associated with the channel so I can ask it for the selected subprotocol.
+        // The thing is that for the time when we receive this event the handshaker wasn't
+        // set as a channel attribute yet. So we MUST wait for it for a few millis, and for
+        // some reason we MUST do it in a separate thread.
+        // See: [WebSocketServerProtocolHandshakeHandler.java:88](https://git.io/v6DcJ)
```
fommil: lovely. I'm starting to prefer your idea of a system property...

fommil:
LGTM ignore my `object` comment, I didn't see that everything was in a `def`.

fommil:
If Vim (and, by extension Atom) works I'm happy for you to merge this once you've updated the CLA checker.

server/src/main/scala/org/ensime/server/WebSocketFrameHandler.scala
ktonga: I've opened an issue for it (netty/netty#5734), it shouldn't be that complicated. Anyways, it's not like they answer all my requests 😛

ktonga:
I'm having a hard time trying to make it backward-compatible for the editors using `/jerky`.

I tried with a redirect, but it's not supported in WS due to security reasons. It makes perfect sense for Javascript in the browser, but I have no idea why on earth it's not allowed in, for instance, the Python WS client.
And as I already knew, it is not possible to configure two WS protocol handlers in the same channel pipeline even if they have mutually exclusive URIs.

So editors will need to adapt to this little change for switching to server 2.0. That's what mayor version changes are for after all 😄 

I just want to properly handle any errors that might occur and add a few logging lines to know what's going on the web layer.
After that, we should be ok to merge it. It's almost ready and it still deletes more than 120 lines!!

fommil:
:+1: @hedefalk can you make the necessary Atom changes to coincide with this getting merged? @ktonga can you please summarise what is needed?

hedefalk:
Is that just s/jerky/ws for me?

fommil:
@hedefalk I think you need to send a header message too as part of the handshake

hedefalk:
Ok, I'll try to read up soon. If it's possible to get a simple client-todo-list-without-need-to-read-code, that'd be great :)

ktonga:
@hedefalk the only URI is now `/websocket` and you have to specify `jerky` or `swanky` as the desired subprotocol.

I've included an example in the [commit message](https://github.com/ensime/ensime-server/pull/1573/commits/05eb2bbf39eb08d53d3c800a90808458ba034f70) but I'll include it here for better visibility:

```python
ws = websocket.create_connection("ws://localhost:1234/websocket", subprotocols=["jerky"])
```

hedefalk:
@ktonga Great, thanks!

fommil:
:shipit:

ktonga:
I was testing the server with __ensime-vim__ as a sanity check before merging, and guess what, I found a bug in the implementation, and obviously related to that `Thread.sleep` nasty hack.

The first message sent by the editor isn't processed correctly since the handler isn't ready yet, and it turns out that the first messages is the most important one since without it the server will stay in `waitingConnectionInfoReq` state forever.
So it seems I have to implement a sort of stashing, which I will happily kill once I fix the [underlying problem](https://github.com/netty/netty/issues/5734) in Netty

ktonga:
Ok, I tested it in Vim and it's working now.
If you're still ok with the hack for the hack, I can squash and merge it.

fommil:
LGTM what size is the assembly jar now?

fommil:
Btw no need for manual squashing we have the github squash merge available. Fire when ready

fommil:
I think I feel a little tear coming.

fommil:
43MB, nice that just reduced our binary size by 14%

awesome stuff, thank you so much @ktonga !!

fommil:
there is version  4.1.2.Final -> 4.1.4.Final of netty, is there a reason we don't use it?

fommil:
damn, we've still not got rid of those parser combinators or scala-xml, I wonder what's keeping them?

fommil:
oh, it's the compiler. what a pain.

ktonga:
Yeah! I noticed that when I built one locally to test with Vim, Awesome.
> 43MB, nice that just reduced our binary size by 14%

No reason for the Netty version, I think I started before they released it. Will upgrade directly with my patch for the subprotocol simplification.

Dunno which XML you are referring to, I replaced them for the docs html page.

fommil:
scala-compiler depends on scala-xml and combinators. I wonder if we could turn those off and it still work

ktonga:
scala-xml seems to be mandatory only for tests and docs tools. The compiler does a [weird check](https://github.com/scala/scala/blob/d6f601d53e53be6f8071c0646c19b68faac69ec0/src/compiler/scala/tools/nsc/typechecker/Contexts.scala#L112) so I guess it's opcional.

ktonga:
scala-parser-blah is only used in tests. So it should be safe to exclude it.

rorygraves:
Nice!

rorygraves:
The merge, not parser combinatory ;)

ktonga:
The deps should be well scoped. Epic fail.

ktonga:
Yeah @rorygraves. It wasn't easy but we finally got there.

hedefalk:
I'm not to familiar with websockets, but is subprotocols something different than protocol? I'm trying to adapt ensime-atom and ensime-vscode

https://github.com/ensime/ensime-node/issues/20

but I get a 404 against 2.0 when doing this:

` new WebSocket("ws://localhost:" + httpPort + "/websocket", {protocol: "jerky"})``

Websocket lib I'm using is this: https://www.npmjs.com/package/ws. The API description there was limited but I installed typescript typings for it here http://definitelytyped.org/docs/ws--ws/classes/ws.websocket.html#constructor, and the only protocol thingie I could find was this single protocol string…

Anything pops up?

hedefalk:
Oh, the typings probably sucked: https://github.com/websockets/ws/blob/master/doc/ws.md

There's a protocols param there…

ktonga:
Try with the top-level param 'protocol', the second one, not the one inside 'options'

hedefalk:
@ktonga Yeah, thanks. The typings definition was wrong :)

PR #1574, opened 2016-07-28
Scala names in search

sugakandrey:
This is an attempt to make use of scalap information (scala names in particular) in `Indexer`. Note that scala names are not being indexed right now and are only used in search results. Files in `SearchService` are now grouped by their top level class/object (the one that has scala signature defined), also this fixes deadlock in `IndexingQueueActor` (thanks to @jschlather for bringing this up on gitter).

### DIFF
--- a/core/src/main/scala/org/ensime/core/ScalapSymbolToFqn.scala
+++ b/core/src/main/scala/org/ensime/core/ScalapSymbolToFqn.scala
@@ -8,8 +8,9 @@ import java.nio.charset.StandardCharsets
 import org.ensime.api.DeclaredAs
 import org.ensime.indexer._
 
-import scala.tools.scalap.scalax.rules.ScalaSigParserError
 import scala.tools.scalap.scalax.rules.scalasig._
+import scala.collection.{ breakOut, mutable }
+import scala.util.control.NonFatal
 
 trait ScalapSymbolToFqn {
   import ScalaSigApi._
@@ -22,7 +23,7 @@ trait ScalapSymbolToFqn {
       code(printer)
       new String(baos.toByteArray, StandardCharsets.UTF_8)
     } catch {
-      case e: ScalaSigParserError => ""
+      case NonFatal(e) => ""
     }
   }
 
@@ -30,6 +31,16 @@ trait ScalapSymbolToFqn {
     if (sym.isPrivate) Private
     else if (sym.isProtected) Protected
     else Public
+
+  def rawType(s: AliasSymbol, parentPrefix: String): RawType = {
+    val javaName = FieldName(className(s.symbolInfo.owner), s.name)
+    val scalaName = parentPrefix + s.name
+    val access = getAccess(s)
+    val typeSignature = withScalaSigPrinter { printer =>
+      printer.printType(s.infoType, " = ")(printer.TypeFlags(true))
+    }
+    RawType(javaName, scalaName, access, typeSignature)
+  }
 
   def rawScalaClass(sym: ClassSymbol): RawScalapClass = {
     val javaName = className(sym)
@@ -51,15 +62,22 @@ trait ScalapSymbolToFqn {
 
     val scalaName = aPackage + "." + name
     val parentPrefix = if (sym.isModule) scalaName + "." else scalaName + "#"
-    val fields = sym.children.collect {
+    val fields: Map[String, RawScalapField] = sym.children.collect {
       case ms: MethodSymbol if !ms.isMethod && ms.isLocal =>
-        rawScalaField(ms, parentPrefix)
-    }
+        val field = rawScalaField(ms, parentPrefix)
+        field.javaName.fqnString -> field
+    }(breakOut)
 
-    val methods = sym.children.collect {
+    val methods: mutable.ArrayBuffer[RawScalapMethod] = sym.children.collect {
       case ms: MethodSymbol if ms.isMethod && !ms.name.contains("default") && !ms.name.contains("<init>") =>
         rawScalaMethod(ms, parentPrefix)
-    }
+    }(breakOut)
+
+    val aliases: Map[String, RawType] = sym.children.collect {
+      case as: AliasSymbol =>
+        val alias = rawType(as, parentPrefix)
+        alias.javaName.fqnString -> alias
+    }(breakOut)
 
     RawScalapClass(
       javaName,
@@ -68,7 +86,8 @@ trait ScalapSymbolToFqn {
       access,
       declaredAs,
       fields,
-      methods
+      methods,
+      aliases
     )
   }
 
--- a/core/src/main/scala/org/ensime/core/ScalapSymbolToFqn.scala
+++ b/core/src/main/scala/org/ensime/core/ScalapSymbolToFqn.scala
@@ -31,7 +31,7 @@ trait ScalapSymbolToFqn {
     else if (sym.isProtected) Protected
     else Public
 
-  def rawScalaClass(sym: ClassSymbol): RawScalaClass = {
+  def rawScalaClass(sym: ClassSymbol): RawScalapClass = {
     val javaName = className(sym)
     val aPackage = sym.enclosingPackage
     val ownerChain = sym.ownerChain
@@ -57,11 +57,11 @@ trait ScalapSymbolToFqn {
     }
 
     val methods = sym.children.collect {
-      case ms: MethodSymbol if ms.isMethod =>
+      case ms: MethodSymbol if ms.isMethod && !ms.name.contains("default") && !ms.name.contains("<init>") =>
         rawScalaMethod(ms, parentPrefix)
     }
 
-    RawScalaClass(
+    RawScalapClass(
       javaName,
       scalaName,
       typeSignature,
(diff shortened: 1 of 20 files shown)


sugakandrey:
seems like I've broken some things, fixing it now

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -63,10 +63,18 @@ trait ClassfileIndexer {
...
-        Nil, Nil, RawSource(None, None)
+        Nil, Nil, RawSource(None, None),
+        isScala = false
       )
     }
 
+    override def visitAttribute(attr: Attribute): Unit = {
+      val attrType = attr.`type`
+      if (attrType == "Scala" || attrType == "ScalaSig") {
+        clazz = clazz.copy(isScala = true)
```
fommil: cool, is this more reliable than looking in the source file name? If debugging symbols or stripped it's hard to tell.

core/src/main/scala/org/ensime/indexer/stringymap/StringyMap.scala
```diff
@@ -95,6 +96,19 @@ package api {
       def toOrientProperty: OrientProperty = OrientProperty(OType.INTEGER)
     }
 
+    implicit object DeclaredAsSPrimitive extends SPrimitive[DeclaredAs] {
+      def toValue(v: DeclaredAs): java.lang.String = if (v == null) null else StringSPrimitive.toValue(v.toString)
+      def fromValue(v: AnyRef): DeclaredAs = StringSPrimitive.fromValue(v) match {
+        case "Method" => DeclaredAs.Method
```
fommil: might be a fancy shapeless way to do this by creating a `Map` from the instances keyed by their name.

fommil:
I'l merge now and review later... oooh the joys of the bleeding edge!

core/src/main/scala/org/ensime/core/ScalapSymbolToFqn.scala
```diff
@@ -51,24 +62,32 @@ trait ScalapSymbolToFqn {
...
       case ms: MethodSymbol if !ms.isMethod && ms.isLocal =>
-        rawScalaField(ms, parentPrefix)
-    }
+        val field = rawScalaField(ms, parentPrefix)
+        field.javaName.fqnString -> field
+    }(breakOut)
 
-    val methods = sym.children.collect {
-      case ms: MethodSymbol if ms.isMethod =>
+    val methods: mutable.ArrayBuffer[RawScalapMethod] = sym.children.collect {
```
fommil: mutable collection is gnarly, can we really not convert into an immutable version for the case class?

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -301,7 +368,7 @@ class IndexingQueueActor(searchService: SearchService) extends Actor with ActorL
   // De-dupes files that have been updated since we were last told to
   // index them. No need to aggregate values: the latest wins. Key is
   // the URI because FileObject doesn't implement equals
-  var todo = Map.empty[String, FileObject]
+  val todo = new m.HashMap[FileObject, m.Set[FileObject]] with m.MultiMap[FileObject, FileObject]
```
fommil: any reason you prefer a `val` of a mutable container? I think it's far better to have a `var` of an immutable structure as it makes it clear that the field is stateful.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -316,50 +383,53 @@ class IndexingQueueActor(searchService: SearchService) extends Actor with ActorL
...
               log.error(t, s"failed to remove stale entries in ${batch.size} files. $advice")
               retry()
             case Success(_) => indexed.foreach {
               case (file, syms) =>
                 val boost = searchService.isUserFile(file.getName)
-                val persisting = searchService.persist(FileCheck(file), syms, commitIndex = true, boost = boost)
+                val persisting = searchService.persist(syms, commitIndex = true, boost = boost)
 
                 persisting.onComplete {
                   case _ => searchService.semaphore.release()
```
fommil: it would be good to have some kind of feedback of when `.jar` files have been persisted... I can use it to report back how long it takes to re-index big jars at my work.

core/src/main/scala/org/ensime/indexer/stringymap/StringyMap.scala
```diff
@@ -95,6 +96,44 @@ package api {
...
+      }
+      object AdtToMap {
+        implicit def fromSingletonByName[A, C <: Coproduct](
+          implicit
+          gen: Generic.Aux[A, C],
+          singletonByName: SingletonByName[A, C]
+        ): AdtToMap[A] = new AdtToMap[A] { def map: Map[String, A] = singletonByName.map }
+      }
+
+      val map: Map[String, DeclaredAs] = implicitly[AdtToMap[DeclaredAs]].map
```
fommil: ha! nice

fommil: btw, did you see http://fommil.github.io/scalax15/#/6/2 ?

fommil:
I did a very quick scan and it looks good. Some general comments:

- we're now starting to really diverge from the 2.0 branch so we should be careful that nobody does big refactorings there without telling you. Please post a message onto, e.g. the SearchService refactor ticket, to this effect
- I need to set some time aside to fully understand your changes to the search service
- please send a short PR with fixes to the minor points (mostly around mutable data structures) above
- don't forget the logging message
- we forgot to bump the schema version again... anything that changes the on-disk format should have a schema version bump so that alpha testers don't have to manually nuke their cache.

then it's onto the next big thing! I'll try this out on a big project next week.

core/src/main/scala/org/ensime/indexer/stringymap/StringyMap.scala
sugakandrey: I saw Travis Brown question on SO, which I guess is the original version of `Values` in`shapeless/example/enum`.

core/src/main/scala/org/ensime/core/ScalapSymbolToFqn.scala
sugakandrey: yeah, I guess  it's a remnant of a previous iteration of this PR, should be fixed

core/src/main/scala/org/ensime/indexer/SearchService.scala
sugakandrey: I needed a multimap so mutable container was my only option, but since the is (afaik) no default way to split multimap into multimaps, I went with `val` and simple key removal.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -316,50 +383,53 @@ class IndexingQueueActor(searchService: SearchService) extends Actor with ActorL
...
 
       def retry(): Unit = {
-        batch.foreach(self !)
+        batch.valuesIterator.foreach(_.foreach(self ! IndexFile(_)))
       }
 
-      Future.sequence(batch.map {
-        case (url, f) =>
-          val filename = f.getName.getPath
+      batch.grouped(10).foreach(chunk => Future.sequence(chunk.map {
```
jschlather: Do we still need the semaphore if we're only processing 10 at a time? It seems to me that it could be removed completely at this point.

sugakandrey: We also make use of it in the `refresh()` route, where we index everything we have. We probably want backpressure there too.

fommil: this soooo needs to be rewritten to use monix / swave / csp. It's become a monster.

jschlather: It seems like somewhere along the way we made this way too hard. Why can't we just have an execution context with a fixed sized executor of 10 threads and a linked blocking queue?

fommil: You're welcome to try.

fommil: historically we didn't feel we needed backpressure, and retrofitting it was a quick hack (it's only really needed on huge projects). A fixed size executor would be a major rewrite because the current situation has different stages of the pipeline creating futures, and fanning in and out, so it wouldn't work without linearising all of that. Strategically we need to move this to a minimal streaming (or csp) framework that will be fast to support new versions of scala.

core/src/main/scala/org/ensime/indexer/SearchService.scala
fommil: oh this is going to cause problems... I just noticed now that you use `FileObject` as the key. Read the comment ;-) Definitely a bug in here.

fommil: as to multimaps, we have some multi-map like support in https://github.com/ensime/ensime-server/blob/2.0/util/src/main/scala/org/ensime/util/map.scala

PR #1578, opened 2016-07-31
Replace semaphore with a fixed size thread pool.

jschlather:
This replaces the semaphore implementation of backpressure with a fixed sized thread pool. So we're now using the global execution context to run the majority of our work and to register callbacks, while we use the fixed sized thread pool for indexing tasks.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -2,7 +2,7 @@
 // License: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.indexer
 
-import java.util.concurrent.Semaphore
+import java.util.concurrent.Executors
 
 import scala.concurrent._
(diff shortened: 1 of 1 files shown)


fommil:
I'm not sure I understand how this works... the problem that the semaphores fix is that the indexing is going much faster than the DB writing, so we just start growing the heap instead. Maybe I'm reading this wrong, but is this not just limiting the amount of indexer workers to 10 threads at a time? But it doesn't stop them when DB writing starts to slow down... so we can still overtake.

Indexing is much much faster than DB writing.

jschlather:
Yeah, I misunderstood the issue. I'll close this for now and reopen if I come up with a way to handle that as well.

fommil:
no problem, if you're interested in this area there are definitely a few things you could look at in the 2.0 milestone tagged with Indexer (swave/monix/csp or in-memory lucene) that would work really nicely with the graphpocalypse.

PR #1581, opened 2016-08-08
Resolve source in non-standard locations

VlachJosef:
This helps with resolving source files in projects which don't follow java notation of putting classes/objects/traits to the file structure matching package name.

[ReactiveMongo](https://github.com/ReactiveMongo/ReactiveMongo) is example of a such broken project. For example class [Authenticate](https://github.com/ReactiveMongo/ReactiveMongo/blob/0b168dc2ec27373cad88ace842de17a0488d72f8/driver/src/main/scala/core/nodeset.scala#L478) would get resolved correctly.


fommil:
I'm concerned about performance for this. Can we perhaps try the standard approaches first, then fall back to this? I also don't expect it to work for larger projects.

VlachJosef:
@fommil I've made a little refactoring. I've made sure that we test root package after all parents and just in case no hit is found "reverse" parents and others are tested.

fommil:
LGTM

PR #1583, opened 2016-08-10
fix the exception related to completion with an empty context

chessman:
Steps to reproduce:
1. Place the cursor here: object A { | }.
2. Ask for completion.
3. See an error "head of empty list" inside the ENSIME log.

### DIFF
--- a/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
@@ -449,6 +449,14 @@ class RichPresentationCompilerSpec extends EnsimeSpec
     ).
       foreach(test(_, cc))
   }
+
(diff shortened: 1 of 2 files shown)


fommil:
Cool! I'll review this over the next few days. Thanks for picking it up.

fommil:
LGTM, excellent thanks! Sorry it took me so long to review.

chessman:
Should I open an issue in Scala Issue Tracker? BTW, I had opened something ensime-related and it is scheduled to fix in 2.12:
https://issues.scala-lang.org/browse/SI-9885

fommil:
wow, what a terrible bug!!!

Always feel free to create tickets upstream. If you send a PR to ensime with a workaround for this upstream bug (assuming it's in 2.10 or 2.11) we'd accept that.

PR #1585, opened 2016-08-12
 duplicate fqns in lucene

sugakandrey:
This illustrates the problem with duplicate fqns in Lucene. 
@fommil maybe you could take a look at this at the hack day? I spend a fair amount of time trying to figure out how this happens, but I don't have much experience with Lucene and may be missing something obvious. If not, I'll continue with this next week.


sugakandrey:
We only had a dedicated analyzer for `fqn` field and `file` field was indexed using our fallback analyzer, which lowercases everything. Because of this our deletion queries never found any matching documents.

fommil:
doh! So this has **always** been broken? ROTFL

Issue #1437, opened 2016-05-02
use java.nio instead of java.io, VFS and File


fommil:
as part of the "move to Java 7" effort.

Will negate the need to do #1257 so we can just close that ticket if it doesn't make it into 1.0.

fommil:
https://github.com/fommil/ensime-server/blob/jar-urls/util/src/main/scala/org/ensime/util/ensimefile.scala

rumoku:
i'm picking it up.

fommil:
@rumoku hi, I added the utils to the 2.0 branch so you should be able to rebase and avoid conflicts now :smile: let me know if you need any help. I also forgot to give you some ENSIME stickers on Saturday, I'll do that next time.

fommil:
we're susceptible to this known JDK issue until we move over to NIO https://bugs.openjdk.java.net/browse/JDK-8024496

arigoldx:
@fommil is this that ticket you were talking about on gitter?

(Asking 'cause it mentions java.io vfs not apache vfs.)

fommil:
This is the one. There is a comma missing between java.io and vfs

arigoldx:
Woohoo! Thanks!

(:eyes:)

PR #1586, opened 2016-08-14
new utility methods, move some things around and remove deprecated uses

fommil:
preparation for #761 and decoupling with #1437

### DIFF
--- a/util/src/main/scala/org/ensime/util/path.scala
+++ b/util/src/main/scala/org/ensime/util/path.scala
@@ -14,9 +14,6 @@ import java.nio.file.{ Files, Path }
 package object path {
 
   implicit class RichPath(val path: Path) extends AnyVal {
-    // PathMatcher is too complex, use http://stackoverflow.com/questions/20531247
-    def isJava: Boolean = path.toString.toLowerCase.endsWith(".java")
-    def isScala: Boolean = path.toString.toLowerCase.endsWith(".scala")
     def exists(): Boolean = Files.exists(path)
     def readBytes(): Array[Byte] = Files.readAllBytes(path)
     def readString()(implicit cs: Charset): String = new String(readBytes(), cs)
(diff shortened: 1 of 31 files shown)


fommil:
@harsh FYI

util/src/main/scala/org/ensime/util/path.scala
```diff
@@ -0,0 +1,57 @@
...
+    def isScala: Boolean = path.toString.toLowerCase.endsWith(".scala")
+    def exists(): Boolean = Files.exists(path)
+    def readBytes(): Array[Byte] = Files.readAllBytes(path)
+    def readString()(implicit cs: Charset): String = new String(readBytes(), cs)
+
+    def /(child: String): Path = path.resolve(child)
+    def canon: Path = ???
+    def isFile: Boolean = path.toFile.isFile
+
+    /*
```
rorygraves: Is this commented code supposed to be in here?

fommil: yeah, it's a reminder about how to use the NIO API, we've to come back to this for the task to remove VFS.

rorygraves: Fair enough

rorygraves:
LGTM (obviously once it actually passes tests ;) )

fommil:
interesting, 

```
info]   java.lang.IllegalArgumentException: Illegal character in opaque part at index 13: jar:file://C:\Program Files\Java\jdk1.7.0\src.zip!\java\lang\String.java
```

fommil:
and the travis environment is a bit weird: 

```
 java.nio.file.FileSystemNotFoundException: /Library/Java/JavaVirtualMachines/jdk1.7.0_45.jdk/Contents/Home/jre/src.zip
```

fommil:
better...

```
ArchiveFile(C:\Program%20Files\Java\jdk1.7.0\src.zip,/) was not equal to ArchiveFile(C:\Program Files\Java\jdk1.7.0\src.zip,/) (EnsimeFileSpec.scala:44)
```

PR #1588, opened 2016-08-14
Migrate away from the Build trait. Fixes #1568

### LINKED ISSUE #1568: change our build to use .sbt files instead of Build.scala
**cough** @dwijnand

dwijnand:
This way preserves most code in the EnsimeBuild.scala file, which I'm guessing you'll probably prefer.


fommil:
Oh wow, cool! I'll try to review this soon. I guess we need a similar thing in ensime-sbt so we can publish for sbt 1.0.

PR #1589, opened 2016-08-16
[WIP] Persist reverse lookups

sugakandrey:
An attempt to persist reverse lookups information to optimise features such as find usages and refactor-rename across multiple files. Can be turned off by system property `ensime.index.no.reverse.lookups`.

# Known limitations

- Using class in self type annotation is not considered a usage atm (need to hack something with scalap to fix that)
- Initialising trait as an anonymous class doesn't always work
- If trying to find references to a field, we instead have to search for references to fqn of generated getter of this field

### DIFF
--- a/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
+++ b/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
@@ -296,12 +296,13 @@ class SearchServiceSpec extends EnsimeSpec
 
   it should "find usages of a regular class" in withSearchService { implicit service =>
     val usages = findUsages("org.reverselookups.MyException")
-    usages.length should ===(5)
+    usages.length should ===(6)
     usages.map(_.toSearchResult) should contain theSameElementsAs List(
       "Method org.reverselookups.MyException#<init>(): org.reverselookups.MyException",
       "Method org.reverselookups.ReverseLookups#throws(): scala.Unit",
       "Method org.reverselookups.ReverseLookups#catches(): scala.Int",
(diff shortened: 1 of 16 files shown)


fommil:
green light, I don't believe it :smile: that means I have to review it

core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -328,4 +400,9 @@ object SearchServiceTestUtils {
     case cdef: ClassDef => Set(cdef)
     case TypeHierarchy(cdef, typeRefs) => Set(cdef) ++ typeRefs.flatMap(hierarchyToSet)
   }
+
+  def findUsages(fqn: String)(implicit service: SearchService): List[FqnSymbol] = Await.result(service.findUsages(fqn), Duration.Inf).toList
+
+  // 2.10 scalap has slightly different formatting for method names
+  def unifyMethodName(s: String): String = s.replaceAll(" : ", ": ")
```
fommil: maybe heavyweight, but this could go into the `scala-2.10` source directory and then implement it per-scala version.

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -2,21 +2,21 @@
...
 package org.ensime.indexer
 
-import scala.collection.immutable.Queue
-
 import akka.event.slf4j.SLF4JLogging
 import org.apache.commons.vfs2.FileObject
 import org.objectweb.asm._
 import org.objectweb.asm.Opcodes._
 
+import scala.collection.immutable.Queue
```
fommil: this doesn't look like you used `C-c C-r o` :stuck_out_tongue:

Issue #1597, opened 2016-08-30
Add line, col for positions other than error report ranges (optionally?)


hedefalk:
I remembered incorrectly about the other ticket I had related to row,col positions, https://github.com/ensime/ensime-server/issues/1286. The current use case I have is not the same as before - before was just about row,col from errors/warnings which I guess might be easier.

In vscode, everything handling positions is row,col based. Specifically, I want to create a "DefinitionsProvider" which is used both for hovering and for hyper-clicking (jump to definition). https://code.visualstudio.com/docs/extensionAPI/vscode-api#DefinitionProvider

In ensime-atom I could get away my manually creating my own callback that was supposed to happen when clicking so I could jump to the correct row,line after opening a buffer and calculate row,col client side. 

In vscode however, the same API is used also for hovering which makes it basically impossible to do this. I would need to do my own file io and calculations client side to resolve an offset to row,col from file source all the time - anytime hovering with cursor on text. This I feel would be way too expensive. At least if this info is in any way easier calculated server-side.

Current WIP in vscode:
```
return instance.api.getSymbolAtPoint(document.fileName, document.offsetAt(position)).then((symbol) => {
               const pos = symbol.declPos
               // TODO: Gotta be a smarter way in ts to do this, at least localise in client code
               if(pos.typehint == 'OffsetSourcePosition') {
                   const offsetPos = <serverProtocol.OffsetSourcePosition> pos
                   // TODO: How to we resolve a character offset into a row,col position before we have the buffer?
                   const uri = vscode.Uri.file(offsetPos.file)
                   return new vscode.Location(uri, new vscode.Position(0, 0)) // FIXME:
                   })
               } else if(pos.typehint == 'LineSourcePosition') {
                   const linePos = <serverProtocol.LineSourcePosition> pos
                   return new vscode.Location(vscode.Uri.file(linePos.file), new vscode.Position(linePos.line, 0))
               }

```

So basically, I can only tell which file a symbol is defined, but can't do much more than that.

I guess its a bit more involved than the range case https://github.com/ensime/ensime-server/issues/1286 since there's this thing here:

```
sealed trait SourcePosition extends RpcResponse
final case class EmptySourcePosition() extends SourcePosition
final case class OffsetSourcePosition(file: File, offset: Int) extends SourcePosition
final case class LineSourcePosition(file: File, line: Int) extends SourcePosition
```

[...]

PR #1598, opened 2016-08-30
adds row,col to OffsetSourcePosition for kicks

hedefalk:
This isn't really a serious pull REQUEST at this point, but it felt like the easiest way to discuss code.

So the change to just add row, col seems pretty minor, but I guess protocol is a problem since we have named ADT:s for this. Probably doesn't make sense to add row, col to something named OffsetSourcePosition like I did here, but I wanted to try it all the way and code navigation now works in vscode for me.

I'm not too sure about how expensive package scala.reflect.internal.util.Position#calculateColumn() is either. Might be a problem if this is too expensive and used unnecessarily for clients that are offset based?

### DIFF
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -233,7 +233,8 @@ case object PosNeededYes extends PosNeeded
 
 sealed trait SourcePosition extends RpcResponse
 final case class EmptySourcePosition() extends SourcePosition
-final case class OffsetSourcePosition(file: File, offset: Int) extends SourcePosition
+final case class OffsetSourcePosition(file: File, offset: Int,
+  row: Option[Int] = None, col: Option[Int] = None) extends SourcePosition
 final case class LineSourcePosition(file: File, line: Int) extends SourcePosition
 
 final case class PackageInfo(
--- a/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
@@ -65,9 +65,9 @@ class JavaCompilerSpec extends EnsimeSpec
       val test1 = SourceFileInfo(new File(config.rootDir, "testing/simple/src/main/java/org/example/Test1.java"))
       val test2 = SourceFileInfo(new File(config.rootDir, "testing/simple/src/main/java/org/example/Test2.java"))
 
-      cc.askLinkPos(ClassName(PackageName(List("org", "example")), "Test2"), test2) should matchPattern { case Some(OffsetSourcePosition(f, 22)) => }
+      cc.askLinkPos(ClassName(PackageName(List("org", "example")), "Test2"), test2) should matchPattern { case Some(OffsetSourcePosition(f, 22, None, None)) => }
       cc.askLinkPos(ClassName(PackageName(List("org", "example")), "Foo"), test2) should matchPattern { case None => }
-      cc.askLinkPos(ClassName(PackageName(List("org", "example")), "Test2.Bar"), test2) should matchPattern { case Some(OffsetSourcePosition(f, 260)) => }
+      cc.askLinkPos(ClassName(PackageName(List("org", "example")), "Test2.Bar"), test2) should matchPattern { case Some(OffsetSourcePosition(f, 260, None, None)) => }
       //    cc.askLinkPos(JavaFqn("org.example", "Test2", Some("compute()")), test2) should matchPattern { case Some(OffsetSourcePosition(f, 58)) => }
 
     }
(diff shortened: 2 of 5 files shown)


hedefalk:
So this will use:

```
  private def calculateColumn(): Int = {
    var idx = source.lineToOffset(source.offsetToLine(point))
    var col = 0
    while (idx != point) {
      col += (if (source.content(idx) == '\t') Position.tabInc - col % Position.tabInc else 1)
      idx += 1
    }
    col + 1
  }

```

which then in turn uses:

```
  /** Convert offset to line in this source file.
   *  Lines are numbered from 0.
   */
  def offsetToLine(offset: Int): Int = {
    val lines = lineIndices
    def findLine(lo: Int, hi: Int, mid: Int): Int = (
      if (mid < lo || hi < mid) mid // minimal sanity check - as written this easily went into infinite loopyland
      else if (offset < lines(mid)) findLine(lo, mid - 1, (lo + mid - 1) / 2)
      else if (offset >= lines(mid + 1)) findLine(mid + 1, hi, (mid + 1 + hi) / 2)
      else mid
    )
    lastLine = findLine(0, lines.length, lastLine)
    lastLine
  }
```

At least the lineIndices are cached in a lazy val and I guessed this is used a lot internally anyways?

hedefalk:
https://github.com/ensime/ensime-server/issues/1597

hedefalk:
:tumbleweed:

Bump!

Yeah, I know this is a super-crappy PR that doesn't even pass CI tests, but maybe someone can talk to me about it? :)

No, but seriously, what's the best way to add optional rows,cols in protocol?

rorygraves:
Will have a look later. Kick me if I don't.

fommil:
yeah sorry for slow response, we're distracted with scala world prep at the moment

hedefalk:
Np!

fommil:
Sorry, we should really chat about this at some point now I'm back from hols.

fommil:
Only comment I have an this stage is I don't like adding optional stuff. If we add this, we make it mandatory.

hedefalk:
@fommil Yeah, ok. I'm just worried about performance. Do we have an easy test for this?

fommil:
No we don't have the hardware for performance tests. If it's noticeably slower after release, then we rethink. I don't care about performance until it hurts.

dragos:
Yeah, we need something like this. It's extremely annoying to compute this in each client, since you need the full sourcefile contents to find the line/col. This becomes tedious, especially for hyperlink targets that come from jars.

fommil:
once I get url-jars out of the way I can look into this PR... but if you want to tidy anything up in the meantime it would only speed things up :smile:

dragos:
Yeah, on the topic of positions: it'd be cool to keep range positions where possible. Sometimes it's useful to know the range of a symbol, rather than it's start position. For example, to show a "breadcrumb" based on where the cursor is or to highlight the region of a definition after "goto definition".

fommil:
@dragos when we find the position of a goto-source, it is often inferred from the debugging table in the `.class` files... so converting into `offset` from `line` number is already tricky. I'm not sure we could guarantee the end. If a source is loaded into the PC, then we know the full position and I would hope we're not dropping the info. There was a discussion at some point to load sources into the PC in such cases to get the position info, but I can't remember the conclusion. It might be very expensive since we include lots of these data points in search results. Perhaps we should persist information as we discover it.

dragos:
I meant for the common case where the source is in the PC. It looks like Ensime knows how to handled attached sources, so most of the times the source is there. I don't see when the classfile position is even useful, if there's no source to display?

fommil:
.class information is for when the sources are not loaded into the PC. We have a source resolver that finds the source files in the source dir/jar list. So consider the case where the user asks for the sources of `java.lang.String`, we have (at index time) worked out that the source lives in `String.java:99` and have a source resolver that links the FQN `java.lang.String` to `/your/jvm/src.zip!/java/lang/String.java`. So we're able to say exactly where the source is, without involving the PC. It also works for `.scala` files.

If we're dropping position information that we obtained from the PC, then that's a bug. It should be easy enough to write a test for that.

dragos:
Thanks for the explanation. I have to look into how hyperlinking is implemented in Ensime, but the PC has support for "goto definition". Note that for Scala it's impossible (in the general case) to know what's the bytecode name of a Scala source name (think name mangling), so you need to involve the PC anyway (and I believe you do, otherwise almost nothing would hyperlink).

Regarding the test, the behavior I mention is by design. You simply don't have the data structures to hold range positions, your concept of Position is limited to one point, not two (or three).

fommil:
We intentionally do not involve the PC, instead we have converters between bytecode name, java name, scalap name, scala language name and scala compiler internal name.

When you ask for the source of a type, we'll convert that into an FQN (so we're only using the PC on files that the user has open). And since everything has been indexed by FQN in the database, we can go look it up and get the source.

Our source resolver has some heuristics to handle scala sources that are not in the expected location (e.g. looking in parent packages). We won't need to do this in dotty, because TASTY will store relative paths (I wonder if scalac could be retrofitted to have this information?).

It would be far too inefficient to involve the PC with every single source file in a project. I've only worked on huge projects (2+ million LOC) for the last 4 years, so that's the main reason why it's been designed this way.

fommil:
that makes it sound like we completely ignore the PC... that's not true. Before we look up in the Indexer / SearchService, we ask the PC if it knows the source of the symbol/type. Then we fallback to looking into the db.

Lucene is only used to handle user searches, we rely on traditional DB indexing (and soon graph indexing) for fast FQN lookup.

dragos:
yeah, the PC has support to locate a symbol **given you provide the source file where it's defined**. So it's good you have that Lucene index. I'm just saying that *inside that file* the PC will likely do a better job than the bytecode. And you'll have to present it to the user anyway, so it's ok to ask the PC about it.

Yeah, performance is important, and I think Ensime could do a lot better ;-)

dragos:
In case I wasn't clear, I'm not suggesting you load everything into the PC. I'm actually trying to unload things when they're not needed, but it seems Ensime will only get support for that in the next release

fommil:
I think you missed my last note about trying the PC first, and what we use Lucene for :wink:

hehe, yeah, it's silly that we don't unload files. I'd even noticed it, but sometimes when you're "in the flow" you just want a quick hack then forget to file a bug report. My workaround in Emacs is to use `C-c C-c r` when anything goes wrong with the typechecker, which basically unloads everything, and starts again with my active emacs set. It's a general cure-all, but we can do better.

fommil:
we've gone totally off topic now, so we may as well continue this :wink: @smarter there is an explanation above about how ensime does "go to source definition" and that we use the PC if possible, falling back to our index. It would be good to make sure that dotty still provides this API and that you don't try to solve the larger problem that we address.

dragos:
Yeah, totally off topic, and yeah, I missed the part about "user-searches". Cool. The weekend is over, so my ensime hacking time is over too, but I made good [progress](https://github.com/dragos/dragos-vscode-scala/releases) on my vscode plugin. I still have some things to iron out but I might be able to use it already on some "real" projects.

dragos:
Also worth keeping an eye on Microsoft/language-server-protocol#96

fommil:
I still want to get into this PR... just been crazy busy

fommil:
do you mind doing a rebase please?

fommil:
wow, almost 5 months and I am still utter rubbish and haven't gotten round to this. If you have any time to hack on this, I'll make the time to look at it in advance of that.

fommil:
admitting defeat

hedefalk:
Crap. I'll reopen! Any year now.

Den sön 11 juni 2017 15:06Sam Halliday <notifications@github.com> skrev:

PR #1602, opened 2016-09-18
[WIP] Use reverse lookups in find usages and refactor rename

sugakandrey:
This should make it possible to only load file, which are actually affected by the refactoring, into pres. compiler. It almost works, except sometimes it doesn't (looks like a race condition or some similar error), should be all good when I fix it.

### DIFF
--- a/core/src/it/scala/org/ensime/intg/ReverseLookupsSpec.scala
+++ b/core/src/it/scala/org/ensime/intg/ReverseLookupsSpec.scala
@@ -9,8 +9,6 @@ import org.ensime.util.EnsimeSpec
 import org.ensime.util.ensimefile.Implicits.DefaultCharset
 import org.ensime.util.file._
 
-import scala.util.Properties
-
 class ReverseLookupsSpec extends EnsimeSpec
     with IsolatedEnsimeConfigFixture
     with IsolatedTestKitFixture
@@ -37,28 +35,6 @@ class ReverseLookupsSpec extends EnsimeSpec
         )
       }
     }
-  }
-
-  it should "only search files loaded into pres. compiler when reverse lookups are turned off" in {
-    try {
-      Properties.setProp("ensime.index.no.reverse.lookups", "true")
-      withEnsimeConfig { implicit config =>
-        withTestKit { implicit testKit =>
-          withProject { (project, asyncHelper) =>
-            import testKit._
-            val sourceRoot = scalaMain(config)
-            val fooFile = sourceRoot / "org/example/Foo.scala"
-
-            project ! UsesOfSymbolAtPointReq(Left(fooFile), 119)
-            val uses = expectMsgType[ERangePositions]
-            uses.positions.map(usage => (s"${File(usage.file).getName}", usage.offset, usage.start, usage.end)) should contain theSameElementsAs List(
-              ("Foo.scala", 114, 110, 172),
-              ("Foo.scala", 273, 269, 283)
-            )
-          }
-        }
-      }
-    } finally Properties.clearProp("ensime.index.no.reverse.lookups")
   }
 
   "Refactor Rename" should "make use of reverse lookups information" in withEnsimeConfig { implicit config =>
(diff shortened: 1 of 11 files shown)


fommil:
I'm on holiday and can't really review, but barring the red lights, let me just say OMG OMG OMG WAAAAHHHHHHH!

sugakandrey:
@fommil Alright, so this is almost good (how many times have I said this already?). This should go green as soon as I figure out what's wrong with refactor-rename, find usages part seem to be working.

rorygraves:
👍

sugakandrey:
finally! :)

fommil:
:open_mouth: wow

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -187,6 +187,16 @@ trait ClassfileIndexer {
         case member => FieldName(ClassName.fromInternal(owner), member)
       }
 
+    private def extractClassNames(className: String): Seq[String] = {
```
fommil: I'm not entirely sure what this method does, but it seems like it might be a memory hog.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -86,7 +86,7 @@ class SearchService(
   // poor man's backpressure.
   val semaphore = new Semaphore(Properties.propOrElse("ensime.index.parallel", "10").toInt, true)
 
-  private val noReverseLookups = Properties.propOrFalse("ensime.index.no.reverse.lookups")
+  def noReverseLookups: Boolean = Properties.propOrFalse("ensime.index.no.reverse.lookups")
```
fommil: this might be best as a `val` for performance reasons

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
sugakandrey: Let's say we have 
```
package org.foo

object Foo {
  class Inner {} 
}
```
 in Foo.scala and
```
object Bar extends App {
  val a = new Foo.Inner()
}
```
 in Bar.scala. The way we had it before is `val a` only had a reference to `org.foo.Foo$Inner`. Now suppose we want to rename `object Foo` we lookup the usages and `val a` won't come up, because it "uses" `org.foo.Foo$Inner` not `org.foo.Foo$`. In short we fail to detect any uses of outer classes when their inner classes are called, which is obviously faulty. This is sort of a hacky way to achieve what we want here (i.e. extract all possible names of outer classes).

core/src/main/scala/org/ensime/indexer/SearchService.scala
sugakandrey: yeah, this totally should be a `val` can't even remember why I changed it :)

core/src/it/scala/org/ensime/intg/ReverseLookupsSpec.scala
```diff
@@ -102,39 +104,41 @@ class ReverseLookupsSpec extends EnsimeSpec
...
-                    |+    def notATestMethod(i: Int, s: String) = {
-                    |       i + s.length
-                    |@@ -16,3 +16,3 @@
-                    |   println("Hello, " + foo.x)
-                    |-  println(foo.testMethod(7, "seven"))
-                    |+  println(foo.notATestMethod(7, "seven"))
-                    | 
+  it should "only search files loaded into pres. compiler when reverse lookups are turned off" in {
+    try {
+      Properties.setProp("ensime.index.no.reverse.lookups", "true")
```
fommil: this is tricky because while this test is running, other tests in the same JVM will not do reverse lookups. I'm ok with this being untested.

fommil:
awesome the new review feature on github is really good

fommil:
@rorygraves I've removed LGTM from this repo and enabled required reviews, it really seems like github are listening to the community!

ktonga:
@fommil I've proposed the very same for ensime-vim a while ago ensime/ensime-vim#338

core/src/it/scala/org/ensime/intg/ReverseLookupsSpec.scala
sugakandrey: nice catch, didn't think about it

PR #1607, opened 2016-10-11
Barrel of monkeys

fommil:
improve the scala monkey patch


fommil:
emacs tests failing consistently, I wonder if they've gone stale... *sigh*

PR #1610, opened 2016-10-14
outer class references in graph

### LINKED ISSUE #1609: persist outer class references
Currently we use an ugly hack, written by me, but as @fommil brought up we could simply get this info from the graph at query time. We even have an edge class for that (`OwningClass`), which should just be generalised to `EdgeT[FqnSymbol, ClassDef]`.

sugakandrey:
Pretty straightforward, closes #1609.


sugakandrey:
@fommil Travis is at it again, but the PR seems to be functioning alright. Should I merge this?

PR #1611, opened 2016-10-15
do not remove newlines from messages in notes

### LINKED ISSUE #1604: Ensime notes should not remove newlines
Eclipse is notorious for not handling multi-line error messages, but it shouldn't force everyone else to follow the same limitation.

Ensime removes newlines from Scala errors, even though some clients (like VS Code) would render them just fine. This is particularly annoying for "type mismatch" errors.

If this is a problem in existing clients, maybe a new field could be added, "rawMessage".

KimStebel:
fixes #1604


fommil:
Windows tests failed to run but GNU/Linux and OSX are looking good. I'll close/reopen to try and force a build.

PR #1612, opened 2016-10-15
Jar urls

### LINKED ISSUE #761: don't unpack jars/zip files

fommil:
This is a joint PR with @hardmettle and requires a detailed review. @rorygraves PTAL

I want to have another good look at it because it's been going on for so long.

Close #761

@hedefalk / @dragos I suspect you'll both want to add `-Densime.legacy.jarurls=false` when starting the server to get the new (much faster) behaviour

@dickwall / @ktonga you might want to do something like this in Vim.

I'd love to do something in Emacs but it seems like it could be a pretty big refactor. I should really spend more time on the emacs lisp code and less on the server.

### DIFF
--- a/api/src/main/scala/org/ensime/api/common.scala
+++ b/api/src/main/scala/org/ensime/api/common.scala
@@ -93,8 +93,8 @@ object RefactorType {
 }
 
 final case class SourceFileInfo(
-    file: File,
-    contents: Option[String] = None,
+    file: EnsimeFile,
+    @deprecating("prefer contentsIn, contents will be removed") contents: Option[String] = None,
     contentsIn: Option[File] = None
 ) {
   // keep the log file sane for unsaved files
--- a/api/src/main/scala/org/ensime/api/common.scala
+++ b/api/src/main/scala/org/ensime/api/common.scala
@@ -92,9 +92,21 @@ object RefactorType {
   def allTypes = Seq(Rename, ExtractMethod, ExtractLocal, InlineLocal, OrganizeImports, AddImport)
 }
 
+/**
+ * Represents a source file that has a physical location (either a
+ * file or an archive entry) with (optional) up-to-date information in
+ * another file, or as a String.
+ *
+ * Clients using a wire protocol should prefer `contentsIn` for
+ * performance (string escaping), whereas in-process clients should
+ * use the `contents` variant.
+ *
+ * If both contents and contentsIn are provided, contents is
+ * preferred.
+ */
 final case class SourceFileInfo(
     file: EnsimeFile,
-    @deprecating("prefer contentsIn, contents will be removed") contents: Option[String] = None,
+    contents: Option[String] = None,
     contentsIn: Option[File] = None
 ) {
   // keep the log file sane for unsaved files
--- a/api/src/test/scala/org/ensime/api/EnsimeTestData.scala
+++ b/api/src/test/scala/org/ensime/api/EnsimeTestData.scala
@@ -62,7 +62,7 @@ trait EnsimeTestData {
   val sourcePos4 = new OffsetSourcePosition(file1, 456)
 
   val breakPoint1 = new Breakpoint(file1.file.toFile, sourcePos1.line)
-  val breakPoint2 = new Breakpoint(file2.file.toFile, sourcePos2.line)
+  val breakPoint2 = new Breakpoint(file1.file.toFile, sourcePos2.line)
 
   val breakpointList = BreakpointList(List(breakPoint1), List(breakPoint2))
 
--- a/api/src/test/scala/org/ensime/api/EnsimeTestData.scala
+++ b/api/src/test/scala/org/ensime/api/EnsimeTestData.scala
@@ -6,12 +6,13 @@ import java.io.File
 
 trait EnsimeTestData {
   // duplicating utils to minimise dependencies
-  private def canon(s: String) = {
+  private def canon(s: String): RawFile = {
     val file = new File(s)
-    try file.getCanonicalFile
+    val canonised = try file.getCanonicalFile
     catch {
       case t: Throwable => file.getAbsoluteFile
     }
+    RawFile(canonised.toPath)
   }
 
   val typeInfo = BasicTypeInfo("type1", DeclaredAs.Method, "FOO.type1", Nil, Nil, None, Nil)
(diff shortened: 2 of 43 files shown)


api/src/main/scala/org/ensime/api/common.scala
```diff
@@ -93,8 +93,8 @@ object RefactorType {
 }
 
 final case class SourceFileInfo(
-    file: File,
-    contents: Option[String] = None,
+    file: EnsimeFile,
+    @deprecating("prefer contentsIn, contents will be removed") contents: Option[String] = None,
```
dragos: But if you remove `contents` how will clients work with in-memory files? Right now this is used, for instance, to ask for completions or for type-check requests.

dragos: You mean you write it in a file system file?

dragos: To expand a bit: faster than what?

rorygraves:
Sending it over the wire. I suspect is because the code that escapes the source was slow.

api/src/main/scala/org/ensime/api/common.scala
fommil: oh, but you're not using wire protocols! Ah, well then you've got a legitimate use case for this and I'll remove the `@deprecating` and replace it with a comment explaining the intended uses. I still don't like that it is two `Option`s... it should really be a sealed family, otherwise what do we do when both are `Some`? :-/

dragos:
Good point about escaping. BTW, I don't know if this was ever a requirement, but requiring a local file means ensime won't work across the network.

In the case of VScode, we get document change events with the file contents, so we already pay the price of escaping/unescaping, but I have no idea how expensive it is.

rorygraves:
Yep, its not a requirement, I've always wanted to support proper remote 
support, but it needs some more work to make it actually function.

fommil:
trigger/continue is failing. I thought we fixed that.

api/src/test/scala/org/ensime/api/EnsimeTestData.scala
```diff
@@ -53,15 +54,15 @@ trait EnsimeTestData {
...
+  val refactorDiffEffect = new RefactorDiffEffect(9, RefactorType.AddImport, file2.file.toFile)
 
   val sourcePos1 = new LineSourcePosition(file1, 57)
   val sourcePos2 = new LineSourcePosition(file1, 59)
   val sourcePos3 = new EmptySourcePosition()
   val sourcePos4 = new OffsetSourcePosition(file1, 456)
 
-  val breakPoint1 = new Breakpoint(sourcePos1.file, sourcePos1.line)
-  val breakPoint2 = new Breakpoint(sourcePos2.file, sourcePos2.line)
+  val breakPoint1 = new Breakpoint(file1.file.toFile, sourcePos1.line)
```
rorygraves: ```file1.file.toFile``` really does not make the intention clear here - is there a better way to refactor this to capture the effective ```soucePos1.file```?

fommil:
weird (non-transient) failure on OS X:

```
 java.lang.IllegalArgumentException: Not supported: WrappedJavaFileObject[org.ensime.core.javac.NotSoSimpleJavaFileObject@55936b13]
[info]   at com.sun.tools.javac.file.JavacFileManager.isSameFile(JavacFileManager.java:647)
[info]   at com.sun.tools.javac.comp.Check.checkForBadAuxiliaryClassAccess(Check.java:3471)
[info]   at com.sun.tools.javac.comp.Attr.checkIdInternal(Attr.java:3546)
[info]   at com.sun.tools.javac.comp.Attr.checkId(Attr.java:3489)
[info]   at com.sun.tools.javac.comp.Attr.visitSelect(Attr.java:3370)
```

hardmettle:
I was developing on Mac OS X all this while .

fommil:
Could be the difference between openjdk and proprietary sun jdk :-/

fommil:
I downloaded horrible proprietary oracle jdk8 to run the test and couldn't reproduce.

fommil:
oh, not true! I reproduce the javacompilerspec failure with the horrible oracle jdk. The other failure looks transient. Will try with OpenJDK, where we can see the sources of the impl that fails.

fommil:
ok, it's a java 8 thing. I can reproduce with openjdk. That means this is a blocker for scala 2.12

fommil:
digging into this will involve obtainng the java8 openjdk-langtools sources, they aren't shipped in src.zip

fommil:
I've added support for the langtools sources and pushed to this PR, that will at least let us investigate. However, the langtools sources I've attached might be old so I might need to recreate them, which is painful and involves using mercurial.

fommil:
relevant failing method is

```java
    public boolean isSameFile(FileObject a, FileObject b) {
        nullCheck(a);
        nullCheck(b);
        if (!(a instanceof BaseFileObject))
            throw new IllegalArgumentException("Not supported: " + a);
        if (!(b instanceof BaseFileObject))
            throw new IllegalArgumentException("Not supported: " + b);
        return a.equals(b);
    }
```

so looks like we need to implement `BaseFileObject` for this to work on java 8.

fommil:
I can't get it to work on java 8, even by implementing that class... seems it is getting wrapped by a `WrappedJavaFileObject`

fommil:
@mwielocha do you think you'd be able to help here? :sob:

mwielocha:
@fommil I'll be happy to take a look, just that right now it might take me some time to do it - my daily job is keeping me super-busy atm :/

hardmettle:
@fommil it needs a `JavacFileManager` `protected BaseFileObject(JavacFileManager var1) {
        this.fileManager = var1;
    }` . Shall I implement it ?

mwielocha:
@hardmettle please do, I don't think I'll have any time to spare this week :(

fommil:
total magic! :smile:

hardmettle:
@fommil I made some changes , but for some reason I cant push/clone my fork repository of ensime-server :\ . It says permission denied.

fommil:
Did you make sure to add your remote with ssh? Post the diff here and i can pick it up. Looking forward to this one!

hardmettle:
@fommil Did an initial commit after fixing the ssh issue. Left the overridden methods unimplemented . Take a look,review and I will implement them as well.

fommil:
sorry @hardmettle I forgot I was supposed to pick from your branch... please do continue on and get this test passing on java 8. Raise a PR from your repo and that'll let us easily see test failures and I can see when you make updates.

fommil:
@hardmettle you've put a TODO into the commit, so it won't run anything in CI

hardmettle:
@fommil forgot to mention , I checked in improved version of my implementation . Can you take a quick look because I forgot the last test we were trying to make green :(

fommil:
Please raise a PR

PR #1619, opened 2016-10-22
Remove FormatOneSourceReq and FormatSourceReq messages

### LINKED ISSUE #1606: Remove scalariform
If we create a small sbt Task to format a single file we can remove the formatting api calls. The emacs commands could be transparently changed to send the relevant sbt command instead. We could provide the task under ensime-sbt until upstream take it.

yyadavalli:
Resolve #1606


yyadavalli:
`formattingPreferences` can also be removed from `EnsimeConfig`but it will affect a every client

fommil:
waiting to hear back from

- [ ] @hedefalk 
- [ ] @ktonga
- [ ] @dickwall / @rorygraves
- [ ] @dragos

that this change won't blow up the various 2.0 clients.

yyadavalli:
I am trying open or respond to issues in all clients as I find them

dragos:
I don't understand the rationale here. I'm sure Sbt (besides locking people into one build tool) adds significant overhead. I use formatting (single file, or sometimes just a selection) very often, and I'd hate to wait for several seconds each time.

fommil:
@dragos ensime-server doesn't support selection, but if scalariform supports this maybe the sbt task could support it too.

The main rationale is to remove scalariform from the server build because it complicates our .ensime format and means we are tied to scalariform making releases for new versions of scala (it is a dead dead project, and I've wasted a lot of hours/days trying to keep even 2.11 alive). Also, we'd really prefer scalafmt if given the choice but it has its own dependency hell.

Additional advantages of moving this to the build tool is that there are no more incompatibilities with sbt-scalariform and we now have a clear upgrade path to scalafmt that isn't blocked on cross scala builds of `scala.meta`.

I've been trialing it for a few days and, even on huge projects on Windows, it is comparable in performance to the ensime-server implementation. Actually, in emacs the change should be seamless (other than the rebinding of the key, but we could always add a compatibility key).

Also, it means the user gets formatting even when the server is not up, which lends itself well to the principle of least power.

fommil:
btw, this isn't locking into one build tool... adding support in gradle / maven shouldn't be too hard either. I'm unlikely to do that work myself because I don't use either tool.

fommil:
@yyadavalli actually, that's a good point... can you please send a pr to ensime-emacs with the old key bound to the same action, perhaps with a popup on first use saying "this will be rebound". Maybe even make that go through the old function name so if anybody (I'm thinking re-bundlings of emacs) has bound their keys, they'll get the popup too.

dragos:
I know the trouble with scalariform, and it's been a "drag" on Scala IDE too. Not too much though, I don't remember the last time there was any problem with building it for a new version of Scala. It has zero dependencies, and that helps. So as along as Scala IDE is around, scalariform will be on life support and will work.

The advantage of having Scalariform in the build is that you can have formtting as a "function" without side-effects through the file-system. It should return back a list of edits that can be applied on the editor. Then you can limit that to a selection range (that's what Scala IDE does.

I understand the problem with two kinds of settings, but editors like VS Code expect a formatting function too ([returning text edits](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#document-formatting-request)). I think it would be a pity to remove this functionality.

fommil:
@dragos can you easily add non-ensime endpoints to your API? Since this one is pretty simple to implement, maybe you'd want to support formatting through the language server API.

dragos:
Yeah, that's what I realized after my comment. Other clients will have a hard time implementing it though.

fommil:
I'm not sure, I think closer build tool support is important. `ensime-server` is not trying to be a full blown IDE, but perhaps the language server is? In which case adding an interface for "build" related things might be something you'll find yourself increasingly wanting to add. Formatting is one of those weird ones that's on the boundary of whether the build or text editor owns it. I'm veering towards the approach of "if the build tool can do it, then the ensime server shouldn't duplicate the effort"...

fommil:
@dragos do you know about the sbt-server initiative? If clients aren't happy talking to the sbt console, that's a JSON interface they can use.

dragos:
I don't know anything about sbt-server. Honestly, I'm trying to rely on Sbt as little as possible and run away to something else as soon as I get the chance (realizing that might never happen, but once can only hope).

dragos:
I'm putting this here as food for thought, as I don't feel too strongly about this (only because I can work around it in my language server ;-)).

A scenario that I stumbled upon many times: working on someone else's project that doesn't have automatic formatting in the build. Fixing a bug or adding a feature in an existing file should not re-format the whole file, but at the same time I enjoy formatting automatically my own code. So there I rely exclusively on IDE/editor support for range-formatting. I step on this way more often than projects with a build-step formatter.

fommil:
yes, I agree. My work project doesn't have a build step formatter, for example, so I also rely on ensime to do the formatting (although I can only do this on files that I own otherwise I reformat other people's code). the `ensimeScalariformOnly` works well there too. If I was using other build tools, I'd bring them up to parity with `sbt-ensime` but I'm just not motivated to do it. I don't think it would be much work for a contributor. My mentality is very much that I'll fix something if I need to and I am quite demanding that I expect this of others too (unless they are willing to trade tickets! I'd gladly bring gradle-ensime up to speed if somebody fixed all the pc issues I experience).

fommil:
@yyadavalli you forgot to remove scalariform from the build dependencies and `.ensime` format :stuck_out_tongue:

yyadavalli:
I thought we are waiting on the `.ensime` changes, that affects all clients.

Scalariform is still need to implement expand region 
https://github.com/ensime/ensime-server/blob/a66bcbbb88e690c7be46cad753399d3cab2dd0f9/core/src/main/scala/org/ensime/core/Refactoring.scala#L76

fommil:
argh. I did not realise that.

fommil:
ok, I think we need to reimplement the expand region in that case

fommil:
we can do this as part 1 of removing scalariform

yyadavalli:
Ok, i'll push the code to remove formattingPreferences and open an issue for expand region for now.

fommil:
cool, and I'll add hacktoberfest to it... sorry the issues are growing!

yyadavalli:
Thanks. Don't worry about the growing issues, I'd rather work on it than leave gaps.

hedefalk:
+1 on range formatting for the same reason as Iulian.  I very seldom format
full files.

fommil:
@hedefalk I don't think scalariform exposes this. I'd also love it... I wonder if @olafurpg has anything to say on that :wink: 

Olaf, FYI, we've ripped out scalariform from ensime-server in preference of closer integration with sbt commands to do the same. A `scalafmt`  sbt command to do the same thing would be awesome, and could allow seemless migration... but do you support reformatting a range?

olafurpg:
Formatting a range has been on the road map for awhile. The cli has had a prototype implementation (since ca v0.2)  but it only work in simple cases. I'm hoping to get range formatting properly working v0.5 (next minor). I want to be able format only a diff for PRs, would be useful.

dragos:
If you return a set of edits instead of a full replacement, pretty-printed, text, this would be trivial to support. I believe that's how scalariform works: it returns a set of text edits for the whole file, but clients can filter those edits based on a formatting range.

fommil:
scalariform also offers the ability to rewrite the file with the change... it would be good to have that as an option too, to make it trivial to implement an sbt task that takes `(file, start, end)`

olafurpg:
@dragos would that still work for cases like this?

```scala
// maxColumn      |
// before
function(a,
         ccccccccccccc, // <- format only this line
         d)
// scalafmt output
function(
  a,
  ccccccccccccc,
  d)
// filter by range
function(a,
  ccccccccccccc,
         d)
```

A lot of things work differently when you have a maxColumn setting, which scalariform doesn't have.

I think we need a step to expand the range to the enclosing statement so the while `function(...)` call is formatted even if the range only covers a single argument.

dragos:
You're probably right in your example, but I think that's the correct behavior and exactly what I expect to happen.

Users can expand selection (another feature in IDEs based on scalariform) if they're not happy. It's really easy, I've done it countless times.

I don't think we can assume people want to reformat the entire file all the time. The common case is the other way around.

olafurpg:
I see, I guess that behavior is OK in an interactive IDE session, although it might surprise some users. I want this feature to format a diff before review, which can be enforced in CI. In that case, the behavior from my example would be unacceptable.

olafurpg:
Just for fun, I came across another example where filtering formatted output by range might produce surprising output

```scala
lst.map {
  x =>
    foobar(a, b) // <- format this line
}
// scalafmt
lst.map { x =>
  foobar(a, b)
}
// filtered by range
lst.map {
  x =>
  foobar(a, b)
}
```

dragos:
Sure, there are cases where this can lead to surprising behavior. I could argue that's precisely what the user wants, and why he selected only that line: because his indentation style isn't understood by scalafmt, and he *really* wants that `x =>` to stay where it is. It's the necessary escape hatch, and of paramount importance when working on other people's code.

fommil:
formatting this way is exactly what I'd have expected, and indeed is what the emacs scala-mode indenter does (which is very basic).

fommil:
now, if you'd asked to format the Position between the `{` and the `}`, that's a different thing... and is perhaps more what I would use this tool for. Emacs / ensime makes it very easy to select increasing scopes of blocks of code.

olafurpg:
@fommil You mean it produces this output?

```
lst.map {
  x =>
  foobar(a, b)
}
```

My current approach is to try and be a bit smarter and use the existing source to guess the correct indentation at the start of the range.

PR #1620, opened 2016-10-22
bump sbt-ensime

fommil:


fommil:
some weird errors here...

```
should support a wide range of queries *** FAILED ***
[info]   Some("docs/scala-library-2.11.8-javadoc.jar/index.html#scala.Some@map[B](f:A=>B):Option[B]") was not equal to Some("docs/scala-library-2.10.6-javadoc.jar/index.html#scala.Some@map[B](f:A=>B):Option[B]") (DocResolverSpec.scala:26)
```

could be that our build is not as clean as it should be.

fommil:
ha! we had `projectSettings` instead of `buildSettings`

PR #1624, opened 2016-10-25
Remove flags for watch disabling and source mode from EnsimeConfig

### LINKED ISSUE #1617: remove watch disabling from the .ensime
It is inconsistent to have server features mixed up with project configuration. Same goes for the "source mode" setting. This should be enabled by system properties for now and #1346 later.

yyadavalli:
Use system properties instead of `.ensime` flags

I'm not sure regarding the unit test, I don't think it is of any use now

Close #1617

### DIFF
--- a/core/src/test/scala/org/ensime/config/EnsimeConfigSpec.scala
+++ b/core/src/test/scala/org/ensime/config/EnsimeConfigSpec.scala
@@ -78,28 +78,4 @@ class EnsimeConfigSpec extends EnsimeSpec {
     })
   }
 
-  it should "base class paths on ensime.sourceMode value" in withTempDir { dir =>
-    val abc = dir / "abc"
-    val cache = dir / ".ensime_cache"
-    val javaHome = File(Properties.javaHome)
-
-    abc.mkdirs()
-    cache.mkdirs()
-
-    test(dir, s"""
-(:name "project"
- :scala-version "2.10.4"
- :java-home "$javaHome"
- :root-dir "$dir"
- :cache-dir "$cache"
- :source-mode ${if (Properties.propOrFalse("ensime.sourceMode")) "t" else "nil"}
- :subprojects ((:name "module1"
-                :scala-version "2.10.4"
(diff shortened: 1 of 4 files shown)


fommil:
I'll need to look at this one in some detail because my description of it in the ticket might not have been top quality. We still want to watch sources for when they are created or deleted, to update the source resolver. But we no longer want to watch sources to cause re-typechecking or updates to the source resolver unless the existence of the files has changed.

yyadavalli:
I thought all you wanted to do is move the flags to disable watching from `EnsimeConfig`  to system properties (or environment variables), so that later they can be part of a `ServerConfig`

fommil:
If that's all you're doing here then that's great. I wanted to remove the unnecessary work too, so maybe create another ticket.

core/src/test/scala/org/ensime/config/EnsimeConfigSpec.scala
```diff
@@ -79,33 +78,28 @@ class EnsimeConfigSpec extends EnsimeSpec {
...
                 :targets ("$abc"))))""", { implicit config =>
-          config.sourceMode shouldBe sourceMode
-          config.compileClasspath shouldBe (
-            if (sourceMode) Set.empty else Set(abc)
-          )
-        })
-      }
-    }
+      config.compileClasspath shouldBe (
+        if (Properties.propOrFalse("ensime.sourceMode")) Set.empty else Set(abc)
```
yyadavalli: In that case, there is nothing this test would cover that's not already covered and can be deleted.

fommil: :+1:

fommil:
weird, the `Test interactive search` emacs test failed. I can't see why it should be impacted :confused:

yyadavalli:
Hmm, weird, I just checked and I have `Test inspector ui.` failing

fommil:
That's probably because you're using Java 8.

PR #1627, opened 2016-10-26
Disable typechecking when files are changed

### LINKED ISSUE #1625: Limit source watching to creation and deletion of files
Source resolver needs an update only when files are added or deleted

Comments  in #1624 
> I'll need to look at this one in some detail because my description of it in the ticket might not have been top quality. We still want to watch sources for when they are created or deleted, to update the source resolver. But we no longer want to watch sources to cause re-typechecking or updates to the source resolver unless the existence of the files has changed.

yyadavalli:
This leaves the `FileChangeListener` unchanged but does not type check on file changes. 

`FileChangeListener` is also being used in `SearchService` to trigger indexing on file changes. If that is also unnecessary then that code can be removed.

Close #1625

### DIFF
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -47,7 +47,7 @@ class Project(
   private val reTypecheck = new FileChangeListener {
     def reTypeCheck(): Unit = self ! AskReTypecheck
     def fileAdded(f: FileObject): Unit = reTypeCheck()
-    def fileChanged(f: FileObject): Unit = reTypeCheck()
+    def fileChanged(f: FileObject): Unit = ()
     def fileRemoved(f: FileObject): Unit = reTypeCheck()
     override def baseReCreated(f: FileObject): Unit = reTypeCheck()
   }


yyadavalli:
We should be fine with removing `fileModified()` from `WatchListener`

https://github.com/ensime/ensime-server/blob/2.0/core/src/main/scala/org/ensime/filewatcher/FileWatchService.scala#L38

That should remove the ability of watchers to watch for file changes

fommil:
We still need to watch binary files for changes, but we don't care about changes to sources: only creation and deletion.

yyadavalli:
Hmm...looks like I mucked it up. Let me rethink and clean it.

yyadavalli:
May be I'm missing something,

The `SourceWatcher` initialized in `Project` gets `SourceResolver` as the only `FileChangeListener`. `SourceResolver` has an empty implementation for `fileChanged`, so `SourceWatcher` is not looking for file changes. I think that is what this issue suggests we implement. If I am right, there are no changes required.

Please let me know if I am misunderstanding any of this.

fommil:
I'll reread the surrounding code but I thought this change would impact binary watching too, not just source watching. We use that file watcher in two places.

yyadavalli:
Just to be clear, I think the changes in the PR are not needed and the current code is as intended.

fommil:
cool, my bad... I forgot how the code worked when I wrote the ticket. I think I'm mixing it up with another bug that was fixed, which is that if there were classes in the source directories, it was responding to those. e.g. the sbt `project` directory contains `target` as a subdirectory and that was causing problems.

This part of the code could certainly be refactored to use a consistent actor model, but such a refactor would be futile because we want to get rid of akka.

Sorry for the confusion!

yyadavalli:
No problem, let's close #1627  if no work is needed as of now. We can add a complete refactor when we move away from Akka.

Issue #1492, opened 2016-06-02
incrementally update the sources map in SourceResolver


fommil:
the comment https://github.com/ensime/ensime-server/blob/2.0/core/src/main/scala/org/ensime/indexer/SourceResolver.scala#L22

```
  // it's not worth doing incremental updates - this is cheap
  // (but it would be nice to have a "debounce" throttler)
```

is hilarious if you have lots of sources, especially if they are huge.

We should we doing this incrementally and with a debounce to avoid duplicating work.

Workaround if this affects you is to disable source watching, which is advised for large projects anyway, which will reduce the number of times this is called.

chiangqiqi:
trying to work on this

jozi-k:
Hi, I am looking for some low-hanging-fruit issue and got here, but it has been probably already solved.

fommil:
Nope, this is still open! Do you need any help getting started? The contributing page is the best plate to start and we're happy to answer questions on gitter or here

jozi-k:
I am greenhorn here, but let me explain why I assumed it had been resolved:
1) Comment from issue description does not exist anymore.
2) PR #1633 mentions this issue as resolved. 
If that is not correct, can you explain in more detail what is this issue about?

fommil:
Ah right, I'm thinking of a different sourcemap ticket... there is a ticket to reuse the source resolver in the debug actor.

PR #1629, opened 2016-10-28
add debounce to SourceResolver

chiangqiqi:
This PR refer to incrementally update the sources map in SourceResolver #1492.

use a simple actor model like Project.scala to debounce SourceResolver


fommil:
why did you close the PR? If you plan on making changes, you can just push to the branch

chiangqiqi:
I just get it wrong in the first place, Should I revert on the same branch or create another PR?

PR #1631, opened 2016-10-28
functionally correct monkeys

### LINKED ISSUE #1628: scalac archive loader should not cache zipentry locations
#1626 introduces another bug, which is that the ziparchive entries are referencing a potentially out of date jar, so the ZipEntry binary locations (and in fact the list of entries), are invalid.

I think we're working around bugs, not just performance problems, that are present in scalac. Right now in scalac it seems to take a snapshot of the file on creation and subsequent updates are ignored. Consider what happens if the jar is a dependency in your project.

Anybody using `exportJars := true` in sbt will experience this, but this is not the default so most users will not notice.

@retronym the saga of https://issues.scala-lang.org/browse/SI-9682 continues...

fommil:
there may be a perf problem here, needs further investigation but should close #1628 for now

### DIFF
--- a/monkeys/src/main/scala-2.10/scala/reflect/io/ZipArchive.scala
+++ b/monkeys/src/main/scala-2.10/scala/reflect/io/ZipArchive.scala
@@ -8,17 +8,12 @@
 package scala.reflect
 package io
 
-import java.lang.reflect.Constructor
 import java.net.URL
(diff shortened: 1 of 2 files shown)


fommil:
interesting, something is still holding onto a reference in windows. The monkey may need a rethink.

fommil:
of course windows fails, we don't have class-monkey. Ok, best thing is to just revert to an earlier version where we don't try to use class-monkey at all.

PR #1633, opened 2016-10-30
Add debounce to SourceResolver

chiangqiqi:
This PR is to solve incrementally update the sources map in SourceResolver #1492.


fommil:
btw if you write close #1492 then it gets closed automatically.

fommil:
oh, and I think this means we should be able to remove source and binary watching hacks.

fommil:
I'm not sure where you're based but if you're anywhere near EU, AU or USA, we can send you some ensime stickers to say thank you!

chiangqiqi:
close #1492

PR #1637, opened 2016-11-04
Jar urls

hardmettle:
@fommil Finally did the right way . Please take a look .

### DIFF
--- a/core/src/it/scala/org/ensime/intg/DebugTest.scala
+++ b/core/src/it/scala/org/ensime/intg/DebugTest.scala
@@ -69,20 +69,20 @@ class DebugTest extends EnsimeSpec
           32
         ) { (threadId, breakpointsFile) =>
             import testkit._
-
+            val ensimeBreakPointsFile = RawFile(breakpointsFile.toPath)
             // NOTE: Can encounter scala/Predef.scala if picking stack trace
             //       at arbitrary point
             project ! DebugBacktraceReq(threadId, 0, 3)
             expectMsgType[DebugBacktrace] should matchPattern {
               case DebugBacktrace(List(
                 DebugStackFrame(0, Nil, 0, "breakpoints.Breakpoints", "mainTest",
-                  LineSourcePosition(`breakpointsFile`, 32), _),
+                  LineSourcePosition(RawFile(rbf1), 32), _),
                 DebugStackFrame(1, List(
                   DebugStackLocal(0, "args", "Array(length = 0)[<EMPTY>]", "java.lang.String[]")
                   ), 1, "breakpoints.Breakpoints$", "main",
-                  LineSourcePosition(`breakpointsFile`, 42), _),
+                  LineSourcePosition(RawFile(rbf2), 42), _),
                 DebugStackFrame(2, Nil, 1, "breakpoints.Breakpoints", "main",
-                  LineSourcePosition(`breakpointsFile`, _), _)
+                  LineSourcePosition(RawFile(rbf3), _), _)
                 ), `threadId`, "main") =>
             }
 
@@ -95,12 +95,12 @@ class DebugTest extends EnsimeSpec
             project ! DebugContinueReq(threadId)
             expectMsg(TrueResponse)
 
-            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", breakpointsFile, 11))
+            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", ensimeBreakPointsFile, 11))
 
             project ! DebugContinueReq(threadId)
             expectMsg(TrueResponse)
 
-            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", breakpointsFile, 13))
+            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", RawFile(breakpointsFile.toPath), 13))
 
             project ! DebugClearBreakReq(breakpointsFile, 11)
             expectMsg(TrueResponse)
@@ -108,7 +108,7 @@ class DebugTest extends EnsimeSpec
             project ! DebugContinueReq(threadId)
             expectMsg(TrueResponse)
 
-            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", breakpointsFile, 13))
+            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", RawFile(breakpointsFile.toPath), 13))
 
             project ! DebugSetBreakReq(breakpointsFile, 11)
             expectMsg(TrueResponse)
@@ -119,12 +119,12 @@ class DebugTest extends EnsimeSpec
             project ! DebugContinueReq(threadId)
             expectMsg(TrueResponse)
 
-            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", breakpointsFile, 11))
+            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", RawFile(breakpointsFile.toPath), 11))
 
             project ! DebugContinueReq(threadId)
             expectMsg(TrueResponse)
 
-            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", breakpointsFile, 11))
+            asyncHelper.expectMsg(DebugBreakEvent(threadId, "main", RawFile(breakpointsFile.toPath), 11))
 
             project ! DebugContinueReq(threadId)
             expectMsg(TrueResponse)
@@ -475,7 +475,7 @@ class DebugTest extends EnsimeSpec
             val backTrace = expectMsgType[DebugBacktrace]
             // just some sanity assertions
             assert(backTrace.frames.forall(_.className.startsWith("debug.Backtrace")))
-            assert(backTrace.frames.forall(_.pcLocation.file.toString.endsWith("Backtrace.scala")))
+            assert(backTrace.frames.forall(_.pcLocation.file.toString.contains("Backtrace.scala")))
           }
       }
     }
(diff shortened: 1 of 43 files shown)


fommil:
Lots of red lights new, any ideas?

hardmettle:
@fommil I only implemented a different version of NotSoSimpleJavaFileObject. It compiled fine for me and I saw two or three tests failing. I thought of getting a review for the changes I did.That's the only cause of test failure. We can always revert back to old state, but just wanted to confirm whether I approached it right. I would try to fix those errors today in the test.

fommil:
I think you did a bad merge or rebase because if I just cherry pick your two commits since taking my branch, then everything passes except the JavaCompilerSpec. I wonder if I can force push to your branch...

fommil:
it is the same error as before, btw

```
[info]   java.lang.IllegalArgumentException: Not supported: WrappedJavaFileObject[NotSoSimpleJavaFileObject[file:///tmp/1478277747110-0/testing/simple/src/main/java/org/example/Test1.java]]
[info]   at com.sun.tools.javac.file.JavacFileManager.isSameFile(JavacFileManager.java:647)
[info]   at com.sun.tools.javac.comp.Check.checkForBadAuxiliaryClassAccess(Check.java:3471)
[info]   at com.sun.tools.javac.comp.Attr.checkIdInternal(Attr.java:3547)
[info]   at com.sun.tools.javac.comp.Attr.checkId(Attr.java:3490)
```

hardmettle:
@fommil This is strange. I will recheck my merge again . Although my changes should fix the above error .

fommil:
@hardmettle there should be an option somewhere on this PR about allowing the maintainers to collaborate on the PR. Not sure if it gives force push powers. In any case, I've force pushed to my branch and will close my PR.

fommil:
I basically have no idea what the JDK8 API is expecting, do you have any insights about what is wrong? Assuming your fix works, we should probably tidy up all the `.toString` calls and test them independently, it's not clear to me why you're manipulating/cleaning them this way.

hardmettle:
@fommil I am a bit confused . Only error I see in the build is :

```
[error] /Users/travis/build/ensime/ensime-server/core/src/test/scala/org/ensime/core/CanonSpec.scala:57: too many arguments for method apply: (rootDir: java.io.File, cacheDir: java.io.File, javaHome: java.io.File, name: String, scalaVersion: String, compilerArgs: List[String], referenceSourceRoots: List[java.io.File], subprojects: List[org.ensime.api.EnsimeModule], javaLibs: List[java.io.File])org.ensime.api.EnsimeConfig in object EnsimeConfig
[error]     Canon.config = EnsimeConfig(
```

Could you help me locating the place where you see the same test failure ?

fommil:
I pushed a clean rebase to my branch. Please reset to there and make sure your potches have been applied fully.

hardmettle:
I rebased to your branch , started getting ```openjdk-langtools/openjdk8-langtools-src.zip is required but does not exist``` error.  It has got something to do with the open jdk? What am I missing here?

fommil:
I am starting to think that your `git` is just broken. Can you maybe reinstall it?

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -211,5 +218,23 @@ private class NotSoSimpleJavaFileObject(
...
+      case p: Path => rf.toString == p.toString
+      case r: RawFile => rf.toString == r.file.toString
+      case _ => false
+    }
+    case ArchiveFile(p, e) => other match {
+      case af: ArchiveFile => af.entry.toString == e && af.jar.toFile.toString == p.toFile.toString
+      case _ => false
+    }
+  }
+  override def hashCode(): Int = f match {
```
fommil: no need to pattern match

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -211,5 +218,23 @@ private class NotSoSimpleJavaFileObject(
...
+    case rf @ RawFile(_) => rf.hashCode()
+    case af @ ArchiveFile(_, _) => af.hashCode()
+  }
+  override def getShortName: String = f match {
+    case RawFile(rf) => rf.getFileName.toString
+    case ArchiveFile(p, e) => s"""${p.getFileName.toString.replace(".jar", "")}(${e.substring(e.lastIndexOf("/"))})"""
```
fommil: I'm certain there is an easier way to do this

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -211,5 +218,23 @@ private class NotSoSimpleJavaFileObject(
...
   ): Boolean = false // this can probably be optimised, but the intended javac use is not clear
-
+  override def equals(other: Any): Boolean = this.f match {
+    case RawFile(rf) => other match {
+      case p: Path => rf.toString == p.toString
+      case r: RawFile => rf.toString == r.file.toString
+      case _ => false
+    }
+    case ArchiveFile(p, e) => other match {
+      case af: ArchiveFile => af.entry.toString == e && af.jar.toFile.toString == p.toFile.toString
```
fommil: all this `.toString` magic is cryptic to me, can we please break this into functions with unit tests that assert the behaviour?

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -187,13 +190,17 @@ class JavaCompiler(
...
 )(
     implicit
     charset: Charset
-) extends JavaFileObject {
+) extends BaseFileObject(javacFileManager) {
   @inline private def unsupported = throw new UnsupportedOperationException
-
+  override def inferBinaryName(path: java.lang.Iterable[_ <: java.io.File]): String = f match {
+    case RawFile(rf) => rf.toFile.toString
+    case ArchiveFile(p, e) => p.toFile.toString.replace(".jar", "") + s"""${if (e.startsWith("/")) "" else "/"}""" + e
```
fommil: this looks copy/paste with the equals method

core/src/main/scala/org/ensime/core/javac/JavaCompiler.scala
```diff
@@ -46,7 +48,8 @@ class JavaCompiler(
   private val workingSet = new ConcurrentHashMap[String, JavaFileObject]()
 
   private implicit def charset: Charset = Charset.defaultCharset() // how can we infer this?
-
+  val context = new com.sun.tools.javac.util.Context() //Need to have a single context and not one for each source??
```
fommil: this looks like an explanation for the line, rather than a suggestion to improve it (which is confusing). Best create a ticket and reference it.

hardmettle:
@fommil there's certainly something I am not doing right with the git thing or may be my git is broken :'( . When did a rebase to origin/2.0 I could see my git log messed up . Now when I did `git reset --hard fommil/jar-urls` I have started to get the above error.

hardmettle:
By the way does the checks seems to go green this time?

fommil:
after donig that also do `git clean -fxd`

fommil:
@hardmettle osx is still failing with 

```
[info] - should find symbol at point *** FAILED ***
[info]   java.lang.IllegalArgumentException: Not supported: WrappedJavaFileObject[NotSoSimpleJavaFileObject[file:///private/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/1478391304122-0/testing/simple/src/main/java/org/example/Test1.java]]
[info]   at com.sun.tools.javac.file.JavacFileManager.isSameFile(JavacFileManager.java:647)
```

so I don't think any of the recent changes made a difference

hardmettle:
We're at the same issue where we started . This was a futile exercise to subclass `NotSoSimpleJavaFileObject` to `BaseFileObject` (which we initially thought would solve the problem). The issue we are facing it seems internally the method call `createJavaFileObject` gets wrapped up in `WrappedJavaFileObject` somehow. We have to find a way so that the fileobject remains or gets identified as a child of  `BaseFileObject`

fommil:
yup, back to where we were. I honestly have no idea what is going wrong. Do you think you could investigate this at the hack day? It will probably involve studying the java 8 source code.

fommil:
Maybe we could back out the notsosimple jfo and only support RawFile, then merge this, and try to solve this Java 8 problem separately.

hardmettle:
@fommil I have been looking into the stack trace , but its a huge chain of call . Also  I did try to match SimpleFileObject and our implementation NotSoSimpleFileObject . I think the difference between them will lead us to the clues on what we need to include so that it doesnt get wrapped around . Scariest way around is to override all internal library calls. Anyways will continue to look into it.

fommil:
let's get a PR together early though, and fix this later. I don't want the rest of this PR to rot

fommil:
hmm, I think I messed up a commit at some point because I'm sure I went and found the java8 langtools source at one point.

core/src/it/scala/org/ensime/intg/DebugTest.scala
```diff
@@ -69,20 +69,20 @@ class DebugTest extends EnsimeSpec
...
-
+            val ensimeBreakPointsFile = RawFile(breakpointsFile.toPath)
             // NOTE: Can encounter scala/Predef.scala if picking stack trace
             //       at arbitrary point
             project ! DebugBacktraceReq(threadId, 0, 3)
             expectMsgType[DebugBacktrace] should matchPattern {
               case DebugBacktrace(List(
                 DebugStackFrame(0, Nil, 0, "breakpoints.Breakpoints", "mainTest",
-                  LineSourcePosition(`breakpointsFile`, 32), _),
+                  LineSourcePosition(RawFile(rbf1), 32), _),
```
fommil: needs to assert the correct file

core/src/it/scala/org/ensime/intg/DebugTest.scala
```diff
@@ -532,8 +532,9 @@ trait DebugTestUtils {
       // but it doesn't always come through
 
       val allEvents = gotOnStartup +: additionalOnStartup
+      //val ensimeResolvedFile = resolvedFile.toPath
       val threadId = allEvents.flatMap {
-        case DebugBreakEvent(foundThreadId, "main", `resolvedFile`, `breakLine`) =>
+        case DebugBreakEvent(foundThreadId, "main", RawFile(ensimeResolvedFile), `breakLine`) =>
```
fommil: same

PR #1638, opened 2016-11-04
API to remove single file from PC

### LINKED ISSUE #1603: add API for "remove file from presentation compiler"
The Scala presentation compiler works by type-checking a set of files in the background. Files are added to the set of "managed" files by `askReload`.  During type-checking errors are reported _only_ for these files.

Usually, files are added to the "managed" set when they are open in a tab and removed when the tab is closed. The smaller the set of files that are managed, the faster the round-trip and the more responsive the PC is. Other than that, there is no harm in keeping files around.

The corresponding method for removing a file is [removeUnitOf](https://github.com/scala/scala/blob/2.12.x/src/interactive/scala/tools/nsc/interactive/CompilerControl.scala#L21). Currently Ensime does not offer an API for this call, only "removeAll". In the Eclipse implementation, this is surrounded by `ask`, so that's probably a good idea.

sugakandrey:
close #1603

### DIFF
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -56,6 +56,11 @@ final case class TypecheckFileReq(fileInfo: SourceFileInfo) extends RpcAnalyserR
 final case class UnloadModuleReq(module: String) extends RpcAnalyserRequest
 
 /**
+ * Responds with a `VoidResponse`
+ */
+final case class UnloadFileReq(file: File) extends RpcAnalyserRequest
+
+/**
  * Response with a `VoidResponse`.
  */
 final case class TypecheckModule(module: String) extends RpcAnalyserRequest
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -58,7 +58,7 @@ final case class UnloadModuleReq(module: String) extends RpcAnalyserRequest
 /**
  * Responds with a `VoidResponse`
  */
-final case class UnloadFileReq(file: File) extends RpcAnalyserRequest
+final case class UnloadFileReq(fileInfo: SourceFileInfo) extends RpcAnalyserRequest
 
 /**
  * Response with a `VoidResponse`.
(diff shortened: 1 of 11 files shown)


api/src/main/scala/org/ensime/api/incoming.scala
```diff
@@ -56,6 +56,11 @@ final case class TypecheckFileReq(fileInfo: SourceFileInfo) extends RpcAnalyserR
 final case class UnloadModuleReq(module: String) extends RpcAnalyserRequest
 
 /**
+ * Responds with a `VoidResponse`
+ */
+final case class UnloadFileReq(file: File) extends RpcAnalyserRequest
```
fommil: can we make this a `SourceFileInfo`? We're trying to deprecate uses of `File` in the API

protocol-swanky/src/main/scala/org/ensime/server/protocol/swank/SwankFormats.scala
```diff
@@ -600,6 +600,7 @@ object SwankProtocolRequest {
   implicit val ConnectionInfoReqHint: TypeHint[ConnectionInfoReq.type] = TypeHint[ConnectionInfoReq.type](SexpSymbol("swank:connection-info"))
 
   implicit val RemoveFileReqHint: TypeHint[RemoveFileReq] = TypeHint[RemoveFileReq](SexpSymbol("swank:remove-file"))
+  implicit val UnloadFileReqHint: TypeHint[UnloadFileReq] = TypeHint[UnloadFileReq](SexpSymbol("swank:unload-file"))
```
fommil: I pity anyone who has to touch this file...

fommil:
@dragos ^^^

api/src/main/scala/org/ensime/api/incoming.scala
dragos: Would it make sense to use URIs to talk about file identity when the contents don't matter?

dragos:
In what regards PC interaction, this looks good to me. Thank you for picking up this ticket, it's may look small but I think it will improve the usability quite a lot in the long term. At least, I plan to use it in VS Code :)

fommil:
yeah, if I have the time adding it to emacs would be great too, e.g. on file close. Maybe save me reloading open files when things get weird.

sugakandrey:
@fommil can't merge this until review is approved.

fommil:
@dragos by the wonders of CI, you should get this if you restart your server in 30 mins (unless you're caching it)

PR #1641, opened 2016-11-05
Use PC instead of scalariform for `ExpandSelectionReq`

### LINKED ISSUE #1622: Rewrite expand region without scalariform
With #1619  done, the only functionality using scalariform is expand region. Rewriting that would enable us to completely remove scalariform from the server and place it in the build tool plugins.

sugakandrey:
close #1622 
The most straightforward implementation of `ExpandSelection` feature using pres. compiler `Locator` modified to return smallest tree with a position **strictly** bigger than given. The downside is that we traverse the all the trees every time client asks to expand/contract. Perhaps we can just collect all the trees, which include the initial position, and make client use them for repeating expand/contract commands. 

Also, this will require some `ensime-emacs` adjustments, current implementation copies file contents into a temporary file, which is then loaded into pres. compiler along with the real one, leading to `duplicate definition`s all over the code.

### DIFF
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -245,13 +245,13 @@ class BasicWorkflow extends EnsimeSpec
           )
 
           // expand selection around 'val foo'
-          project ! ExpandSelectionReq(fooFile, 215, 215)
+          project ! ExpandSelectionReq(fooFile, 290, 290)
           val expandRange1 = expectMsgType[FileRange]
-          expandRange1 shouldBe FileRange(fooFilePath, 214, 217)
+          expandRange1 shouldBe FileRange(fooFilePath, 287, 294)
 
-          project ! ExpandSelectionReq(fooFile, 214, 217)
+          project ! ExpandSelectionReq(fooFile, 287, 294)
           val expandRange2 = expectMsgType[FileRange]
-          expandRange2 shouldBe FileRange(fooFilePath, 210, 229)
+          expandRange2 shouldBe FileRange(fooFilePath, 269, 295)
 
           project ! RefactorReq(1234, RenameRefactorDesc("bar", fooFile, 215, 215), false)
           expectMsgPF() {
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -245,13 +245,13 @@ class BasicWorkflow extends EnsimeSpec
           )
 
           // expand selection around "seven" in `foo.testMethod` call
-          project ! ExpandSelectionReq(fooFile, 290, 290)
+          project ! ExpandSelectionReq(fooFile, 215, 215)
           val expandRange1 = expectMsgType[FileRange]
-          expandRange1 shouldBe FileRange(fooFilePath, 287, 294)
+          expandRange1 shouldBe FileRange(fooFilePath, 214, 217)
 
-          project ! ExpandSelectionReq(fooFile, 287, 294)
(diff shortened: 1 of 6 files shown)


fommil:
Hmm, W not too keen on an api change for this. Can we unload and reload the file or do something more like scala-ide where it just parses the code bust doesn't typecheck, like @dragos recommended.

fommil:
Hehe, nice. No api change is good though. Parsing the fire is basically free.

sugakandrey:
alright, should be better now :)

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -245,13 +245,13 @@ class BasicWorkflow extends EnsimeSpec
           )
 
           // expand selection around 'val foo'
-          project ! ExpandSelectionReq(fooFile, 215, 215)
+          project ! ExpandSelectionReq(fooFile, 290, 290)
```
fommil: why did you need to change the expectations?

project/EnsimeBuild.scala
```diff
@@ -71,9 +71,6 @@ object EnsimeBuild {
   )
 
   lazy val api = Project("api", file("api")) settings (commonSettings) settings (
-    libraryDependencies ++= Seq(
-      "org.scalariform" %% "scalariform" % "0.1.8"
```
fommil: :tada:

core/src/main/scala/org/ensime/core/PositionLocator.scala
```diff
@@ -0,0 +1,33 @@
+// Copyright: 2010 - 2016 https://github.com/ensime/ensime-server/graphs
+// License: http://www.gnu.org/licenses/gpl-3.0.en.html
+package org.ensime.core
+
+import scala.tools.nsc.interactive.Global
+
+trait PositionLocator {
```
fommil: if this file has been derived from scala-ide it needs to be properly attributed to them

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
sugakandrey: PC implementation of `expand-selection` is not 100% equivalent to scalariform.
For example `val f\*cursor here*\oo = new Foo()` scalariform would expand to just the name `foo`, while the smallest tree is the whole `ValDef`.

dragos: Yeah, that's a bummer. I think scala-refactoring has some code to reliably find positions of names (for instance, for Rename). Otherwise could also handle this "on site" (same goes for all definitions, I imagine).

fommil:
still flagged as WIP, let me know when you're happy with it... I'm not too keen on changing the test expectation. Adding to it is better.

sugakandrey:
I'm looking into scala-refactoring as @dragos suggested, pretty sure they have what we need.

sugakandrey:
@fommil Looks fine to me now.

fommil:
Awesome!

fommil:
this feels snappier, good stuff!

fommil:
I'm tempted to increase the API here so that editors can do more, e.g. a way to ask for all blocks in a file... would allow structural editing. Like in Emacs we could have something like smart-parens but specifically for implied scala brackets.

fommil:
protip, combine with

```elisp
(defun scala-sp-blockify-region ()
  (interactive)
  (when (region-active-p)
    (kill-region (region-beginning) (region-end))
    (insert ?{) (insert ?}) (left-char)
    (comment-indent-new-line)
    (comment-indent-new-line)
    (forward-line -1)
    (yank)
    (call-interactively 'indent-region)))
(bind-key "C-+" 'scala-sp-blockify-region)
```

for awesome `+++ C-+` workflow to blockify / indent regions.

PR #1644, opened 2016-11-11
bump akka version for 2.11+, introduce AkkaBackCompat

sugakandrey:
Akka keeps writing to `STDOUT` on shutdown in 2.4.12. We may want to wait until we find a way to fix this before merging.

### DIFF
--- a/core/src/it/scala/org/ensime/fixture/SearchServiceFixture.scala
+++ b/core/src/it/scala/org/ensime/fixture/SearchServiceFixture.scala
@@ -2,15 +2,17 @@
 // License: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.fixture
 
-import akka.actor.ActorSystem
-import org.ensime.api._
-import org.ensime.AkkaBackCompat._
-import org.ensime.vfs._
-import org.ensime.indexer.SearchService
 import scala.concurrent._
 import scala.concurrent.duration._
+import scala.util.Try
 
-trait IsolatedSearchServiceFixture extends IsolatedSourceResolverFixture {
+import akka.actor.ActorSystem
+import org.ensime.api._
+import org.ensime.AkkaBackCompat
+import org.ensime.indexer.SearchService
+import org.ensime.vfs._
+
+trait IsolatedSearchServiceFixture extends IsolatedSourceResolverFixture with AkkaBackCompat {
 
   def withSearchService(testCode: (EnsimeConfig, SearchService) => Any)(implicit actorSystem: ActorSystem, vfs: EnsimeVFS): Any = withSourceResolver { (config, resolver) =>
     val searchService = new SearchService(config, resolver)
@@ -18,8 +20,8 @@ trait IsolatedSearchServiceFixture extends IsolatedSourceResolverFixture {
       testCode(config, searchService)
     } finally {
       Await.ready(searchService.shutdown(), Duration.Inf)
-      actorSystem.close()
-      actorSystem.awaitClosure(10.seconds)
+      Try(actorSystem.terminate())
+      Try(Await.result(actorSystem.whenTerminated, 10.seconds))
     }
   }
 }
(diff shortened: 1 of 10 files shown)


fommil:
@bwmcadams said he may look into this

core/src/it/scala/org/ensime/fixture/SearchServiceFixture.scala
```diff
@@ -17,8 +18,8 @@ trait IsolatedSearchServiceFixture extends IsolatedSourceResolverFixture {
       testCode(config, searchService)
     } finally {
       Await.ready(searchService.shutdown(), Duration.Inf)
-      actorSystem.shutdown()
-      actorSystem.awaitTermination(10.seconds)
+      actorSystem.close()
```
fommil: I was thinking you could make the BackCompat only for 2.10 and use the akka 2.4 names... that way you don't need to write an impl for 2.11+

project/SensibleSettings.scala
```diff
@@ -143,15 +151,15 @@ object Sensible {
     "com.google.code.findbugs" % "jsr305" % "3.0.1" % "provided"
   )
 
-  def testLibs(config: String = "test") = Seq(
+  def testLibs(config: String = "test") = Def.setting(Seq(
```
fommil: I will probably remove a bunch of stuff from this `testLibs` when I publish it as a separate plugin

core/src/it/scala/org/ensime/fixture/SearchServiceFixture.scala
sugakandrey: Makes sense. Will do.

sugakandrey:
damn, forgot we have `-Ywarn-unused-import` & `-Xfatal-warnings`

sugakandrey:
@fommil what should I do about this? Having `BackCompat` only for 2.10 forces me to have an unused import in 2.11 code.

fommil:
you could make it a trait instead of an object, and mix it in. Sorry while I go and barf somewhere...

fommil:
We have defined timeouts in EnsimeSpec for akka that get scaled by envvars, might be best to use those.

fommil:
Do we really care about waiting for shutdown in tests?

fommil:
I guess we do, otherwise it might never work.

fommil:
On those stdouts, I'm tempted to just monkey patch whatever part of akka is doing it.

sugakandrey:
@fommil funny thing, our shutdown tests seem to have been broken all along
```
eventually(timeout(30 seconds), interval(1 second)) {
          system.isTerminated
}
```
This block never fails, since `eventually` succeeds whenever block passed to it returns a value, doesn't matter if it is `true` or `false` or whatever else.

fommil:
very interesting, this might be what @dragos was seeing

sugakandrey:
Currently our `WebServer` binds to `/0.0.0.0:port` and not `/127.0.0.1:port`. See https://github.com/ensime/ensime-server/blob/2.0/server/src/main/scala/org/ensime/server/WebServer.scala#L55. Is this intentional?

fommil:
nope, should definitely be 127.0.0.1, well spotted!

sugakandrey:
Alright, so this should fix `WebServer` startup address. Also we failed to report bind failures in [WebServer](https://github.com/ensime/ensime-server/blob/2.0/server/src/main/scala/org/ensime/server/WebServer.scala#L56-L64), which resulted in [this code](https://github.com/ensime/ensime-server/blob/2.0/server/src/main/scala/org/ensime/server/Server.scala#L75-L90) going into `Success` branch even if `Server` failed to bind and then dying with `NullPointerException` since `ch.localAddress()` was `null`. Fixed that too.

fommil:
cool! I just merged a huge PR today, can you please rebase?

dragos:
All right, glad to see I'm not the only one seeing zombies.

sugakandrey:
Sure, I'll rebase, anyways we need a way to deal with akka logging.

PR #1651, opened 2016-11-16
Fix CanonSpec and EnsimeFileSpec for cases where JAVA_HOME is not set.

### LINKED ISSUE #1650: CanonSpec and EnsimeFileSpec fail if JAVA_HOME does not exist
### Steps to reproduce

1. Remove the JAVA_HOME environment variable:
```shell
unset JAVA_HOME
```
2. Run all tests, or just CanonSpec and EnsimeFileSpec:
```shell
cd ensime-server
sbt test
``` 

### Expected behaviour
CanonSpec and EnsimeFileSpec pass.

### Actual behaviour
CanonSpec and EnsimeFileSpec use the path to the JDK source archive, "src.zip", which is located within the java home directory.
The Path is built from the string returned from ```scala 
scala.util.Properties.jdkHome```.
If JAVA_HOME is set, then jdkHome returns the java home directory. If JAVA_HOME is not set (such as the default on some systems like macOS), jdkHome returns '.<java home directoyr>/jre' which is a sibling to 'src.zip'

tonylotts:
Fixes #1650 

- Remove (if exists) suffix "/jre" from result of 'Properties.jdkHome'.

### DIFF
--- a/core/src/test/scala/org/ensime/core/CanonSpec.scala
+++ b/core/src/test/scala/org/ensime/core/CanonSpec.scala
@@ -3,12 +3,10 @@
 package org.ensime.core
 
 import java.io.File
-import java.nio.file._
-
-import scala.util.Properties.jdkHome
 
 import org.ensime.util.file._
 import org.ensime.util.path._
+import org.ensime.util.TestPaths.src
 import org.ensime.util.ensimefile._
 import org.ensime.util.EnsimeSpec
 import org.ensime.api._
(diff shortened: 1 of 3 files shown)


fommil:
I'm not keen on this change... you need to set up your dev environment according to http://ensime.github.io/contributing/ to be able to hack on ensime as this sort of thing is very subtle and fixing it for you will break it for the many other environments.

Thanks for thinking about it though, happy to help get you into getting started, just ask around on the gitter channel.

PR #1654, opened 2016-11-16
Add EnsimeProject to EnsimeConfig

yyadavalli:
Simple addition of EnsimeProject

### DIFF
--- a/api/src/main/scala/org/ensime/api/config.scala
+++ b/api/src/main/scala/org/ensime/api/config.scala
@@ -19,6 +19,7 @@ final case class EnsimeConfig(
     compilerArgs: List[String],
     referenceSourceRoots: List[File],
     subprojects: List[EnsimeModule],
+    projects: List[EnsimeProject],
     javaLibs: List[File]
 ) {
   (rootDir :: cacheDir :: javaHome :: referenceSourceRoots ::: javaLibs).foreach { f =>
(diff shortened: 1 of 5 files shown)


fommil:
cool! But we don't need to support the old format at all, so you can remove stuff :grinning:

yyadavalli:
From sbt-ensime too ??  I forgot if the clients used the .ensime for anything.

fommil:
no, `sbt-ensime` needs to be backwards compatible. But the server only needs to read the new bits.

yyadavalli:
I keep forgetting we use the same plugin for 1.0, also i'd prefer to merge it this way. We can remove once the changes are integrated downstream in the analyzer and file watchers, or else we have to reconstruct `compileClasspath` and others in `EnsimeConfig`.

fommil:
ok, fair enough. I want to do some changes on the sbt plugin first, some naming etc. Sorry I've been slow on it, but I'll definitely get to it by the next hack day (5th december). I'm currently moving the plugin over to use coursier.

yyadavalli:
That's fine, even I didn't get a chance to finish the deduping in the plugin. Just thought of adding this in as I already had it done locally. The coursier move should be good.

fommil:
needs an update for the new format

fommil:
closing until updated for the new api (no need to create a new PR... just push here if you get round to this and we'll reopen)

yyadavalli:
I will work on this next week after my exams

PR #1665, opened 2016-11-29
embarrassing... file watcher tests were disabled

fommil:


fommil:
oh great, the build is broken (not just this PR)

PR #1668, opened 2016-12-01
allows index and exiting

### LINKED ISSUE #1551: Project actor init should happen on creation, not on connection info request

fommil:
close #1551


fommil:
`org.ensime.server.ServerStartupSpec` is definitely flakey now

sugakandrey:
I'll have to take a look at `ServerStartupSpec` after I'm done with priority tickets, since it was probably me, who introduced this and it is not supposed to be flakey at all.

fommil:
@sugakandrey I've not seen it be flaky anywhere except in this PR, so almost certainly me

sugakandrey:
@fommil alright then, ping me if it comes up somewhere else

PR #1671, opened 2016-12-04
Rewriting parsers from parboiled2 to fastparse

vovapolu:
This PR removes parboiled2 module and rewrites parsers to fastparse architecture.


fommil:
@sugakandrey is this going to cause any trouble with graphpocalypse? Which one should get merged first?

sugakandrey:
@fommil don't see why it should cause any problems. I'm still fighting Windows build anyways.

Issue #1673, opened 2016-12-06
fix the build


fommil:
build is still broken. Looks like

- duplicate resource files when creating assembly file (might need to build this even in PRs)
- flaky ServerStartSpec

fommil:
I get the feeling that my hosts have made my VPS slower :-/ must be sharing with heavy users. Reducing parallelism might help.

fommil:
looks like upgrading drone and getting machines that run on metal (even though they are not particularly fast) has improved things.

PR #1674, opened 2016-12-06
fix build

fommil:
partially address #1673 and upgrade some middleware

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,7 +22,7 @@ build:
         git --no-pager diff ;
         exit 1 ;
       fi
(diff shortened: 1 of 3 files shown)


fommil:
Looks like I really need to fix the flaky server start test (and fix the formatting of the stack dump)

sugakandrey:
Looks like you got the exact same error I was getting on graph build. `FileWatcherSpec` doesn't like Windows much, in particular `  it should "survive deletion of the watched directory" ` seems to be hanging.

fommil:
argh

fommil:
well, since that's not a problem with this PR... I'll merge and we can deal with it later

PR #1675, opened 2016-12-07
Transits to scala-refactoring 0.11.0

wpopielarski:
Enables Organize Imports functionality to work with scala-refactoring 0.11.0 family version


fommil:
awesome! Thank you so much :grinning:

fommil:
also, now you've signed the GPL so you have to sing https://www.gnu.org/music/free-software-song.en.html

PR #1677, opened 2016-12-10
Debounced actors refactor

mwesolowski:
It's my first real attempt to write something in scala so it can be messed up. Proceed with caution.
I wanted to examine how issue 1492 was implemented and as a result this refactor has came up.

### DIFF
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -14,7 +14,7 @@ import scala.collection.immutable.ListSet
 import scala.concurrent.duration._
 import scala.util.Properties._
 import scala.util._
-import org.ensime.util.FileUtils
+import org.ensime.util.{ DebouncingActor, FileUtils }
 import org.ensime.util.ensimefile._
 
 final case class ShutdownRequest(reason: String, isError: Boolean = false)
@@ -27,7 +27,7 @@ class Project(
     broadcaster: ActorRef,
     implicit val config: EnsimeConfig
 ) extends Actor with ActorLogging with Stash {
-  import context.{ dispatcher, system }
+  import context.system
 
   import FileUtils._
 
@@ -123,16 +123,12 @@ class Project(
     Try(vfs.close())
   }
 
-  // debounces ReloadExistingFilesEvent
-  private var rechecking: Cancellable = _
+  private val debouncedTypeRechekingActor =
+    context.actorOf(Props(classOf[DebouncingActor], () => scalac ! ReloadExistingFilesEvent, 5 seconds), "debouncedTypeRechecking")
 
   def handleRequests: Receive = withLabel("handleRequests") {
     case ShutdownRequest => context.parent forward ShutdownRequest
-    case AskReTypecheck =>
-      Option(rechecking).foreach(_.cancel())
-      rechecking = system.scheduler.scheduleOnce(
-        5 seconds, scalac, ReloadExistingFilesEvent
-      )
+    case AskReTypecheck => debouncedTypeRechekingActor ! DebouncingActor.Debounce
     // HACK: to expedite initial dev, Java requests use the Scala API
     case m @ TypecheckFileReq(sfi) if sfi.file.isJava => javac forward m
     case m @ CompletionsReq(sfi, _, _, _, _) if sfi.file.isJava => javac forward m
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -7,13 +7,13 @@ import java.util.concurrent.Semaphore
 import scala.concurrent._
 import scala.concurrent.duration._
 import scala.util.{ Failure, Properties, Success }
-
 import akka.actor._
 import akka.event.slf4j.SLF4JLogging
 import org.apache.commons.vfs2._
 import org.ensime.api._
 import org.ensime.indexer.database._
 import org.ensime.indexer.database.DatabaseService._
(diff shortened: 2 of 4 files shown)


fommil:
cool! I was hoping somebody would clean this up :smile:

Instead of taking in a lambda, how about taking a target actor? (and a delay period, and maybe an option to not debounce the first action)

at you at ScalaX? Come round to the Shift room if so

mwesolowski:
Sure. Does it mean I should also turn SourceResolver into an actor?
No, I'm not at ScalaX. My adventure with Scala just begins. Maybe next year :)

fommil:
that would be a pretty good change if you moved it to be an actor... I think I wrote it before we were using akka.

fommil:
sorry, I missed that you'd updated this! Could you please rebase? I've been ill recently and it slipped my mind. Do not be afraid to ping me every week if I seem to have forgotten.

fommil:
@mwesolowski ping

mwesolowski:
I've rebased but my change is still based on lambdas. I've made some attempts to move SourceResolver into an actor but it still needs some extra workload. I will try to find some time for it next week.

fommil:
@mwesolowski ping, any more thoughts on this one?

mwesolowski:
First I have to get the tests pass locally on my machine (I use macOS Sierra 10.12.4 and sbt 0.13.13). When I run `sbt ++2.11.8 ensimeConfig test` on the clean commit 4e580733 (current HEAD of 2.0) I get the following errors in the tests related to file watchers. Have you ever come across such a case?

SourceResolver is a FileChangeListener itself so I feel quite insecure trying to move this class to be an actor without having working tests in this area. 

```
[info] - should detect added / changed files *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.ensime.indexer.Added
```
```
[info] - should detect added / removed files *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.ensime.indexer.Added
```
```
[info] - should detect removed base directory *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during fishForMessage, hint:
```
```
[info] - should detect removed parent base directory *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during fishForMessage, hint:
```
```
[info] - should survive deletion of the watched directory *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.ensime.indexer.Added
```
```
[info] - should survive removed parent base directory and recreated base *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.ensime.indexer.Added
```
```
[info] - should detect changes to a file base *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.ensime.indexer.Changed
```
```
[info] - should detect removal of a file base *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.ensime.indexer.Removed
```
```
[info] - should be able to start up from a non-existent base file *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.ensime.indexer.Added
```
```
[info] - should survive removal of a file base *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class org.ensime.indexer.Removed
```

fommil:
If you're on Mac you can ignore FileWatcherSpec. But integration tests are more important.

timcharper:
Any reason to not use akka streams? Akka streams has some nice components to use for debouncing.

iravid:
Heartily agree on akka-streams but Sam seems to be set on FS2 :-)

fommil:
Akka is too big and has been the biggest problem for us maintaining crossbuild matrices. We have many ensime users who also contribute to fs2.

timcharper:
@fommil I propose we use the following, non-akka specific implementation of Debouncer:

https://gist.github.com/timcharper/51b4a6c4a5291cdfe7970e3f855b2a7c

It has several performance advantages:

- Debounced attempts are filtered synchronously, rather than asynchronously; this means significantly less overhead for the ExecutionContext. Implementation uses AtomicBoolean to guarantee that an invocation will happen after each `call()` attempt.
- It has significantly less interactions with the scheduler. Akka's scheduler instantiates a new TaskHolder instance for each `scheduleOnce` call, and puts it in the queue. Worth noting, the `cancel()` method for the `scheduleOnce` returned cancellable only marks the task as cancelled. This means that the scheduler will still attempt to call every scheduled task.
- Supports a `maxDelay` parameter, which will ensure that the debounced function will be called in the event that events are getting called frequently.
- Less dependency on Akka. Scheduler implementation can be easily swapped out.

timcharper:
@mwesolowski I pushed up a new version of this pull request; I set you as the author, with me as the committer, so we both get credit for the change.

I wanted debouncing to happen _before_ we sent it to the actor, as otherwise it generates a lot of log messages. Also, the above ^

See #1773 

Are you okay with this? @fommil is this kosher with you?

If preferred, we can just merge your patch, @mwesolowski , and then I can rebase mine on top, too. Whatever you prefer.

timcharper:
Doh. I just saw in the history that you requested it be moved to an actor, and before it _was_ not an actor (and took a Lambda).

Perhaps we can discuss tomorrow :)

fommil:
closing in preference of #1773

PR #1681, opened 2016-12-10
Graphpocalypse

sugakandrey:
An attempt to merge `Graphpocalypse` into `2.0`.
Windows build is most probably broken in this one.


fommil:
Apparently your altor ego must sign the CLA.

fommil:
Omg, evil Andry **is** Andrey!!!!

.drone.yml
```diff
@@ -4,10 +4,11 @@ build:
   environment:
     - SONATYPE_USERNAME=$$SONATYPE_USERNAME
     - SONATYPE_PASSWORD=$$SONATYPE_PASSWORD
-    - ENSIME_SERVER_VERSION=2.0.0-SNAPSHOT
+    - ENSIME_SERVER_VERSION=2.0.0-graph-SNAPSHOT
```
fommil: typo

.drone.yml
```diff
@@ -4,10 +4,11 @@ build:
...
     - AKKA_TEST_TIMEFACTOR=10
     - SBT_TASK_LIMIT=6
     - COURSIER_MODE=update-local-changing,missing
+    - ENSIME_JVM_TEST_FLAGS=-XX:MaxDirectMemorySize=4g
```
fommil: needs a more permament fix. Put in .jvmopts

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -292,22 +289,6 @@ class Analyzer(
       sender ! VoidResponse
   }
 
-  def handleReloadFiles(files: List[SourceFileInfo]): RpcResponse = {
-    val (existing, missingFiles) = files.partition(_.exists())
```
fommil: is there a reason why we're throwing away the existence check? I recall this was causing problems

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -181,7 +182,7 @@ trait RichCompilerControl extends CompilerControl with RefactoringControl with C
         file <- config.scalaSourceFiles
         source = createSourceFile(file)
       } yield source
-    }.toSet ++ activeUnits().map(_.source)
+    } ++ activeUnits().map(_.source)
```
fommil: reason to drop the Set? might mean we do more work than needed

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -126,51 +153,19 @@ final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
   private trait ReferenceInClassHunter {
     this: ClassVisitor =>
 
-    def clazz: RawClassfile
+    var clazz: RawClassfile
```
fommil: we probably need to refactor this trait at some point, vars in traits is bad but also deprecated soon

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -126,51 +153,19 @@ final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
...
-    }
-
-    override def visitInnerClass(name: String, outerName: String, innerName: String, access: Int): Unit = {
-      addRef(ClassName.fromInternal(name))
-    }
-
-    override def visitOuterClass(owner: String, name: String, desc: String): Unit = {
-      addRef(ClassName.fromInternal(owner))
-    }
+    var internalRefs = Set.empty[FullyQualifiedName]
```
fommil: we might want to use a java HashSet, much faster

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -179,11 +174,26 @@ final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
     ) = handleAnn(desc)
   }
 
+  private trait ReferenceInFieldHunter {
+    this: FieldVisitor =>
+
+    protected var internalRefs = Set.empty[FullyQualifiedName]
```
fommil: java set might be faster

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -179,11 +174,26 @@ final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
...
+    override def visitTypeAnnotation(typeRef: Int, typePath: TypePath, desc: String, visible: Boolean): AnnotationVisitor = handleAnn(desc)
+    override def visitAnnotation(desc: String, visible: Boolean): AnnotationVisitor = handleAnn(desc)
+  }
+
   private trait ReferenceInMethodHunter {
     this: MethodVisitor =>
 
-    // NOTE: :+ and :+= are really slow (scala 2.10), prefer "enqueue"
-    protected var internalRefs = Queue.empty[FullyQualifiedName]
+    protected var internalRefs = Set.empty[FullyQualifiedName]
```
fommil: same again. We'd profiled this and Queue was pretty fast... using scala Sets is surely going to slow things down

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -34,15 +33,29 @@ class SearchService(
...
-  private[indexer] def isUserFile(file: FileName): Boolean = {
-    (config.allTargets map (vfs.vfile)) exists (file isAncestor _.getName)
-  }
+  private[indexer] val allTargets = config.allTargets.map(vfs.vfile)
+
+  private[indexer] def isUserFile(file: FileName): Boolean = allTargets.exists(file isAncestor _.getName)
 
   private val QUERY_TIMEOUT = 30 seconds
 
+  private[indexer] val ensimeVfs = vfs
```
fommil: why not just do this on the `vfs` param?

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -303,10 +459,12 @@ class IndexingQueueActor(searchService: SearchService) extends Actor with ActorL
   // De-dupes files that have been updated since we were last told to
   // index them. No need to aggregate values: the latest wins. Key is
   // the URI because FileObject doesn't implement equals
-  var todo = Map.empty[String, FileObject]
+  private val todo = new m.HashMap[FileName, m.Set[FileObject]] with m.MultiMap[FileName, FileObject]
```
fommil: how come you needed to make these changes? Seem pretty independent of the graph

core/src/main/scala/org/ensime/indexer/orientdb/SimpleOrient.scala
```diff
@@ -0,0 +1,372 @@
...
+  // NOTE: although providing isolation, this destroys stacktraces
+  def withGraphAsync[T](
+    f: OrientBaseGraph => T
+  )(
+    implicit
+    factory: OrientGraphFactory,
+    ec: ExecutionContext
+  ): Future[T] = Future { blocking { withGraph(f) } }
+
+  // the presentation complier doesn't like it if we pimp the Graph,
```
fommil: old terminology: pimp => enrich (this was probably me)

project/EnsimeBuild.scala
```diff
@@ -17,15 +17,15 @@ import org.ensime.EnsimeKeys._
...
+ 
 object ProjectPlugin extends AutoPlugin {
   override def requires = fommil.SensiblePlugin
   override def trigger = allRequirements
 
   override def buildSettings = Seq(
     scalaVersion := "2.11.8",
     organization := "org.ensime",
-    version := "2.0.0-SNAPSHOT",
+    version := "2.0.0-graph-SNAPSHOT",
```
fommil: typo

fommil (changes requested):

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
sugakandrey: This is already a `Set` this conversion is just redundant.

core/src/main/scala/org/ensime/indexer/SearchService.scala
sugakandrey: This has to do with grouping all inner classes with their top level class to get scalasig out of it.

sugakandrey:
@fommil I think I made one commit from fommil.com and that's what confusing the CLA. Not sure if I can fix that without rewriting all the history.

core/src/main/scala/org/ensime/core/Analyzer.scala
sugakandrey: Not sure what do you mean, `handleReloadFiles` was just moved to `RichPresentationCompiler.scala` but is left intact.

.drone.yml
sugakandrey: Whats the typo? Am I missing something obvious? Also, yours, not mine :)

fommil: Since you're merging into 2.0 we need to drop references to the graph branch.

fommil:
Rewriting  history (and rebasing) is fine.

fommil:
If we really need this jvm flag we should be adding it in ensime-sbt by default on java 8. It's a real shame we need it. I can't remember if the envvar is even used.

fommil:
I've been doing some more research on MaxDirectMemorySize. Are you sure this is needed in Java 8? It seems that the default is unbounded... http://docs.oracle.com/cd/E15289_01/doc.40/e15062/optionxx.htm#BABGCFFB

if that is the case, we shouldn't be demanding any changes for our users.

core/src/main/scala/org/ensime/core/Analyzer.scala
fommil: ok, I missed that

sugakandrey:
I'll check, it definitely was needed on 1.7

fommil:
@sugakandrey it might only have been needed because the max heap size was too low. Increasing the max heap size might be enough, in which case we should reduce the parallelism of our tests accordingly.

sugakandrey:
Damn, it's 2017 already, need to bump headers

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
```diff
@@ -57,7 +57,7 @@ class FileWatcherSpec extends EnsimeSpec
 
   val maxWait = 20 seconds
 
-  "FileWatcher" should "detect added files" taggedAs (Retryable) in
+  "FileWatcher" should "detect added files" taggedAs (IgnoreOnTravis, IgnoreOnAppVeyor, Retryable) in
```
fommil: We don't want to ignore these on windows or osx. The problem was rogue threads, but I fixed it.

core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
```diff
@@ -74,7 +74,7 @@ class JavaCompilerSpec extends EnsimeSpec with OptionValues
     }
   }
 
-  it should "find symbol at point" in withJavaCompiler { (_, config, cc, store, search) =>
+  it should "find symbol at point" taggedAs IgnoreOnTravis in withJavaCompiler { (_, config, cc, store, search) =>
```
fommil: there is no reason we need to ignore this on travis

core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
```diff
@@ -27,7 +27,7 @@ class RichPresentationCompilerThatNeedsJavaLibsSpec extends EnsimeSpec
 
   val original = EnsimeConfigFixture.SimpleTestProject
 
-  "RichPresentationCompiler" should "locate source position of Java classes in import statements" in {
+  "RichPresentationCompiler" should "locate source position of Java classes in import statements" taggedAs (IgnoreOnTravis) in {
```
fommil: no need to ignore on travis

core/src/it/scala/org/ensime/core/ScalapSymbolToFqnSpec.scala
```diff
@@ -3,13 +3,13 @@
...
 import org.ensime.api.{ DeclaredAs, EnsimeConfig }
-import org.ensime.fixture.{ EnsimeConfigFixture, IsolatedRichPresentationCompilerFixture }
+import org.ensime.fixture.{ EnsimeConfigFixture, SharedRichPresentationCompilerFixture }
 import org.ensime.indexer._
 import org.ensime.util.EnsimeSpec
 import org.ensime.vfs._
 
 class ScalapSymbolToFqnSpec extends EnsimeSpec
-    with IsolatedRichPresentationCompilerFixture
+    with SharedRichPresentationCompilerFixture
```
fommil: is this for performance?

core/src/it/scala/org/ensime/core/ScalapSymbolToFqnSpec.scala
```diff
@@ -31,16 +31,17 @@ class ScalapSymbolToFqnSpec extends EnsimeSpec
...
+        verify(scalaClass.javaName, scalaClass.scalaName, scalaClass.declaredAs, cc)
     }
   }
 
   it should "find and resolve field names in Predef" in withPresCompiler { (_, cc) =>
     val vfs = cc.vfs
 
     val predef = vfs.vres("scala/Predef.class")
-    val fieldNames = new ClassfileDepickler(predef).getClasses.flatMap(_.fields)
+    val fieldNames = new ClassfileDepickler(predef).getClasses.values.flatMap(_.fields.valuesIterator)
```
fommil: `.values` is almost always better than `valuesIterator` as the latter exposes (mutable) implementation details

core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -200,23 +205,155 @@ class SearchServiceSpec extends EnsimeSpec
...
-    val hits = service.searchClassesMethods("toString" :: Nil, 10).map(_.fqn)
+    val hits = service.searchClassesMethods("toString" :: Nil, 8).map(_.fqn)
     all(hits) should startWith regex ("org.example|org.boost")
   }
 
+  it should "distinguish between traits/classes/objects" in withSearchService { implicit service =>
+    val aTrait = service.findUnique("org.scalatest.FunSuiteLike")
+    val aClass = service.findUnique("org.scalatest.FunSuite")
+    val anObject = service.findUnique("org.scalatest.SuperEngine$Bundle$")
+    aTrait shouldBe defined
```
fommil: FYI http://www.scalatest.org/at_a_glance/FlatSpec#matchers_option_either

```scala
import org.scalatest.EitherValues._
val either = Right(3)
either.right.value should be < 7 
```

core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -200,23 +205,155 @@ class SearchServiceSpec extends EnsimeSpec
...
+    aTrait.get.toSearchResult should startWith("Trait")
+    aClass shouldBe defined
+    aClass.get.toSearchResult should startWith("Class")
+    anObject shouldBe defined
+    anObject.get.toSearchResult should startWith("Object")
+  }
+
+  it should "find scala names for scala symbols" in withSearchService { implicit service =>
+    val hits = service.searchClassesMethods(List("TestSuite"), 10)
+    hits.length should be > 0
```
fommil: `hits should be ('nonEmpty)`

core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -200,23 +205,155 @@ class SearchServiceSpec extends EnsimeSpec
...
+  }
+
   "lucene index" should "not contain duplicates" in withSearchService { implicit service =>
     import scala.collection.JavaConverters._
     val lucene = service.index.lucene
     val terms = List("org", "example", "bar")
     val query = new DisjunctionMaxQuery(
       terms.map(service.index.buildTermClassMethodQuery).asJavaCollection, 0f
     )
-    val fqns = lucene.search(query, 10).map(_.get("fqn"))
```
fommil: looks like we dropped a check here?

core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -225,7 +362,7 @@ object SearchServiceTestUtils {
   def refresh()(implicit service: SearchService): (Int, Int) =
     Await.result(service.refresh(), Duration.Inf)
 
-  def searchClasses(expect: String, query: String)(implicit service: SearchService) = {
+  def searchClasses(expect: String, query: String)(implicit service: SearchService, p: Position) = {
```
fommil: interesting, I thought the implicit `Position`s were in the 2.0 branch. I must have missed this test.

core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -255,19 +392,34 @@ object SearchServiceTestUtils {
...
+  def searchesEmpty(queries: String*)(implicit service: SearchService, p: Position) =
     queries.toList.foreach(searchExpectEmpty)
 
   // doesn't assert that expect finds itself because the lucene query
   // syntax conflicts with the characters in method FQNs
-  def searchesMethods(expect: String, queries: String*)(implicit service: SearchService) =
+  def searchesMethods(expect: String, queries: String*)(implicit service: SearchService, p: Position) =
     (queries.toList).foreach(searchClassesAndMethods(expect, _))
 
+  def getClassHierarchy(fqn: String, hierarchyType: Hierarchy.Direction)(implicit service: SearchService, p: Position): Hierarchy = {
```
fommil: might make sense just to return this as a `Future` (but then you probably end up blocking somewhere else...)

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -251,6 +255,7 @@ class BasicWorkflow extends EnsimeSpec
             BasicTypeInfo("Test2", DeclaredAs.Object, "org.example.Test2"),
             BasicTypeInfo("WithPolyMethod", DeclaredAs.Object, "org.example.WithPolyMethod"),
             BasicTypeInfo("WithPolyMethod", DeclaredAs.Class, "org.example.WithPolyMethod"),
+            BasicTypeInfo("package", DeclaredAs.Object, "org.example.package"),
```
fommil: is this a dupe line?

core/src/it/scala/org/ensime/intg/JarTargetTest.scala
```diff
@@ -41,11 +41,17 @@ class JarTargetTest extends EnsimeSpec
...
       withTestKit { implicit tk =>
         withProject { (project, asyncHelper) =>
+          import tk._
           mainTarget should be a 'file
 
           eventually(timeout(scaled(10 seconds)), interval(scaled(1 second))) {
             mainTarget.delete() shouldBe true
           }
+
+          eventually(interval(1 second)) {
```
fommil: we don't really need this assertion

core/src/it/scala/org/ensime/intg/JavaWorkflow.scala
```diff
@@ -16,7 +16,7 @@ class JavaWorkflow extends EnsimeSpec
 
   val original = EnsimeConfigFixture.JavaTestProject
 
-  "ensime-server" should "open the pure Java test project" in {
+  "ensime-server" should "open the pure Java test project" taggedAs (IgnoreOnTravis) in {
```
fommil: ignoring on travis again...

core/src/it/scala/org/ensime/intg/UnsavedFileTest.scala
```diff
@@ -11,9 +11,7 @@ import org.ensime.util.file._
  * Verifies common operations work correctly for unsaved files.
  */
 class UnsavedFileTest extends EnsimeSpec
-    with IsolatedEnsimeConfigFixture
-    with IsolatedTestKitFixture
-    with IsolatedProjectFixture {
+    with SharedProjectFixture {
```
fommil: perf again?

core/src/main/scala/org/ensime/core/Refactoring.scala
```diff
@@ -184,8 +187,20 @@ trait RefactoringImpl {
           reloadAndType(file)
           doInlineLocal(procId, tpe, file, start, end)
         case RenameRefactorDesc(newName, file, start, end) =>
-          reloadAndType(file)
-          doRename(procId, tpe, newName, file, start, end)
+          import scala.reflect.internal.util.{ RangePosition, OffsetPosition }
+          val sourceFile = createSourceFile(file.getPath)
+          askLoadedTyped(sourceFile)
```
fommil: we seem to be doing rather a lot of blocking in this actor... I suppose that's ok because we can't use the PC from other responses anyway. But we definitely shouldn't be blocking on the DB, since this actor manages PC state.

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -205,8 +206,84 @@ trait RichCompilerControl extends CompilerControl with RefactoringControl with C
...
+  def usesOfSym(sym: Symbol): collection.Set[RawFile] = {
+    import scala.concurrent.Await
+    import scala.concurrent.duration.Duration
+
+    val noReverseLookups = search.noReverseLookups
+    if (noReverseLookups) {
+      Set.empty
+    } else {
+      val fqn = askSymbolFqn(sym)
+      val usages = Await.result(search.findUsages(fqn.fqnString), Duration.Inf)
```
fommil: yeah, this is where the blocking happens... I'm not too keen on this. It means the PC actor is blocking on DB access. It may be a big refactor though.

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -96,27 +119,33 @@ final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
...
-          region match {
-            case "<init>" | "<clinit>" =>
-              (clazz.source.line, firstLine) match {
-                case (_, None) =>
-                case (Some(existing), Some(latest)) if existing <= latest =>
-                case _ =>
-                  clazz = clazz.copy(source = clazz.source.copy(line = firstLine))
-              }
-
-            case name =>
```
fommil: I'm not 100% sure why we no longer skip this for init sections.

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -12,11 +13,8 @@ import org.objectweb.asm._
 import org.objectweb.asm.Opcodes._
 
 final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
```
fommil: a lot of these changes could be raised against 2.0 without the graph. Might be an option if the performance issue drags on for much longer...

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -109,42 +144,57 @@ class SearchService(
...
-            case d => scan(vfs.vfile(d))
-          } ::: m.testTargets.flatMap {
-            case d if !d.exists() => Nil
-            case d if d.isJar => List(vfs.vfile(d))
-            case d => scan(vfs.vfile(d))
-          } :::
-            m.compileJars.map(vfs.vfile) ::: m.testJars.map(vfs.vfile)
-      }
-    }.toSet ++ config.javaLibs.map(vfs.vfile)
+            case d => scanGrouped(vfs.vfile(d), grouped); Nil
```
fommil: `scanGrouped` is ignored then `Nil` is returned.... a typo?

core/src/main/scala/org/ensime/indexer/stringymap/StringyMap.scala
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  implicit def CNilBigDataFormat[T]: BigDataFormat[CNil] = new BigDataFormat[CNil] {
+    override def label: String = ???
+    override def toProperties(t: CNil): StringyMap = ???
+    override def toSchema: Map[String, OrientProperty] = ???
+    override def fromProperties(m: StringyMap): BigResult[CNil] = ???
+  }
+
+  implicit def CoproductBigDataFormat[Key <: Symbol, Value, Tail <: Coproduct](
```
fommil: do we really care about Coproducts?

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
fommil: travis/appveyor reminder

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
```diff
@@ -85,7 +84,7 @@ class FileWatcherSpec extends EnsimeSpec
       }
     }
 
-  it should "detect added / changed files" taggedAs (Retryable) in
+  it should "detect added / changed files" taggedAs (IgnoreOnTravis, IgnoreOnAppVeyor, Retryable) in
```
fommil: (whole file)

project/EnsimeBuild.scala
```diff
@@ -49,7 +49,13 @@ object EnsimeBuild {
     ),
 
     // disabling shared memory gives a small performance boost to tests
-    javaOptions ++= Seq("-XX:+PerfDisableSharedMem"),
+    javaOptions ++= Seq(
+      "-XX:+PerfDisableSharedMem",
+      "-Xms1g",
+      "-Xmx1g",
+      "-XX:MaxDirectMemorySize=4g"
```
fommil: do we really need 4g here? Seems a lot, maybe 1g?

project/EnsimeBuild.scala
```diff
@@ -155,14 +161,10 @@ object EnsimeBuild {
...
-        "com.h2database" % "h2" % "1.4.193",
-        "com.typesafe.slick" %% "slick" % {
-          CrossVersion.partialVersion(scalaVersion.value) match {
-            case Some((2, 10)) => "3.1.1"
-            case _             => "3.2.0-M2"
-          }
-        },
-        "com.zaxxer" % "HikariCP" % "2.5.1",
+        "com.orientechnologies" % "orientdb-graphdb" % orientVersion
+          exclude ("commons-collections", "commons-collections")
```
fommil: anything else we can exclude? It seems a huge dependency

fommil (changes requested):

fommil:
Good news is that 2.11 and 2.12 are passing on the drone server. 2.10 could be transient and just require a longer wait.

Bad news is the drone build is over an hour per scala version now (ouch), and my host doesn't have any more powerful machines to use, and windows/osx builds are failing.

I think we should do some things to improve the performance before we can merge this. Ideas?

fommil:
btw, I think performance is probably good enough for experimental use... the user impact is not that bad. But the impact to the CI farm is really not good, we can't merge something that breaks travis and appveyor, and makes drone grind to a halt.

fommil:
maybe the hack of indexing a basic scala library once, and using that in the integration tests, is not so bad. It could be made easy if we can get the sbt index-then-exit task to work.

fommil:
Travis seems ok! But then we are ignoring a lot of tests.

core/src/main/scala/org/ensime/indexer/stringymap/StringyMap.scala
```diff
@@ -0,0 +1,283 @@
+// Copyright: 2010 - 2017 https://github.com/ensime/ensime-server/graphs
+// License: http://www.gnu.org/licenses/gpl-3.0.en.html
+/**
+ * TypeClass (api/impl/syntax) for marshalling objects into
+ * `java.util.HashMap<String,Object>` (yay, big data!).
+ */
+package org.ensime.indexer.stringymap
+
+import org.ensime.indexer.orientdb.api.OrientProperty
```
fommil: hmm, I'd not too keen on making this depend on orientdb's API... perhaps we need a separate typeclass for schemas.

fommil (changes requested):

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -0,0 +1,368 @@
...
+
+  implicit val FqnSymbolS: BigDataFormat[FqnSymbol] = cachedImplicit
+  implicit val MemberS: BigDataFormat[Member] = cachedImplicit
+  implicit val FileCheckS: BigDataFormat[FileCheck] = cachedImplicit
+  implicit val ClassDefS: BigDataFormat[ClassDef] = cachedImplicit
+  implicit val MethodS: BigDataFormat[Method] = cachedImplicit
+  implicit val FieldS: BigDataFormat[Field] = cachedImplicit
+
+  import shapeless._
+  implicit val UniqueFileCheckV = LensId("filename", lens[FileCheck] >> 'filename)
```
fommil: implicit vals should have explicit types

fommil (changes requested):

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -0,0 +1,368 @@
...
+  implicit val FieldV = LensId("fqn", lens[Field] >> 'fqn)
+  implicit val MethodV = LensId("fqn", lens[Method] >> 'fqn)
+  implicit val ClassDefV = LensId("fqn", lens[ClassDef] >> 'fqn)
+
+  private implicit val FqnSymbolLens = new Lens[FqnSymbol, String] {
```
fommil: perhaps using `Lens` here in the first place was heavy handed, since we only want to use it for reading values (not returning copies with a value changed).

It would be a lot simpler if you just defined

```scala
  private implicit val UniqueFqnSymbolV: BigDataFormatId[FqnSymbol, String] =
    new BigDataFormatId[FqnSymbol, String] {
      override def key = "fqn"
      override def value(t: FqnSymbol): String = t.fqn
    }
```

without bothering with `Lens`.

fommil (changes requested):

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
fommil: and if this is defined for `FqnSymbol`, then it probably isn't needed for all the subtypes. A lot of this code could probably be removed (including all the `LensId` stuff)

fommil:
the shapeless enum stuff, and the stringy map typeclass, need tests. You could use these as a starting point https://github.com/fommil/shapeless-for-mortals/blob/master/smbd/src/test/scala/StringyMapBigDataSpec.scala

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
```diff
@@ -44,11 +47,11 @@ class RefactoringHandlerSpec extends EnsimeSpec
     val analyzer = analyzerRef.underlyingActor
 
     val procId = 1
-    val result = analyzer.handleRefactorRequest(
+    val result = Await.result(analyzer.handleRefactorRequest(
```
fommil: I recently discovered `.futureValue` which does this :man_shrugging:

sugakandrey: cool stuff!

fommil: if you rebase, it's been added to `EnsimeSpec` already (you will probably get a few trivial conflicts, best to read that PR to see what changed)

fommil:
@sugakandrey I think we're crossing each others paths :smile: I already did the rebase with 2.0 and added a few small changes.

fommil:
reverse lookups test is broken for me

sugakandrey:
@fommil damn, only saw that, after I did the merge myself :sweat: I'll look at it now.

sugakandrey:
@fommil hmm, `ReverseLookupsSpec` runs fine for me locally, I'll investigate more tomorrow.

fommil:
I put some thoughts in gitter and the orient ticket. That's about all I can do for a few days, sorry. I think there is a memory leak in orient.

core/src/it/scala/org/ensime/core/RichPresentationCompilerSpec.scala
```diff
@@ -52,7 +52,7 @@ class RichPresentationCompilerThatNeedsJavaLibsSpec extends EnsimeSpec
 }
 
 class RichPresentationCompilerSpec extends EnsimeSpec
-    with IsolatedRichPresentationCompilerFixture
+    with SharedRichPresentationCompilerFixture
```
fommil: (I added this... but I'd rather we didn't have to do this)

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -65,29 +63,54 @@ final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
...
-        interfaces.map(ClassName.fromInternal)(breakOut),
+        Set.empty,
+        superClass,
+        interfaceNames,
         Access(access),
         (ACC_DEPRECATED & access) > 0,
-        Queue.empty, Queue.empty, RawSource(None, None)
+        Nil, Queue.empty, RawSource(None, None),
+        isScala = false,
+        internalRefs.asScala
```
fommil: it will be quicker for this to be converted into a `List` (we'd be relying on convention that it's actually distinct)

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -195,61 +198,95 @@ final case class Descriptor(params: List[DescriptorType], ret: DescriptorType) {
...
-  name: MethodName,
-  access: Access,
-  generics: Option[String],
-  line: Option[Int]
-)
+    name: MethodName,
+    access: Access,
+    generics: Option[String],
+    line: Option[Int],
+    internalRefs: Set[FullyQualifiedName]
```
fommil: this is where we'd change the collection type

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -195,61 +198,95 @@ final case class Descriptor(params: List[DescriptorType], ret: DescriptorType) {
...
+final case class RawScalapClass(
   javaName: ClassName,
   scalaName: String,
   typeSignature: String,
   access: Access,
   declaredAs: DeclaredAs,
-  fields: Seq[RawScalaField],
-  methods: Seq[RawScalaMethod]
-)
+  fields: Map[String, RawScalapField],
```
fommil: using a `Seq[(String, RawScalapField)]` may not be glamorous, but it could also speed this up. I think I would prefer the `RawScalapField` to contain its name. Is the `String` here the bytecode name or the scala name? We need some convention/documentation on this or it will drive us crazy.

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -453,15 +452,12 @@ class BasicWorkflow extends EnsimeSpec
...
           project ! UnloadFileReq(toBeUnloaded2)
           expectMsg(VoidResponse)
-          // file with warning have been unloaded
+          // file with warning has been unloaded
           // `NewScalaNotesEvent` should now not appear when typechecking `bazFile`
 
           project ! TypecheckFilesReq(List(Left(bazFile)))
           expectMsg(VoidResponse)
-          all(asyncHelper.receiveN(2)) should matchPattern {
-            case note: NewScalaNotesEvent =>
```
fommil: yeah, this was always weird

fommil: but why is it passing now?

sugakandrey: I fixed test project file to use `null` instead of `???`

sugakandrey:
Travis & Drone seem to be a lot faster with reusing cache across tests, I unignored most of the tests on Travis to see how it goes.

NOTICE
```diff
@@ -0,0 +1,14 @@
+ENSIME Server
+Copyright 2010 - 2016 Aemon Cannon
+Copyright 2014 - 2017 Sam Halliday
+Copyright 2010 - 2017 https://github.com/ensime/ensime-server/graphs
```
fommil: @sugakandrey you'll be a major committer once graph branch lands, you should add your name here :smile:

fommil:
been a while, but I'm looking into this today. Getting OOM in the SemanticHighlightingSpec.

fommil:
I've forgotten how to push to your branch.

sugakandrey:
@fommil you should be able to push with `https` I think

fommil:
@sugakandrey I added 2FA and I think that breaks it.

fommil:
@sugakandrey can you please add me as collaborator on your repo? I think it's the only way I can push to your branch now that I have 2FA.

fommil:
moved to #1747

Issue #1346, opened 2016-03-12
project/server config as HOCON or JSON


fommil:
it would be good if we had a ServerConfig as well as an EnsimeConfig, and if both were in HOCON or JSON so that clients don't have to implement an s-expression parser.

fommil:
and when we do this, perhaps we should change the `.ensime` structure so that each project configuration is explicitly listed (i.e. `compile`, `test`, `fun`, `it`)

fommil:
@yyadavalli how about this one? :smile:

fommil:
the `.ensime` format changes are already tracked in another ticket

yyadavalli:
I was thinking of doing https://github.com/ensime/ensime-sbt/issues/217 first, but either one should be fine

I think finishing this before #1608 would be better, we can avoid rewriting `SexpParser.scala`

yyadavalli:
I believe we can use https://github.com/typesafehub/config with HOCON

fommil:
yeah that would be good! for the sbt one make sure to sync up with @sugakandrey so that you don't duplicate each others work. Maybe communicate on the gitter ensime-server channel? I think the sbt one would be more important than this.

The clients won't need to read the (new) server config files, but the clients will still need to read the `.ensime` files. There might be (or should be) a separate ticket to make the format of `.ensime` optional.

My hope with "server config files" (as opposed to "project config files") is that we can introduce a bunch of user preferences there, that are currently hard coded or hacked with system properties.

yyadavalli:
Yup, agreed on everything.

The clients can determine whether to read a Sexp or HOCON `.ensime` with their own preferences. They will optionally communicate the format to ensime-server at startup, with the default formats based on protocol. 

I can think of a couple of system properties that can be moved to a `ServerConfig` object, I am sure we can provide more options as we move forward. `ServerConfig` does not involve the clients at all, we read it off and persist to project `:root-dir` for each instance of the server.

yyadavalli:
Any thoughts on using typesafehub/config

fommil:
yeah, it's a java library... I had assumed we'd use it :smile:

but we might want to write a little typeclass derivation thing using shapeless so we don't need to write custom parsers. might be fun... have you seen my shapeless for mortals workshop video? https://github.com/fommil/shapeless-for-mortals

I am aware of a few projects that do this, but they use their own macros and we're trying to reduce external dependencies.

yyadavalli:
[...]

PR #1687, opened 2016-12-12
convert hard coded system properties into configuration classes

hardmettle:
This is to make .server file mandatory as part of #1346 . Added a new `EnsimeServerConfig` on the lines of `EnsimeConfig` . The `parse` method is left unimplemented and the WIP . @fommil please take a look .

### DIFF
--- a/api/src/main/scala/org/ensime/api/config.scala
+++ b/api/src/main/scala/org/ensime/api/config.scala
@@ -3,7 +3,6 @@
 package org.ensime.api
 
 import java.io.File
-import scala.util.Properties._
 
 // there is quite a lot of code in this file, when we clean up the
 // config file format so that a lot of these hacks are no longer
@@ -34,10 +33,6 @@ final case class EnsimeConfig(
   // some marshalling libs (e.g. spray-json) might not like extra vals
   val modules = subprojects.map { module => (module.name, module) }.toMap
 
-  def compileClasspath: Set[File] = modules.values.toSet.flatMap {
-    m: EnsimeModule => m.compileDeps ++ m.testDeps
-  } ++ (if (propOrFalse("ensime.sourceMode")) List.empty else targetClasspath)
-
   def targetClasspath: Set[File] = modules.values.toSet.flatMap {
     m: EnsimeModule => m.targets ++ m.testTargets
   }
@@ -55,7 +50,6 @@ final case class EnsimeConfig(
 
   def scalaLibrary: Option[File] = allJars.find(_.getName.startsWith("scala-library"))
 }
-
 final case class EnsimeModule(
     name: String,
     targets: List[File],
@@ -101,3 +95,5 @@ final case class EnsimeProject(
 ) {
   sources.foreach(f => require(f.exists, "" + f + " is required but does not exist"))
 }
+
+final case class EnsimeServerConfig(shutDownOnDisconnect: Boolean, test: Boolean, exitAfterIndex: Boolean, disableSourceMonitoring: Boolean, disableClassMonitoring: Boolean, sourceMode: Boolean, legacyJarurls: Boolean, protocol: String, ENSIME_EXPERIMENTAL_H2: String = "jdbc:h2:file:", ENSIME_SKIP_JRE_INDEX: Boolean, parallelThread: Int)
--- a/api/src/main/scala/org/ensime/api/config.scala
+++ b/api/src/main/scala/org/ensime/api/config.scala
@@ -96,4 +96,4 @@ final case class EnsimeProject(
   sources.foreach(f => require(f.exists, "" + f + " is required but does not exist"))
 }
 
-final case class EnsimeServerConfig(shutDownOnDisconnect: Boolean, test: Boolean, exitAfterIndex: Boolean, disableSourceMonitoring: Boolean, disableClassMonitoring: Boolean, sourceMode: Boolean, legacyJarurls: Boolean, protocol: String, ENSIME_EXPERIMENTAL_H2: String = "jdbc:h2:file:", ENSIME_SKIP_JRE_INDEX: Boolean, parallelThread: Int)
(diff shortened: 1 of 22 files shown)


fommil:
thanks!

We don't want to make this mandatory.

The initial implementation should set it up with the following fallback rules, which can be achieved using the typesafe config library:

- read the legacy system properties
- read `.ensime-server` from the same folder as the `.ensime` (if it exists), which itself is passed as a system property (this is not part of the server config)
- read `~/.ensime-server`

yyadavalli:
Good start !!

Adding to @fommil suggestions, add a XDG compliant location for the global fallback option would be good and also system properties should be the last fallback i.e.

1.  read .ensime-server from the same folder as the .ensime (if it exists), which itself is passed as a system property (this is not part of the server config)

fommil:
Yup, that's better

hardmettle:
@fommil sorry for  delayed progress.I have created the fallback mechanism for server file, but there are are several tests failing.I have tried fixing lot of them but I will need some assistance from here on from you. Can you please take a look when you can.

yyadavalli:
Looks like you are missing an implicit format for `EnsimeServerConfig` also the `parse` method is undefined causing the implicit derivation to fail

fommil:
Thanks! I'll try to take a look

api/src/main/scala/org/ensime/api/config.scala
```diff
@@ -101,3 +95,5 @@ final case class EnsimeProject(
 ) {
   sources.foreach(f => require(f.exists, "" + f + " is required but does not exist"))
 }
+
+final case class EnsimeServerConfig(shutDownOnDisconnect: Boolean, test: Boolean, exitAfterIndex: Boolean, disableSourceMonitoring: Boolean, disableClassMonitoring: Boolean, sourceMode: Boolean, protocol: String, ENSIME_EXPERIMENTAL_H2: String = "jdbc:h2:file:", ENSIME_SKIP_JRE_INDEX: Boolean, parallelThread: Int)
```
fommil: a new field on a new line is much easier to read

core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
```diff
@@ -31,6 +32,7 @@ trait IsolatedAnalyzerFixture
     withVFS { implicit vfs =>
       withTestKit { testkit =>
         import testkit._
+        implicit val s = EnsimeConfigProtocol.parse("")
```
fommil: why are you doing this? It'll throw an exception. But there is already an implicit EnsimeConfig in scope

core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
```diff
@@ -22,7 +22,7 @@ trait EnsimeConfigFixture {
 
   def copyTargets: Boolean = true
 
-  def withEnsimeConfig(testCode: EnsimeConfig => Any): Any
+  def withEnsimeConfig(testCode: (EnsimeConfig) => Any): Any
```
fommil: why did you add parentheses here?

core/src/main/scala/org/ensime/config/EnsimeConfigProtocol.scala
```diff
@@ -28,17 +27,18 @@ object EnsimeConfigProtocol {
   private implicit val moduleFormat: SexpFormat[EnsimeModule] = cachedImplicit
   private implicit val configFormat: SexpFormat[EnsimeConfig] = cachedImplicit
 
-  def parse(config: String): EnsimeConfig = {
+  def parse(config: String, serverConfig: String): EnsimeConfig = {
```
fommil: why would you want to parse both files at the same time?

That doesn't matter. There is no need to parse anything for the serverconfig at this time, the construction of the object should be done by reading system properties at this stage, e.g. in a convenience method in `util`. After that works, we can look at creating a HOCON parser (**not** an s-expression parser).

yyadavalli: @hardmettle if you need the server configuration for parsing `.ensime` , construct a `EnsimeServerConfig` either by parsing it or from system properties.

Adding a `sexp` parser for `EnsimeServerConfig` should be as simple as adding an implicit `SexpFormat` for it.

fommil: I think we want a HOCON file though, I'm undecided. The s-expression format is difficult for a lot of editors to handle... so I'm even thinking that `.ensime` should be `.ensime.conf` in HOCON format.

yyadavalli: I thought we can support both. 
I started a writing the parser for HOCON about a month ago and think got stopped, parsing Lists.
Might be @hardmettle  can pick it up if he gets to it first

yyadavalli: We could also use something like https://github.com/ccadllc/cedi-config but that adds another scala dependency

fommil: @yyadavalli we don't need to write a parser, we can just use a minimal shapeless impl and typesafe config. We don't need to support sealed traits or anything at this stage just an extremely small subset of https://github.com/melrief/pureconfig

yyadavalli: That's what I was working on. Same confusion as before.

fommil: ah ok! have a look in s-express's collection formats for some hints

yyadavalli: Thanks, will do, just not able to find the time these days.

fommil: I know the feeling... but I have some light at the end of the tunnel here.

yyadavalli:
Hi @hardmettle can I pull your changes into my branch and start a new PR

hardmettle:
@yyadavalli @fommil Sorry folks, I have been away for quite a while. Would like to restart on this again. I havent pulled the changes in a long time. Need to revisit the progress.

yyadavalli:
@hardmettle It is still relevant, you can pick it up from where you left it. Just rebase the recent changes and we are good.

hardmettle:
@yyadavalli Will do a rebase, thanks for the clarification.

hardmettle:
@fommil @yyadavalli Closing this PR as it is way behind 2.0 and rebasing is turning out to be a pain. Will start afresh.

PR #1689, opened 2016-12-13
enable the emacs tests again

### LINKED ISSUE #1686: re-enable the emacs tests
now that they are wired up correctly, the emacs tests are failing... I think it's probably a regression in the server startup logic.

I'm confident the problem is actually a bad assertion in the test, not a real regression, so I disabled them. But this is obviously bad, so they need to come back real soon.

fommil:
close #1686 I don't know what the issue was, but it passes just fine for me locally.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -21,19 +21,19 @@ build:
         exit 1 ;
       fi
     - sbt ++$SCALA_VERSION ";test ;it:test" ;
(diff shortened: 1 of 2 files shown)


fommil:
my fancy thread pruning introduced a bug. lovely.

fommil:
Build is broken anyway...

PR #1692, opened 2016-12-17
CI env specific scalatest tags

sugakandrey:


fommil:
I wonder how sbt figures out how to parse multiple args like the test/it ones here.

sugakandrey:
hmm, looks like windows doesn't understand commands like 
```
sbt ";test-only * -- -l org.ensime.util.IgnoreOnAppVeyor ;it:test-only * -- -l org.ensime.util.IgnoreOnAppVeyor"
```
It just goes into sbt prompt. Sometimes I even wish I had windows installation available to me somewhere for situations like this :)

fommil:
I'm surprised sbt can parse it. Maybe doing unit tests followed by integration is OK now.

Issue #1662, opened 2016-11-27
clean up akka logs during testing


fommil:
tracked upstream as https://github.com/akka/akka/issues/21895

we could always just monkey patch the file that writes out to stdout...

Issue #1683, opened 2016-12-11
FileWatcherSpec hangs on Windows


fommil:

fommil:
I wonder if this might be the same JDK bug (that they refuse to acknowledge) that we seen in `CompileTimingSpec`... there will be an accessdenied exception in the logs if so.

fommil:
maybe we're being too aggressively parallel with these tests in CI

fommil:
ignoring for now. When this does hang, we get thread dumps like the above. I suspect we're not forcing the `monitorThread` to die on shutdown. Thoughts @lxol ?

PR #1694, opened 2016-12-19
Ignore tests on ci

fommil:
replaces #1692


fommil:
closes #1662

fommil:
and try to elleviate #1683

project/EnsimeBuild.scala
```diff
@@ -234,7 +234,7 @@ object EnsimeBuild {
 
   private def akkaVersion: Def.Initialize[String] = Def.setting {
     CrossVersion.partialVersion(scalaVersion.value) match {
-      case Some((2, minor)) if minor >= 11 => "2.4.14"
+      case Some((2, minor)) if minor >= 11 => "2.4.16"
```
sugakandrey: Yay! No more noisy logging :tada:

fommil:
hmm, I see the same windows hanging nonsense that you seen.

sugakandrey:
yeah, looks like windows doesn't like test run's arguments at all. I think even 
```
it:test-only * -- -oD
```
behaved the same way for me.

fommil:
could be the escaping rules, maybe a dash or the asterix is being expanded...

fommil:
this used to work `it:testOnly -- -l Debugger`. I think the `*` is both superflous and confusing on Windows.

sugakandrey:
@fommil https://github.com/sbt/sbt-launcher-package/pull/122 check this out

fommil:
even if the CIs still timeout, I'll merge this... it's an improvement already.

fommil:
file watchers hanging on windows... I'll ignore them for now

fommil:
That seems to have helped. Might be able to remove the ignore tags

fommil:
Just server startup causing trouble now

Issue #1557, opened 2016-07-10
remove tuples from the API


fommil:
they result in really ugly wire formats, it's much better to have named fields.

fommil:
the old SWANK protocol shouldn't change, but we should make sure that the JERKY consumers are ready for any changes (vim/atom/vscode) so a PR could sit for a few weeks

PR #1698, opened 2016-12-26
Refactoring/do not use sexp format with q marks in impl

lukasz-golebiewski:


fommil:
heh, this is a horrible file, I don't envy you poking around it.

lukasz-golebiewski:
Actually I was looking for the tuples mentioned here: https://github.com/ensime/ensime-server/issues/1557
@fommil Which tuples do you mean exactly?

fommil:
any tuples in these files https://github.com/ensime/ensime-server/tree/2.0/api/src/main/scala/org/ensime/api

e.g. ParamSectionInfo https://github.com/ensime/ensime-server/blob/2.0/api/src/main/scala/org/ensime/api/outgoing.scala#L482

PR #1704, opened 2017-01-09
aggressive timeouts in tests

fommil:
and should speed up the drone CI (if permissions allow)


fommil:
ok, that seems much faster than normal on drone... I'll try to wrap this up and we're good to merge.

fommil:
could be my imagination, previous builds have been around 38mins and this one is 35mins. Still, it saves the SSDs...

fommil:
those builds only took 25 mins... Wnidows is still flakey

PR #1707, opened 2017-01-21
fix BasicWorkflow DocSig assertion

sugakandrey:


core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -121,7 +121,10 @@ class BasicWorkflow extends EnsimeSpec
           expectMsg(Some(intDocSig))
 
           project ! intDocSig
-          expectMsgType[StringResponse].text should endWith("/index.html#scala.Int")
+          expectMsgType[StringResponse].text should (
+            endWith("/index.html#scala.Int") or // <= 2.11
```
fommil: this `or` stuff is awesome. I never knew about it until you started using it

fommil:
timeouts in the debugtest, transient. Probably just need to bump the timeouts or scaling factor.

PR #1709, opened 2017-01-23
Recover from extractSymbolsFromClassOrJar method when indexing bases

jcazevedo:
This PR enables recovering from a possibly failed `Future` in the `extractSymbolsFromClassOrJar` method when indexing bases. A `Future` would fail in this case if the method attempted to open a file that wasn't a jar (`.so`, `.dll` can easily exist in `java.library.path`). Since we're sequencing the `Future`s and only commiting to index on success, the presence of an unindexable file would make the whole process fail.

I think I can create a test that exposes this behaviour, but I would like to know first if you're OK with the general direction on this.

### DIFF
--- a/core/src/main/scala/org/ensime/indexer/SearchService.scala
+++ b/core/src/main/scala/org/ensime/indexer/SearchService.scala
@@ -133,7 +133,11 @@ class SearchService(
         val check = FileCheck(base)
         val indexed = extractSymbolsFromClassOrJar(base).flatMap(persist(check, _, commitIndex = false, boost = boost))
         indexed.onComplete { _ => semaphore.release() }
-        indexed
+        indexed.recover {
+          case t: Throwable =>
+            log.warn(s"Failed to index ${base.getName()}.", t)
+            0
+        }
       }
     }
 


fommil:
Hi, how do you expect a non-jar file to end up here. Typically we prefer to fail if the input in bad, so the preferred approach would be to just not have bad entries in the .ensime file in the first place.

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -133,7 +133,11 @@ class SearchService(
         val check = FileCheck(base)
         val indexed = extractSymbolsFromClassOrJar(base).flatMap(persist(check, _, commitIndex = false, boost = boost))
         indexed.onComplete { _ => semaphore.release() }
-        indexed
+        indexed.recover {
+          case t: Throwable =>
+            log.warn(s"Failed to index ${base.getName()}.", t)
```
fommil: I think I'd need to see a convincing example of when this fails to understand why it needs to be handled here instead of caught earlier on. I don't like adding in arbitrary recovery code when we can catch the problems before doing any work.

jcazevedo:
You're right. I think `sbt-ensime` is blindly adding everything under the `/lib` directory to the `.ensime` file. I'll try to fix this on the [ensime/ensime-sbt](https://github.com/ensime/ensime-sbt) end.

fommil:
thanks!

PR #1715, opened 2017-01-29
Avoid passing NoSymbol to docSig.

### LINKED ISSUE #1713: MatchError on NoSymbol in askDocSignatureAtPoint
This seems to completely kill my ensime server (no requests are served anymore). I don't see why that should be the case though, so maybe the error is on my side. But nevertheless ensime should be a bit more robust when things aren't found by the PC.

```
scala.MatchError: <none> (of class scala.reflect.internal.Symbols$NoSymbol)
	at org.ensime.core.SymbolToFqn$class.toFqn(SymbolToFqn.scala:89)
	at org.ensime.core.RichPresentationCompiler.toFqn(RichPresentationCompiler.scala:280)
	at org.ensime.core.DocFinding$class.javaFqn(DocFinding.scala:55)
	at org.ensime.core.DocFinding$class.linkName(DocFinding.scala:88)
	at org.ensime.core.DocFinding$class.docSig$1(DocFinding.scala:115)
	at org.ensime.core.DocFinding$class.docSignature(DocFinding.scala:117)
	at org.ensime.core.RichPresentationCompiler.docSignature(RichPresentationCompiler.scala:280)
	at org.ensime.core.RichCompilerControl$$anonfun$askDocSignatureAtPoint$1$$anonfun$apply$3.apply(RichPresentationCompiler.scala:89)
	at org.ensime.core.RichCompilerControl$$anonfun$askDocSignatureAtPoint$1$$anonfun$apply$3.apply(RichPresentationCompiler.scala:89)
	at scala.Option.flatMap(Option.scala:171)
	at org.ensime.core.RichCompilerControl$$anonfun$askDocSignatureAtPoint$1.apply(RichPresentationCompiler.scala:89)
	at org.ensime.core.RichCompilerControl$$anonfun$askDocSignatureAtPoint$1.apply(RichPresentationCompiler.scala:89)
	at scala.tools.nsc.util.InterruptReq.execute(InterruptReq.scala:26)
	at scala.tools.nsc.interactive.Global$$anonfun$pollForWork$1.apply$mcV$sp(Global.scala:457)
	at scala.util.control.Breaks.breakable(Breaks.scala:38)
	at scala.tools.nsc.interactive.Global.pollForWork(Global.scala:431)
	at scala.tools.nsc.interactive.PresentationCompilerThread.run(PresentationCompilerThread.scala:22)
```

dragos:
Fix #1713.


dragos:
Are tests flaky? At least, the failed tests don't seem related to my changes.

sugakandrey:
Yep,  these tests are known to be flaky.

fommil:
Cool I'll take a look.

Hmm, thought we caught the flakey tests. Obvs not.

fommil:
Arg, yeah this one is flakey: should shutdown if preferred HTTP port is not available *** FAILED

core/src/main/scala/org/ensime/core/DocFinding.scala
```diff
@@ -105,16 +105,20 @@ trait DocFinding { self: RichPresentationCompiler with SymbolToFqn =>
...
-        DocSig(linkName(ownerAtSite, java), Some(signatureString(sym, java)))
+        Some(DocSig(linkName(ownerAtSite, java), Some(signatureString(sym, java))))
       } else
-        DocSig(linkName(sym.tpe.typeSymbol, java), None)
+        sym.tpe.typeSymbol.toOption.map(tsym => DocSig(linkName(tsym, java), None))
+    }
+
+    (docSig(java = false), docSig(java = true)) match {
+      case (Some(scalaSig), Some(javaSig)) => Some(DocSigPair(scalaSig, javaSig))
+      case _ => None
```
fommil: it seems unfortunate that we need to handle this case, can you think of a unit test that would exercise this codepath?

fommil:
thanks! Merging for now, but I would really like a follow up with a unit test if you can come up with one please

core/src/main/scala/org/ensime/core/DocFinding.scala
dragos: For me it seems to happen every time I hover over an arrow, for example in
```
def createUsageMsg(cond: Setting => Boolean): String = {..}
```

dragos: I'll check if I can come up with a unit test for that scenario, but probably not before the weekend.

PR #1721, opened 2017-02-11
use a pre-warmed index in intg tests

### LINKED ISSUE #1720: ideas for integration tests performance improvement
Our CI times are becoming crazy, we need to make it faster.

Integration test times dominate, followed by compile time.

And in the integration tests, performing indexing is one of the slowest bits (it is much worse in the graph branch).

If we could do what we do in the emacs tests, and pre-build a basic DB/index with the java/scala standard libraries, that may speed up the CI. Integration tests could opt-in to use a pre-indexed DB.

We could either do it in code or (my preference) use #1690 (adding the advantage of actually testing this feature). A complexity is that we need to generate a `.ensime` for the basic project (in addition to the `.ensime` for the main build) and run the `indexAndExit` command. We could automate this with an sbt task `ensimeServerIntgTestPrep` or similar.

fommil:
close #1720

### DIFF
--- a/util/src/main/scala/org/ensime/util/path.scala
+++ b/util/src/main/scala/org/ensime/util/path.scala
@@ -9,6 +9,7 @@ import java.nio.file.attribute.BasicFileAttributes
 import java.util.{ Collections, EnumSet }
 
 import scala.collection.JavaConverters._
+import scala.util.Try
 
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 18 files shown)


util/src/main/scala/org/ensime/util/path.scala
```diff
@@ -28,6 +28,58 @@ package object path {
     def /(child: String): Path = path.resolve(child)
     def canon: Path = try path.normalize.toRealPath() catch { case e: IOException => path.normalize() }
     def isFile: Boolean = path.toFile.isFile
+    def isDirectory: Boolean = Files.isDirectory(path)
+    def copyDirTo(to: Path): Unit = {
```
fommil: I should really write some unit tests for these methods.

fommil:
hmm, doesn't seem to have made much of a dent in the travis build time :cry:

fommil:
mmm, JVM crash on the 2.10 build... wtaf?

fommil:
JVM crashes in the 2.10 build, I have no idea

PR #1722, opened 2017-02-11
Adapt SimpleLucene to use a Future-based API

### LINKED ISSUE #1481: wrap lucene with an async API

iravid:
This commit adds an internal `FixedThreadPoolExecutor` to `SimpleLucene`
to be used for executing blocking operations against Lucene. The API is
also converted to return `Future`s.

An implicit `ExecutionContext` was also added to `IndexService`'s
constructor, used for transforming futures returned from `SimpleLucene`.

Resolves #1481.

### DIFF
--- a/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
+++ b/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
@@ -215,7 +215,7 @@ class SearchServiceSpec extends EnsimeSpec
     val query = new DisjunctionMaxQuery(
       terms.map(service.index.buildTermClassMethodQuery).asJavaCollection, 0f
     )
-    val fqns = lucene.search(query, 10).map(_.get("fqn"))
+    val fqns = lucene.search(query, 10).futureValue.map(_.get("fqn"))
     fqns.distinct should ===(fqns)
   }
 }
--- a/core/src/main/scala/org/ensime/indexer/IndexService.scala
+++ b/core/src/main/scala/org/ensime/indexer/IndexService.scala
@@ -6,7 +6,6 @@ import java.io.FileNotFoundException
 import java.nio.file.Path
 
 import scala.collection.JavaConverters._
-
 import akka.event.slf4j.SLF4JLogging
 import org.apache.commons.vfs2.FileObject
 import org.apache.lucene.document.{ Document, TextField }
@@ -19,6 +18,8 @@ import org.apache.lucene.analysis.core.KeywordAnalyzer
 import org.ensime.indexer.lucene._
 import org.ensime.util.list._
 import shapeless.Typeable
+
+import scala.concurrent.{ ExecutionContext, Future }
 
 object IndexService extends SLF4JLogging {
 
@@ -79,7 +80,7 @@ object IndexService extends SLF4JLogging {
     build()
 }
 
-class IndexService(path: Path) {
+class IndexService(path: Path)(implicit ec: ExecutionContext) {
   import org.ensime.indexer.IndexService._
 
   private val analyzers = Map(
@@ -94,7 +95,7 @@ class IndexService(path: Path) {
     1 - .25f * nonTrailing$s
   }
 
-  def persist(check: FileCheck, symbols: List[FqnSymbol], commit: Boolean, boost: Boolean): Unit = {
+  def persist(check: FileCheck, symbols: List[FqnSymbol], commit: Boolean, boost: Boolean): Future[Unit] = {
     val f = Some(check)
     val fqns: List[Document] = symbols.map {
       case FqnSymbol(_, _, _, fqn, _, _, _, _) if fqn.contains("(") => MethodIndex(fqn, f).toDocument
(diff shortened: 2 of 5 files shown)


core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -215,7 +215,7 @@ class SearchServiceSpec extends EnsimeSpec
     val query = new DisjunctionMaxQuery(
       terms.map(service.index.buildTermClassMethodQuery).asJavaCollection, 0f
     )
-    val fqns = lucene.search(query, 10).map(_.get("fqn"))
+    val fqns = lucene.search(query, 10).futureValue.map(_.get("fqn"))
```
fommil: heh, I did not know about `futureValue` in scalatest. That's very useful.

core/src/main/scala/org/ensime/indexer/IndexService.scala
```diff
@@ -79,7 +80,7 @@ object IndexService extends SLF4JLogging {
     build()
 }
 
-class IndexService(path: Path) {
+class IndexService(path: Path)(implicit ec: ExecutionContext) {
```
fommil: I think we want this to manage its own EC, like the H2

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -149,7 +148,7 @@ class SearchService(
 
     def commitIndex(): Future[Unit] = {
       log.debug("committing index to disk...")
-      val i = Future { blocking { index.commit() } }
+      val i = index.commit()
```
fommil: I don't think this needs to be outside the for comprehension anymore

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -173,7 +172,7 @@ class SearchService(
   def refreshResolver(): Unit = resolver.update()
 
   def persist(check: FileCheck, symbols: List[FqnSymbol], commitIndex: Boolean, boost: Boolean): Future[Int] = {
-    val iwork = Future { blocking { index.persist(check, symbols, commitIndex, boost) } }
+    val iwork = index.persist(check, symbols, commitIndex, boost)
```
fommil: for comp makes more sense here now

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -277,7 +276,7 @@ class SearchService(
   def delete(files: List[FileObject]): Future[Int] = {
     // this doesn't speed up Lucene deletes, but it means that we
     // don't wait for Lucene before starting the H2 deletions.
-    val iwork = Future { blocking { index.remove(files) } }
+    val iwork = index.remove(files)
```
fommil: ditto

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -286,9 +285,9 @@ class SearchService(
   def fileRemoved(f: FileObject): Unit = fileChanged(f)
   def fileAdded(f: FileObject): Unit = fileChanged(f)
 
-  def shutdown(): Future[Unit] = {
-    db.shutdown()
-  }
+  def shutdown(): Future[Unit] = Future.sequence {
+    List(db.shutdown(), index.shutdown())
```
fommil: heh, I actually **just** added a `index.shutdown` in my PR too because I noticed it was missing :smile:

core/src/main/scala/org/ensime/indexer/lucene/SimpleLucene.scala
```diff
@@ -16,12 +17,22 @@ import org.apache.lucene.util._
...
+import scala.concurrent.{ ExecutionContext, Future, blocking }
 
 object SimpleLucene {
   // from DirectDocValuesFormat.MAX_SORTED_SET_ORDS = 2147483391 but
   // it's kind of an insane number, so let's pick something halfway
   // sensible.
   val MaxResults = 10000
+
+  // our fallback analyzer
+  class LowercaseAnalyzer extends Analyzer {
```
fommil: how come you moved this out?

core/src/main/scala/org/ensime/indexer/lucene/SimpleLucene.scala
```diff
@@ -43,25 +54,22 @@ object SimpleLucene {
  * filtering rules based on tags.
  */
 class SimpleLucene(path: Path, analyzers: Map[String, Analyzer]) extends SLF4JLogging {
+  import SimpleLucene._
+
   Files.createDirectories(path)
 
+  private val executorService = Executors.newFixedThreadPool(8)
```
fommil: right ok, so you put the EC here and then the IndexService is just for combining results, I see. Ok.

core/src/main/scala/org/ensime/indexer/lucene/SimpleLucene.scala
```diff
@@ -108,24 +119,42 @@ class SimpleLucene(path: Path, analyzers: Map[String, Analyzer]) extends SLF4JLo
...
-  def update(delete: Seq[Query], create: Seq[Document], commit: Boolean = true): Unit = {
-    if (delete.nonEmpty) {
-      writer.deleteDocuments(delete.toArray: _*)
-      if (commit) writer.commit()
-    }
-
-    if (create.nonEmpty) {
-      create foreach { doc =>
-        writer addDocument doc
+  def update(delete: Seq[Query], create: Seq[Document], commit: Boolean = true): Future[Unit] =
```
fommil: if you add the `Future` onto the method declaration line, you can dramatically reduce the diff size (and then avoid getting the blame for writing the code in here!)

core/src/main/scala/org/ensime/indexer/lucene/SimpleLucene.scala
```diff
@@ -108,24 +119,42 @@ class SimpleLucene(path: Path, analyzers: Map[String, Analyzer]) extends SLF4JLo
...
-  def delete(queries: Seq[Query], commit: Boolean = true): Unit = update(queries, Nil, commit)
+  def create(docs: Seq[Document], commit: Boolean = true): Future[Unit] = update(Nil, docs, commit)
+  def delete(queries: Seq[Query], commit: Boolean = true): Future[Unit] = update(queries, Nil, commit)
 
   // for manual committing after multiple insertions
-  def commit(): Unit = writer.commit()
+  def commit(): Future[Unit] = Future(blocking(writer.commit()))
+
+  def shutdown(): Future[Unit] = {
+    executionContext.shutdown()
```
fommil: we are probably missing these in the DB code

core/src/main/scala/org/ensime/indexer/lucene/SimpleLucene.scala
```diff
@@ -108,24 +119,42 @@ class SimpleLucene(path: Path, analyzers: Map[String, Analyzer]) extends SLF4JLo
...
+    executorService.shutdown()
+
```
fommil: oh, you're missing a `writer.close` in here.

core/src/main/scala/org/ensime/indexer/lucene/SimpleLucene.scala
iravid: It has no dependencies on any members or types defined in the class, so made more sense to me to make it an independent helper class. WDYT?

core/src/main/scala/org/ensime/indexer/SearchService.scala
iravid: Good point!

iravid:
@fommil thanks for the comments! Pushed the tweaks

fommil:
great stuff! This is now released

iravid:
🎉🎉

PR #1725, opened 2017-02-13
only run drone

fommil:
(testing AWS agents)


fommil:
experiment with Google Cloud was an utter failure, but I think I managed to start a drone agent on an ECS instance, it just didn't accept any work. I think I need a hand holding from somebody with ECS experience http://discourse.drone.io/t/amazon-spot-instances/263/4

PR #1739, opened 2017-04-03
Remove (broken) semaphore code from IndexingQueueActor.

dangerousben:
Remove (broken) semaphore code from IndexingQueueActor.


fommil:
Travis is crap, i wouldn't trust it. I'll kick off a build when USA is sleeping ;-)

Thanks!

Issue #1643, opened 2016-11-08
remove commons-io, commons-lang and guava


fommil:
there is no reason to use these utils since much of the functionality is now rolled into java 7 or 8 (especially on the IO side) and these are poor substitutes.

We may still wish to use them in the testing projects for assertions on using them from client code, but we shouldn't use them directly in our runtime or test code.

fommil:
the use of the `unescapeHtml` could probably handle a much smaller subset of html things.

pascr:
Will be happy to give that a go

fommil:
cool! starting here is a good place http://ensime.github.io//contributing/

fommil:
btw if you can move us to use `Path` instead of `File` that's always good but be careful about conflicting too much with #1637

pascr:
Starting with commons-lang first.

fommil:
@pascr I completely forgot about the stickers I had in my bag at the hack day. Be sure to ask for some at the next hack day, or email/gitter me your mail address and I'll post some.

fommil:
I've made this a lot easier with my latest PR. We now have a lot of functionality around nio `Path` objects that previously was coming from Guava.

strobe:
@pascr Could you create pull request with your commons-lang and HtmlUtil changes? 
I tried to do same things for #1740, but after I found that your code was updated recently, therefore I don't wish to repeat your work.

pascr:
@strobe sure, will do it later today

fommil:
Awesome

jozi-k:
I was going through "low hanging fruit" issues and this one seems to be already resolved.

PR #1740, opened 2017-04-04
[WIP] remove commons-io and guava 

strobe:
this a WIP PR for #1643.

currently only guava removed, but I will continue working on this PR

### DIFF
--- a/core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
+++ b/core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
@@ -3,11 +3,11 @@
 package org.ensime.indexer
 
 import java.nio.charset.Charset
+import java.nio.file.Files
 
 import scala.concurrent.duration._
 
 import akka.testkit._
-import com.google.common.io.Files
 import org.apache.commons.vfs2._
 import org.ensime.fixture._
 import org.ensime.util._
@@ -175,7 +175,7 @@ class FileWatcherSpec extends EnsimeSpec with TimeLimitedTests
   it should "detect removed parent base directory" taggedAs (Retryable) in
     withVFS { implicit vfs =>
       withTestKit { implicit tk =>
-        val parent = Files.createTempDir().canon
+        val parent = Files.createTempDirectory("ensime").toFile.canon
         val dir = parent / "base"
         dir.mkdirs()
         try {
@@ -245,7 +245,7 @@ class FileWatcherSpec extends EnsimeSpec with TimeLimitedTests
   it should "be able to start up from a non-existent directory" taggedAs (Retryable) in
     withVFS { implicit vfs =>
       withTestKit { implicit tk =>
-        val dir = Files.createTempDir().canon / "root"
+        val dir = Files.createTempDirectory("ensime").toFile.canon / "root"
         try {
           withClassWatcher(dir) { watcher =>
             tk.ignoreMsg {
@@ -278,7 +278,7 @@ class FileWatcherSpec extends EnsimeSpec with TimeLimitedTests
     withVFS { implicit vfs =>
       withTestKit { implicit tk =>
 
-        val parent = Files.createTempDir().canon
+        val parent = Files.createTempDirectory("ensime").toFile.canon
         val dir = parent / "base"
         dir.mkdirs()
         try {
--- a/core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
+++ b/core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
@@ -184,7 +184,7 @@ class FileWatcherSpec extends EnsimeSpec with TimeLimitedTests
               // would be better if this was atomic (not possible from JVM?)
               waitForBaseRegistered(tk)
 
-              parent.tree.reverse.foreach(_.delete())
+              parent.toPath().deleteDirRecursively()
               val createOrDelete: Fish = {
                 case r: BaseRemoved => true
                 case a: BaseAdded => true
@@ -193,7 +193,7 @@ class FileWatcherSpec extends EnsimeSpec with TimeLimitedTests
               tk.fishForMessage()(createOrDelete)
               tk.fishForMessage()(createOrDelete)
             }
-          } finally parent.tree.reverse.foreach(_.delete())
+          } finally parent.toPath().deleteDirRecursively()
         }
       }
     }
--- a/core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
+++ b/core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
@@ -3,7 +3,6 @@
 package org.ensime.indexer
 
 import java.nio.charset.Charset
-import java.nio.file.Files
 
 import scala.concurrent.duration._
 
@@ -13,6 +12,7 @@ import org.ensime.fixture._
 import org.ensime.util._
 import org.ensime.util.file._
 import org.ensime.util.fileobject._
+import org.ensime.util.path._
 import org.ensime.vfs._
 import org.scalatest.concurrent.TimeLimitedTests
 import org.scalatest.tagobjects.Retryable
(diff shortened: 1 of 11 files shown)


core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
```diff
@@ -175,7 +175,7 @@ class FileWatcherSpec extends EnsimeSpec with TimeLimitedTests
   it should "detect removed parent base directory" taggedAs (Retryable) in
     withVFS { implicit vfs =>
       withTestKit { implicit tk =>
-        val parent = Files.createTempDir().canon
+        val parent = Files.createTempDirectory("ensime").toFile.canon
```
fommil: I think we have something to create temp dirs in our utils

server/src/main/scala/org/ensime/server/Server.scala
```diff
@@ -129,7 +127,11 @@ object Server extends AkkaBackCompat {
       throw new RuntimeException(s".ensime file ($ensimeFile) not found")
 
     implicit val config: EnsimeConfig = try {
-      EnsimeConfigProtocol.parse(Files.toString(ensimeFile, Charsets.UTF_8))
+      val source = scala.io.Source.fromFile(ensimeFile, "utf-8")
```
fommil: we definitely have something to do this in our files / paths implicit classes. `scala.io.Source` is ridiculously inefficient btw.

util/src/main/scala/org/ensime/util/file.scala
```diff
@@ -6,11 +6,10 @@ import java.io.{ File => JFile, _ }
 import java.nio.charset.Charset
 import java.util.regex.Pattern
 import java.nio.file.Files
+import scala.io.Source
```
fommil: can you please do this using the NIO library instead of `scala.io.Source`? The latter is **very** inefficient.

fommil: also, you might find that a lot of these methods can just call out to their `path.scala` counterpart (in the same directory).

Ideally I'd like to remove our use of `File` and only use `Path` (actually `EnsimeFile`, which wraps `Path`)

util/src/main/scala/org/ensime/util/file.scala
```diff
@@ -91,7 +90,28 @@ package object file {
      */
     @deprecating("prefer path approaches")
     def tree: Stream[File] = {
-      file #:: GFiles.fileTreeTraverser().breadthFirstTraversal(file).asScala.toStream
```
fommil: we should definitely use NIO for this... again have a look at `path.scala` to see if we've done it already

util/src/main/scala/org/ensime/util/io.scala
```diff
@@ -3,15 +3,19 @@
...
 /**
  * NOTE: prefer NIO via the path utilities.
  */
 package object io {
 
   implicit class RichInputStream(val is: InputStream) extends AnyVal {
-    def toByteArray(): Array[Byte] = ByteStreams.toByteArray(is)
+    def toByteArray(): Array[Byte] = {
+      def inputStreamToByteArray(is: InputStream): Array[Byte] = {
+        Iterator continually is.read takeWhile (-1 !=) map (_.toByte) toArray
```
fommil: nio will be more efficient.

util/src/main/scala/org/ensime/util/io.scala
```diff
@@ -20,11 +24,18 @@ package object io {
      * endeavours to close everything afterward (even on failure).
      */
     def drain(in: InputStream): Unit =
-      try ByteStreams.copy(in, os)
-      finally {
+      try {
+        var buffer = Array.fill[Byte](1024)(0) // size does affect perfomace
```
fommil: I think `Array[Byte](1024)` does the same thing? (typo in comment)

util/src/test/scala/org/ensime/util/FileSpec.scala
```diff
@@ -56,7 +54,11 @@ class FileSpec extends FlatSpec with Matchers {
     out.write("hello".getBytes())
     out.close()
 
-    Files.toString(file, Charsets.UTF_8) shouldBe "hello"
+    val source = scala.io.Source.fromFile(file, "utf-8")
```
fommil: `scala.io.Source`

fommil:
I restarted the travis build, it's a bit flakey.

core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala
```diff
@@ -175,24 +175,26 @@ class FileWatcherSpec extends EnsimeSpec with TimeLimitedTests
...
+        withTempDirPath { tmpDir =>
+          val parent = tmpDir.toFile.canon
+          val dir = parent / "base"
+          dir.mkdirs()
+          try {
+            withClassWatcher(dir) { watcher =>
+              // would be better if this was atomic (not possible from JVM?)
+              waitForBaseRegistered(tk)
+
+              parent.tree.reverse.foreach(_.delete())
```
fommil: perhaps tangential to this PR, but I think we now have a method on `Path` to do this

server/src/main/scala/org/ensime/server/Server.scala
```diff
@@ -129,7 +129,7 @@ object Server extends AkkaBackCompat {
       throw new RuntimeException(s".ensime file ($ensimeFile) not found")
 
     implicit val config: EnsimeConfig = try {
-      EnsimeConfigProtocol.parse(Files.toString(ensimeFile, Charsets.UTF_8))
+      EnsimeConfigProtocol.parse(new String(Files.readAllBytes(ensimeFile.toPath), StandardCharsets.UTF_8))
```
fommil: `.readString` on `Path`

util/src/main/scala/org/ensime/util/file.scala
```diff
@@ -64,20 +64,20 @@ package object file {
...
 
     def createWithParents(): Boolean = {
-      GFiles.createParentDirs(file)
+      file.getParentFile().mkdirs()
       file.createNewFile()
     }
 
     def readLines()(implicit cs: Charset): List[String] = {
-      GFiles.readLines(file, cs).asScala.toList
+      Files.readAllLines(file.toPath(), cs).asScala.toList
```
fommil: `.readLines` in `path.scala`

fommil: (or are you trying to avoid calling `path.scala` from `files.scala`?)

util/src/main/scala/org/ensime/util/io.scala
```diff
@@ -3,15 +3,27 @@
...
  */
 package object io {
 
   implicit class RichInputStream(val is: InputStream) extends AnyVal {
-    def toByteArray(): Array[Byte] = ByteStreams.toByteArray(is)
+    def toByteArray(): Array[Byte] = {
+      val baos = new ByteArrayOutputStream()
+
+      val data = Array.ofDim[Byte](16384)
+      var nRead = is.read(data, 0, data.length)
```
fommil: a typical pattern for this sort of thing in scala is to define an inner method `def read` that does the mutating read and it avoids typing that line twice.

util/src/main/scala/org/ensime/util/io.scala
```diff
@@ -3,15 +3,27 @@
...
+
+      while (nRead != -1) {
+        baos.write(data, 0, nRead)
+        nRead = is.read(data, 0, data.length)
+      }
+
+      baos.flush()
+      baos.toByteArray()
```
fommil: doesn't look like we're closing the input at the end, does that agree with the guava semantics?

fommil: (in which case, maybe we have a leak in the existing code?)

util/src/main/scala/org/ensime/util/io.scala
```diff
@@ -3,15 +3,27 @@
...
+
+      val data = Array.ofDim[Byte](16384)
+      var nRead = is.read(data, 0, data.length)
+
+      while (nRead != -1) {
+        baos.write(data, 0, nRead)
+        nRead = is.read(data, 0, data.length)
+      }
+
+      baos.flush()
```
fommil: I'm pretty sure you don't need the flush.

util/src/main/scala/org/ensime/util/io.scala
```diff
@@ -20,11 +32,20 @@ package object io {
      * endeavours to close everything afterward (even on failure).
      */
     def drain(in: InputStream): Unit =
-      try ByteStreams.copy(in, os)
-      finally {
+      try {
+        var data = Array.ofDim[Byte](16384) // size does affect perfomance
```
fommil: where did you get the 16384 from? Link to the source would be really cool (also, we should probably do this everywhere if it is a known magical speed up)

util/src/main/scala/org/ensime/util/io.scala
```diff
@@ -20,11 +32,20 @@ package object io {
...
+
+        var len: Int = 0
+        def read(): Int = { len = in.read(data); len }
```
fommil: ha! You did it here.

util/src/main/scala/org/ensime/util/path.scala
```diff
@@ -57,6 +57,13 @@ package object path {
       require(path.isDirectory && to.isDirectory)
       Files.walkFileTree(path, EnumSet.of(FileVisitOption.FOLLOW_LINKS), Integer.MAX_VALUE, new CopyDirVisitor(path, to))
     }
+    def copyFileTo(to: Path): Unit = {
+      require(path.isFile)
+      if (!Files.exists(to)) {
+        Files.createDirectories(to.getParent())
```
fommil: will this fail if the parent directory already exists?

util/src/test/scala/org/ensime/util/FileSpec.scala
```diff
@@ -56,7 +58,9 @@ class FileSpec extends FlatSpec with Matchers {
     out.write("hello".getBytes())
     out.close()
 
-    Files.toString(file, Charsets.UTF_8) shouldBe "hello"
+    val lines = Files.readAllLines(file.toPath(), StandardCharsets.UTF_8).asScala.mkString("\n")
```
fommil: `path.scala` has `readLines`

fommil:
looking good! Just a few comments above, and you need to rebase.

util/src/main/scala/org/ensime/util/io.scala
strobe: guava toByteArray also doesn't close stream, see:
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/io/ByteStreams.html#toByteArray(java.io.InputStream)

util/src/main/scala/org/ensime/util/path.scala
strobe: looks like it works well if the parent directory already exists, but maybe is a good idea to add test for this case?

fommil:
great stuff, thanks!

PR #1743, opened 2017-04-11
Remove commons-lang and introduce HtmlUtil to unescape html

pascr:
This small PR is a partial fix for #1643 as it only removes commons-lang by introducing a simple HTML unescape utility. It's based on a minimal unescape map that should hopefully be enough for most ScalaDoc/JavaDoc HTML.

### DIFF
--- a/core/src/test/scala/org/ensime/util/HtmlUtilSpec.scala
+++ b/core/src/test/scala/org/ensime/util/HtmlUtilSpec.scala
@@ -1,0 +1,55 @@
+// Copyright: 2010 - 2017 https://github.com/ensime/ensime-server/graphs
+// License: http://www.gnu.org/licenses/gpl-3.0.en.html
+package org.ensime.util
+
+class HtmlUtilsSpec extends EnsimeSpec {
+
+  "HtmlUtil" should "not unescape plain text" in {
+    val originalEscaped = "plain text"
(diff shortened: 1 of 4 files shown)


fommil:
Looks like you forget to run one of the code variation steps, check .drone.yml to see what to run.

fommil:
validation*

pascr:
Should be all good now

core/src/test/scala/org/ensime/util/HtmlUtilSpec.scala
```diff
@@ -0,0 +1,55 @@
+// Copyright: 2010 - 2017 https://github.com/ensime/ensime-server/graphs
+// License: http://www.gnu.org/licenses/gpl-3.0.en.html
+package org.ensime.util
+
+class HtmlUtilsSpec extends EnsimeSpec {
+
+  "HtmlUtil" should "not unescape plain text" in {
+    val originalEscaped = "plain text"
+    HtmlUtil.unescapeHtml(originalEscaped) should ===(Some(originalEscaped))
```
fommil: do you know about `.value` which will unwrap the `Some` when testing `Option`s?

fommil:
@pascr thanks! Feel free to send a follow up with the `.value` as requested, but it shouldn't block merging.

PR #1746, opened 2017-04-15
2.0 graph

fommil:
alt to #1681 because I can't push to Andrey's branch anymore (looks like a github 2FA bug)

The goal is to get this green on all the CIs and then merge in the changes from 2.0 with the debug source lookup (needs some simple additions to the data model)


fommil:
oh great, it won't even let CI run it because it can't be merged in.

fommil:
I've change the branch protection rules

PR #1748, opened 2017-04-15
Make completionsAt return a Future instead of blocking indefinitely

### LINKED ISSUE #1749: completionsAt blocks indefinitely while waiting for the search service
This is unfriendly to the responsiveness of the Analyzer actor, and can be solved by returning the Future instead of awaiting on it.

iravid:
Resolves #1749.


fommil:
looks good!

PR #1751, opened 2017-04-17
Implement an expand match cases refactoring

### LINKED ISSUE #1729: idris style pattern match generation
Using `allKnownSubtypes` we should offer an api call to auto-generate a pattern match over a sealed family.

For unsealed, we could use the graphpocalypse to find known implementations (plus the obligatory runtime failure at the bottom)

iravid:
This correctly handles case classes, case objects and sealed traits. The indentation handling is so-so.

This will turn this block:
```
val f: Family = ???
f match {
<point>
}
```

To:
```
val f: Family = ???
f match {
  case ctor1(f1, f2, f3) =>
  case ctor2(f1, f2, f3) =>
  ...
}
```
with some permutation of the whitespace.

This is still missing automatic importing of the case constructors, but is useful as it is right now.

### DIFF
--- a/api/src/main/scala/org/ensime/api/common.scala
+++ b/api/src/main/scala/org/ensime/api/common.scala
@@ -88,8 +88,9 @@ object RefactorType {
   case object InlineLocal extends RefactorType('inlineLocal)
   case object OrganizeImports extends RefactorType('organizeImports)
   case object AddImport extends RefactorType('addImport)
+  case object ExpandMatchCases extends RefactorType('expandMatchCases)
 
-  def allTypes = Seq(Rename, ExtractMethod, ExtractLocal, InlineLocal, OrganizeImports, AddImport)
+  def allTypes = Seq(Rename, ExtractMethod, ExtractLocal, InlineLocal, OrganizeImports, AddImport, ExpandMatchCases)
 }
 
 /**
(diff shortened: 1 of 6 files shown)


iravid:
Ah, and looks like `showCode` doesn't exist in 2.10... any ideas?

fommil:
You have a compile error in scala 2.10. Do you see how we're adding `BackCompat` support?

fommil:
@mlangc @sschaef ^^^

fommil:
@dragos you might also like ^^^

iravid:
If someone could suggest an alternative for `showCode` in 2.10 it'd be great :-) or, alternatively, we could try and figure out why is scala-refactoring's API returning a malformed diff, obviating the need to generate the code ourselves.

iravid:
Ok, should hopefully go all green now.

iravid:
... or not

fommil:
this is pretty awesome. Can you please put together a gif of the feature in action? If not, I can do it.

iravid:
Would be great if you could help out with the gif :-)

fommil:
I'll create a GIF and tweet tonight, remind me if I don't!

fommil:
argh, we got some kind of publishing error https://ci.fommil.com/ensime/ensime-server/253/1

iravid:
Is it transient?

________________________________

fommil:
no, I think I know what it is ... I messed up

fommil:
(I'll publish manually in the meantime)

fommil:
@iravid ah, ok, I didn't mess up. sbt 0.13.15 did. *sigh* But I'll have to fix it.

fommil:
hmm, I can't quite get it to work...   `ensime-refactor-diff-handler((:procedure-id 14 :reason "No local value selected." :status failure))`

```scala
        val thing = Option("foo")
        thing match {
           // `C-c C-r e` here
        }
```

fommil:
oh, stale server version. It worked! But it's **very** slow.

PR #1764, opened 2017-04-21
failing test for Java 8 lambda usage crash

yangwen0228:
This PR closes the ticket #1763 .
It's my first time to write a test, and write something in Scala. So, if there is anything wrong, please let me know and help me out.

Thank you in advance!

### DIFF
--- a/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
@@ -390,4 +390,38 @@ class JavaCompilerSpec extends EnsimeSpec with OptionValues
       }
     }
   }
+  it should "support Java 8 JavaFx addListener features" in {
+    withJavaCompiler { (_, config, cc, store, search) =>
+      runForPositionInCompiledSource(config, cc, """
+        | import java.lang.Boolean;
+        | import javafx.application.Application;
+        | import javafx.beans.value.ObservableValue;
+        | import javafx.scene.Scene;
+        | import javafx.scene.control.CheckBox;
+        | import javafx.scene.layout.VBox;
+        | import javafx.stage.Stage;
+        |
+        | public class JavaFxAddListenerTest extends Application {
+        |     @Override
+        |     public void start(Stage primaryStage) {
+        |         VBox root = new VBox();
+        |         CheckBox cb = new CheckBox();
+        |         cb.selectedProperty().addListener((ObservableValue<? extends Bool@0@> ov, Boolean oldValue, Boolean newValue) -> {});
+        |         Scene scene = new Scene(root);
+        |         primaryStage.setScene(scene);
+        |         primaryStage.show();
+        |     }
+        |
+        |     public static void main(String[] args) {
+        |         launch(args);
+        |     }
+        | }
+        """.stripMargin) { (sf, offset, label, cc) =>
+        val info = cc.askCompletionsAtPoint(sf, offset, 0, false)
+        label match {
+          case "0" => forAtLeast(1, info.completions)(_.name shouldBe "Boolean")
+        }
+      }
+    }
+  }
 }
--- a/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
@@ -391,7 +391,7 @@ class JavaCompilerSpec extends EnsimeSpec with OptionValues
     }
   }
 
-  it should "support Java 8 JavaFx addListener lambda features" in {
+  it should "support Java 8 generic type completion in a lambda" in {
     withJavaCompiler { (_, config, cc, store, search) =>
       runForPositionInCompiledSource(config, cc, """
         | import java.lang.Boolean;
@@ -426,7 +426,7 @@ class JavaCompilerSpec extends EnsimeSpec with OptionValues
     }
   }
 
-  it should "support Java 8 JavaFx simple wildcard features" in {
+  it should "support Java 8 generic type completion not in a lambda" in {
     withJavaCompiler { (_, config, cc, store, search) =>
       runForPositionInCompiledSource(config, cc, """
         | import java.lang.Boolean;
--- a/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala
@@ -390,7 +390,8 @@ class JavaCompilerSpec extends EnsimeSpec with OptionValues
       }
     }
   }
-  it should "support Java 8 JavaFx addListener features" in {
+
+  it should "support Java 8 JavaFx addListener lambda features" in {
     withJavaCompiler { (_, config, cc, store, search) =>
       runForPositionInCompiledSource(config, cc, """
         | import java.lang.Boolean;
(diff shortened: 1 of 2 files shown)


fommil:
fantastic! It shouldn't close the ticket until it fixes the bug. Would you like to try to fix this? I suspect you only need to read the `JavaCompiler.scala` file to see what is going on. We're probably not handling some Java 8 feature correctly.

fommil:
@mwielocha any ideas?

fommil:
@yangwen0228 I think you need to rebase to latest 2.0 for the CI to run correctly

fommil:
@yangwen0228 were you able to run the test and see it fail locally? read the http://ensime.org/learning_scala/ page to install sbt and then `M-x sbt-start`, then in the sbt buffer do `ensimeConfig` (wait a while, it'll download deps) then `core/it:testOnly *.JavaCompilerSpec -- -z addListener` to run just your test.

yangwen0228:
@fommil Yes, I can run the test locally by following your method. And an error occurs : 
`[error] Uncaught exception when running org.ensime.core.JavaCompilerSpec: java.lang.StackOverflowError.`

yangwen0228:
@fommil Looks like there is an infinite loop:
```
[info]   at com.sun.tools.javac.comp.Attr.visitLambda(Attr.java:2266)
[info]   at com.sun.tools.javac.tree.JCTree$JCLambda.accept(JCTree.java:1624)
[info]   at com.sun.tools.javac.comp.Attr.attribTree(Attr.java:566)
[info]   at com.sun.tools.javac.comp.Attr.visitLambda(Attr.java:2404)
...
...

```

fommil:
perfect, that's exactly what you were seeing.

yangwen0228:
@fommil It seems that the error is started from:
line 73: `Some(fetchTypeSearchCompletions(defaultPrefix, maxResults, indexer))`
This function return this:
`Some(Future(<not completed>))`
The correct return should be something like this:
`Some(JavacScope[env=Env[AttrContext[Scope[root, primaryStage, super, this]],outer=Env[AttrContext[Scope[javafx.stage.Stage, javafx.scene.layout.VBox, javafx.scene.control.CheckBox, javafx.scene.Scene, javafx.beans.value.ObservableValue, javafx.application.Application, java.lang.Boolean, JavaFxAddListenerTest]]]],starImport=false])`

Any suggestions? Thanks!

fommil:
very interesting. Can you debug further by finding what the value of that =Future= is when it completes?

yangwen0228:
It's the result of the `akka.pattern.Patterns.ask()` function:
 ```
 def fetchTypeSearchCompletions(prefix: String, maxResults: Int, indexer: ActorRef): Future[Option[List[CompletionInfo]]] = {
    val req = TypeCompletionsReq(prefix, maxResults)
    import scala.concurrent.ExecutionContext.Implicits.{ global => exe }
    val askRes = Patterns.ask(indexer, req, Timeout((1 * dilation).seconds))
    askRes.map {
      case s: SymbolSearchResults =>
        s.syms.map { s =>
          CompletionInfo(
            None,
            s.localName, 40, None
          )
        }
      case unknown =>
        throw new IllegalStateException("Unexpected response type from request:" + unknown)
    }.map(Some(_)).recover { case _ => None }
  }

```

**Comment by fommil** [comment]
try sticking a `println` into each of the `case` paths. If the future throws an exception, then it should default to `None` (you won't see the `println` in that case)

**Comment by yangwen0228** [comment]
@fommil The problem is from this file:
./core/src/main/scala/org/ensime/core/javac/ScopeFor.scala 
```

      val statements: List[_ <: StatementTree] = path.getLeaf.getKind match {
        case Kind.BLOCK => path.getLeaf.asInstanceOf[BlockTree].getStatements.asScala.toList
        case Kind.FOR_LOOP => path.getLeaf.asInstanceOf[ForLoopTree].getInitializer.asScala.toList
        case Kind.ENHANCED_FOR_LOOP => List(path.getLeaf.asInstanceOf[EnhancedForLoopTree].getStatement)
        case Kind.CASE => path.getLeaf.asInstanceOf[CaseTree].getStatements.asScala.toList
        case Kind.METHOD => path.getLeaf.asInstanceOf[MethodTree].getParameters.asScala.toList
        case otherwise => Nil
      }

```
The `path.getLeaf.getKind` is `Kind.EXTENDS_WILDCARD` . 
So `statements` is `List()`. 
Then `treePathOption` is `None`. 
And `compilation.trees.getScope(treePathOption.getOrElse(path))` return the origin path. So this is an infinite loop. 
I think we should terminate the function and return a `None` instead?

I am very new to Scala, and don't know how to handle this. I just throw out an exception, and it seems to work on my machine. At least, don't disconnect to ensime-server again.
```
      treePathOption match {
        case Some(newPath) => compilation.trees.getScope(treePathOption.getOrElse(path))
        case _ => throw new IllegalStateException("Unexpected response type from request.")
      }
```
Is this the right way to handle this case?

**Comment by fommil** [comment]
That is some pretty impressive investigation for somebody new to scala! It sends like a nasty hack but I prefer a nasty hack to broken code. We probably need to rethink the control flow and / or add support for extra features.

**Comment by fommil** [comment]
(i.e. yes please add that fix to this PR)

File: core/src/it/scala/org/ensime/core/JavaCompilerSpec.scala [hunk 1]
```diff
@@ -390,7 +390,8 @@ class JavaCompilerSpec extends EnsimeSpec with OptionValues
       }
     }
   }
-  it should "support Java 8 JavaFx addListener features" in {
+
+  it should "support Java 8 JavaFx addListener lambda features" in {
```
fommil: The JavaFx addListener but is probably not relevant, I suspect this is a more general lambda thing.

yangwen0228: Yes, it's a general lambda problem. When there is only `T<? extends T>`, no lambda, it works. But, if this is in a lambda with wildcard `(T<>)`, it fails.

fommil:
the CI fails because it looks like you left a TODO or a FIXME in the code, we prefer those to go into tickets.

fommil:
hmm, and it seems to have broken completions at point. Darn... that's annoying.

yangwen0228:
Sorry for the late reply. So, what shall I do now?
BTW, Ensime doesn't work now, it always take almost 100% CPU and make the Emacs very slow.
The process is this:
1, M-x : ensime
2, Message: failed to connect to port 1975, trying again...
3, Connected to ENSIME speaking protocol 1.9.1, please wait while the project is loaded.
4, CPU -> 100%, very slow ...

It's very strange. It works before, but after I tried the self-built version.
```
 import org.ensime.EnsimeKeys._
 ensimeServerJars in ThisBuild := Seq(file("d:/git/ensime-server/target/scala-2.12/ensime_2.12-2.0.0-SNAPSHOT-assembly.jar"))
```
The strange phonamenon comes. After that, I tried to comment above codes, and use the older configuration again. However, the phonamenon doesn't disappear. Then I tried delete all the .sbt .ivy2 .coursier .m2, and build the Ensime enviroment again. But still not work. So, shall I create a new issue?

fommil:
it's likely because you need to increase the heap for your project http://ensime.org/build_tools/sbt/#customise because of the new graph feature

I've set it to 4g for `ensime-server`

fommil:
sounds like you're running an old version of the server because of https://github.com/ensime/ensime-sbt/issues/321

yangwen0228:
Great! You are right! After I add this line, it works again:
`ensimeJavaFlags in ThisBuild := Seq("-Xss2m", "-Xmx2g", "-XX:MaxMetaspaceSize=512m")`

If we don't add this line, the ensime-server default uses 1GB memory, then the CPU is alway around 100%. Increase the memory to 1.5GB or higher, then it works fine.

fommil:
we'll try to fix that automatically https://github.com/ensime/ensime-sbt/issues/324

you're welcome to submit a PR there too since you're on a winning spree :smile:

yangwen0228:
Many thanks for your help and encouragement. It's my great honor to take part in this great software development. However, I am not available to write a PR in the following 5 days,  because I am on a May Day trip with my families.

fommil:
that's fine :smile: you're the only one experiencing the bug so fix it in your own time. We'll just throw encouragement in your direction. I wish I was able to help.

yangwen0228:
OK😄No problem!

fommil:
FYI, I was thinking about this branch today, so I just checked it out to have a look. No promises that I can help...

fommil:
related? https://netbeans.org/bugzilla/show_bug.cgi?id=227851

fommil:
this looks like netbeans works around it by having their own fork of `javac`? That is ugly. http://hg.netbeans.org/main/nb-javac/rev/2b9b60de0c69

yangwen0228:
@fommil Actually, currently, I switch to use MacbookPro, I don't suffer this issue. The problem seems only to happen in Windows OS.

fommil:
I'm getting it on ArchLinux. It looks like a bug in the JDK :confused: perhaps Apple have pulled that fix into their branch but it's not yet in OpenJDK.

fommil:
I think I'll merge your tests as canary tests.

yangwen0228:
@fommil Oh, really sorry! Actually, in MacOs, this issue still exists. But I write a yasnippet to omit the type.
`(Object obj) -> {...}` as `(obj) -> {...}`

fommil:
@yangwen0228 can you please add me as a collaborator to your branch so I can push some changes to this PR and merge?

fommil:
nevermind, closing in favour of https://github.com/ensime/ensime-server/pull/1837

yangwen0228:
Of course sure! Sorry for the late reply. I was just too tired to go to bed yesterday. It was over 12pm in China time. 😄

PR #1773, opened 2017-05-08
Debounce AskReTypecheck

### LINKED ISSUE #1772: EnsimeServer becomes unresponsive for search-related queries if many files change and compiler is invoked
Continuation of #1770

Video recording:

https://www.youtube.com/watch?v=vf5YKqQlC3E

Action starts at about 3:30. The rest shows a fresh clone, compile, setup, etc. Video has been edited to reduce length.

### Steps to reproduce

1. clone git@github.com:timcharper/marathon.git
1. `git checkout tharper/akka-http-apps-controller-ensime-bug`
1. `sbt test:compile ensimeConfig ensimeServerIndex`
1. Open `./src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala` in emacs
1. Launch ensime (`m-x ensime`)
1. Wait for ensime to launch
1. Type `Try` in the top of some method, and use `C-c C-r t`, verify that it imports properly
1. Undo all changes to the file
1. Back in terminal: `git checkout tharper/akka-http-apps-controller-ensime-bug-2`
1. Now run `sbt test:compile`
1. As the compilation is running, notice in the ensime connection log the following message is repeated very often (multiple times per millisecond):

        17:27:49.264 DEBUG akka://ENSIME/user/ensime-main/project o.e.c.Project - received handled message AskReTypecheck from Actor[akka://ENSIME/user/ensime-main/project#548004105] in state handleRequests

1. When compilation finishes, notice that typechecking still works. Make some invalid change to the file, watch it go red; undo, watch it go green again.
1. Type `Try` again in the top of some method, and try to import the symbol with `C-c C-r t`
1. Emacs becomes unresponsive because ensime does not respond with import suggestions

I think the main culprit is the `AskReTypecheck infinite loop` we get in to. I've seen this happen pretty often and think many of the freeze-up issues I am seeing are due to it.

### Other notes

I have 4gb of ram for my ensime process. At no point does it come close to exceeding that when the issue manifests (watching via VisualVM). This can be seen in the video. The issue does not appear to be heap size related in any sense.

### LINKED ISSUE #1774: EnsimeServer floods the log with `AskRetypeCheck` messages
# Summary

See #1772 for reproduction case and video

This issue can occur when changing a handful of files and then running the compiler. The file watcher triggers as high as 20,000 / second (I'm not exaggerating). This, by default, produces a LOT of log messages, because, by default, each message received by the `Project` actor is logged.

The message looks like this:

```
19:23:42.100 INFO  None o.e.u.Debouncer - Debouncer action AskReTypecheck invoked after maxDelay 20 seconds surpassed; 405220 calls occurred since last invocation.
```

Further, it's a lot of burden on the Akka scheduler, as each of these messages gets scheduled  to be processed 5 seconds in the future. Further, there is a lot of context switching overhead evaluating tiny tasks, consuming resources in the fork join pool work steal logic.

timcharper:
Debounce in a way that is efficient (debouncing happens in the same
thread as invoker, and without locks), and in a way that vastly limits
the log output amount. (received handled message AskReTypecheck from
...)

Resolves #1774

### DIFF
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -7,6 +7,7 @@ import akka.event.LoggingReceive.withLabel
 import org.apache.commons.vfs2.FileObject
 import org.ensime.api._
 import org.ensime.core.debug.DebugActor
+import org.ensime.util.{ Debouncer, Timing }
 import org.ensime.vfs._
 import org.ensime.indexer._
 
@@ -28,7 +29,7 @@ class Project(
     broadcaster: ActorRef,
     implicit val config: EnsimeConfig
 ) extends Actor with ActorLogging with Stash {
-  import context.{ dispatcher, system }
+  import context.system
 
   import FileUtils._
 
@@ -49,11 +50,17 @@ class Project(
   private val searchService = new SearchService(config, resolver)
   private val sourceWatcher = new SourceWatcher(config, resolver :: Nil)
   private val reTypecheck = new FileChangeListener {
-    def reTypeCheck(): Unit = self ! AskReTypecheck
-    def fileAdded(f: FileObject): Unit = reTypeCheck()
-    def fileChanged(f: FileObject): Unit = reTypeCheck()
-    def fileRemoved(f: FileObject): Unit = reTypeCheck()
-    override def baseReCreated(f: FileObject): Unit = reTypeCheck()
+    private val askReTypeCheck =
+      Debouncer.forActor(
+        self,
+        AskReTypecheck,
+        delay = (5 * Timing.dilation).seconds,
+        maxDelay = (20 * Timing.dilation).seconds
+      )
+    def fileAdded(f: FileObject): Unit = askReTypeCheck.call()
+    def fileChanged(f: FileObject): Unit = askReTypeCheck.call()
+    def fileRemoved(f: FileObject): Unit = askReTypeCheck.call()
+    override def baseReCreated(f: FileObject): Unit = askReTypeCheck.call()
   }
   context.actorOf(Props(new ClassfileWatcher(config, searchService :: reTypecheck :: Nil)), "classFileWatcher")
 
@@ -130,10 +137,7 @@ class Project(
   def handleRequests: Receive = withLabel("handleRequests") {
     case ShutdownRequest => context.parent forward ShutdownRequest
     case AskReTypecheck =>
-      Option(rechecking).foreach(_.cancel())
-      rechecking = system.scheduler.scheduleOnce(
-        5 seconds, scalac, ReloadExistingFilesEvent
-      )
+      scalac ! ReloadExistingFilesEvent
     // HACK: to expedite initial dev, Java requests use the Scala API
     case m @ TypecheckFileReq(sfi) if sfi.file.isJava => javac forward m
     case m @ CompletionsReq(sfi, _, _, _, _) if sfi.file.isJava => javac forward m
--- a/util/src/main/scala/org/ensime/util/Debouncer.scala
+++ b/util/src/main/scala/org/ensime/util/Debouncer.scala
@@ -1,0 +1,145 @@
+// Copyright: 2010 - 2017 https://github.com/ensime/ensime-server/graphs
+// License: http://www.gnu.org/licenses/gpl-3.0.en.html
+package org.ensime.util
+
+import akka.actor.{ ActorRef, Scheduler, ActorContext }
+import java.time.Clock
+import java.util.concurrent.atomic.{ AtomicBoolean, AtomicInteger }
+import org.slf4j.LoggerFactory
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+/**
+ * Highly performant generic debouncer
+ *
+ * Note: This is not written as an Actor for performance reasons. Debounced calls are filtered synchronously, in the
+ * caller thread, without the need for locks, context switches, or heap allocations. We use AtomicBoolean to resolve
+ * concurrent races; the probability of contending on an AtomicBoolean transition is very low.
+ *
+ * @param name The name of this debouncer
+ * @param scheduler The scheduler implementation to use; Can accept an akka scheduler a la magnet pattern
+ * @param delay The time after which to invoke action; each time [[Debouncer.call]] is invoked, this delay is reset
+ * @param maxDelay The maximum time to delay the invocation of action()
+ * @param action The action to invoke
+ * @param debounceFirst If false, then the first invocation of call() calls action exactly once. Subsequent calls are debounced.
+ * @param Clock The source from which we get the current time. This input should use the same source. Specified for testing purposes
+ */
+class Debouncer(
+    name: String,
+    scheduler: Debouncer.SchedulerLike,
+    delay: FiniteDuration,
+    maxDelay: FiniteDuration,
+    action: () => Unit,
+    debounceFirst: Boolean = true,
+    clock: Clock = Clock.systemUTC
+) {
+  import Debouncer.logger
+  require(delay <= maxDelay, "delay should be <= maxDelay")
+  private[this] val pending = new AtomicBoolean(false)
+  private[this] val calls = new AtomicInteger(0)
+  private[this] val debouncingActive = new AtomicBoolean(debounceFirst)
+  private val delayMs = delay.toMillis
+  private val maxDelayMs = maxDelay.toMillis
+  @volatile private[this] var lastCallAttempt: Long = clock.millis()
+  @volatile private[this] var lastInvocation: Long = clock.millis()
+
+  /**
+   * Register that the provided action should be called according to the debounce logic
+   */
+  def call(): Unit = if (activateDebouncing) {
+    action()
+  } else {
+    lastCallAttempt = clock.millis()
+    calls.incrementAndGet()
+    if (pending.compareAndSet(false, true)) {
+      scheduler.scheduleOnce(delay) { () => tryActionOrPostpone() }
+    }
+  }
+
+  /** Returns true if debouncing was inactive, and we just activated it */
+  private[this] def activateDebouncing =
+    debouncingActive.compareAndSet(false, true)
+
+  /** See if it's time to invoke the debounce action */
+  private[this] def tryActionOrPostpone(): Unit = {
+    val now = clock.millis()
+    val delaySurpassed = ((now - lastCallAttempt) >= delayMs)
+    val maxDelaySurpassed = ((now - lastInvocation) >= maxDelayMs)
+
+    if (delaySurpassed || maxDelaySurpassed) {
+      lastInvocation = now
+      if (pending.compareAndSet(true, false)) {
+        val foldedCalls = calls.getAndSet(0)
+        if (maxDelaySurpassed)
+          logger.info(s"Debouncer action $name invoked after maxDelay $maxDelay surpassed ($foldedCalls calls)")
+        else if (logger.isDebugEnabled)
+          logger.debug(s"Debouncer action $name invoked after delay $delay surpassed ($foldedCalls calls)")
+        try action()
+        catch {
+          case ex: Throwable => logger.error(s"Debouncer ${name} action resulted in error", ex)
+        }
+      } else
+        logger.error(s"Invalid state in debouncer. Should not have reached here!")
+    } else {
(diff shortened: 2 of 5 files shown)


timcharper:
@fommil per your request on #1677 I added the `debounceFirst` parameter. I'm not sure what its for, but I'm sure you had a good reason for asking it :)

Also, not sure if we should clear debounceFirst state after debounced event fires. We can totally do this. Just depends on the desired behavior. Would like to discuss.

util/src/main/scala/org/ensime/util/Debouncer.scala
```diff
@@ -0,0 +1,118 @@
...
+    name: String,
+    scheduler: Debouncer.SchedulerLike,
+    delay: FiniteDuration,
+    maxDelay: FiniteDuration,
+    action: () => Unit,
+    debounceFirst: Boolean = true,
+    clock: Clock = Clock.systemUTC
+) {
+  require(delay <= maxDelay, "delay should be <= maxDelay")
+  private val logger = LoggerFactory.getLogger(getClass)
```
fommil: it's usually better to put a `Logger` in the companion to avoid excessive heap usage... but in this case it's probably ok since we don't create too many `Debouncer` instances.

util/src/main/scala/org/ensime/util/Debouncer.scala
```diff
@@ -0,0 +1,118 @@
...
+
+import akka.actor.Scheduler
+import java.time.Clock
+import java.util.concurrent.atomic.{ AtomicBoolean, AtomicInteger }
+import org.slf4j.LoggerFactory
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+/**
+ * Generic debouncer implementation
```
fommil: it is kind of weird that this isn't an `Actor` since we use akka everywhere else (I'd prefer to use fs2). Can you please add a comment explaining that writing this as an akka actor has performance problems OR (if you don't think that is completely correct) create a ticket to rewrite this as an actor. Perhaps if we were to disable logging for the debouncer it might be ok.

util/src/main/scala/org/ensime/util/Debouncer.scala
```diff
@@ -0,0 +1,118 @@
...
+  private val delayMs = delay.toMillis
+  private val maxDelayMs = maxDelay.toMillis
+  @volatile private[this] var lastCallAttempt: Long = clock.millis()
+  @volatile private[this] var lastInvocation: Long = clock.millis()
+
+  /**
+   * Register that the provided action should be called according to the debounce logic
+   */
+  def call(): Unit = if (debouncingActive.compareAndSet(false, true)) {
+    // debounceFirst was false
```
fommil: Perhaps `getAndSet(true)` may be easier to reason about... I think the double negative makes it hard to understand. But maybe that makes the code messier. I don't know.

util/src/main/scala/org/ensime/util/Debouncer.scala
```diff
@@ -0,0 +1,118 @@
...
+    val delaySurpassed = ((now - lastCallAttempt) >= delayMs)
+    val maxDelaySurpassed = ((now - lastInvocation) >= maxDelayMs)
+
+    if (delaySurpassed || maxDelaySurpassed) {
+      lastInvocation = now
+      if (pending.compareAndSet(true, false)) {
+        val foldedCalls = calls.getAndSet(0)
+        if (maxDelaySurpassed)
+          logger.info(
+            "Debouncer action {} invoked after maxDelay {} surpassed; {} calls occurred since last invocation.",
```
fommil: let's just use scala interpolation. Protect the `debug` with an `isDebugEnabled`.

util/src/main/scala/org/ensime/util/Debouncer.scala
```diff
@@ -0,0 +1,118 @@
...
+        catch {
+          case ex: Throwable => logger.error(s"Debouncer ${name} action resulted in error", ex)
+        }
+      } else
+        logger.error(s"Invalid state in debouncer. Should not have reached here!")
+    } else {
+      // reschedule at the earliest of origLastInvocation + delayMs or origLastCallAttempt + maxDelayMs
+      // Note: we use Math.max as there's a _very_ small chance lastCallAttempt could advance in another thread, and
+      // result in a negative calcaulation
+      val delay = Math.max(1, Math.min((lastCallAttempt - now + delayMs), (lastInvocation - now + maxDelayMs)))
```
fommil: slightly cleaner as?

```scala
val delay = 1L max (((lastCallAttempt + delayMs) min (lastInvocation + maxDelayMs)) - now)
```

util/src/test/scala/org/ensime/util/DebouncerSpec.scala
```diff
@@ -0,0 +1,98 @@
+// Copyright: 2010 - 2017 https://github.com/ensime/ensime-server/graphs
+// License: http://www.gnu.org/licenses/gpl-3.0.en.html
+package org.ensime.util
+
+import java.time.{ Clock, Instant, ZoneId }
+import org.scalatest._
+import scala.concurrent.duration._
+
+class DebouncerSpec extends FlatSpec with Matchers {
+  trait Fixture {
```
fommil: the typical style has been to use the loan pattern to pass only the things that you want to make visible in the test, and taking config parameters... this exposes rather a lot of fields. Is that intentional?

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -49,11 +50,13 @@ class Project(
   private val searchService = new SearchService(config, resolver)
   private val sourceWatcher = new SourceWatcher(config, resolver :: Nil)
   private val reTypecheck = new FileChangeListener {
-    def reTypeCheck(): Unit = self ! AskReTypecheck
-    def fileAdded(f: FileObject): Unit = reTypeCheck()
-    def fileChanged(f: FileObject): Unit = reTypeCheck()
-    def fileRemoved(f: FileObject): Unit = reTypeCheck()
-    override def baseReCreated(f: FileObject): Unit = reTypeCheck()
+    private val reTypeCheck = Debouncer("AskReTypecheck", system.scheduler, 5.seconds, 20.seconds) { () =>
```
fommil: I'm a bit nervous about non-akka hardcoded time durations because of the slower Windows / OSX build machines that we use. We have used manual dilation (e.g. see `Timing` in utils) where akka doesn't manually manage this. I wonder if perhaps we should do the same here.

fommil:
I suppose theoretically we can now conceivably run the action twice because things are happening in different threads. In practice, it'll happen so rarely as to not matter unless you wanted to reset `debouncingActive` after every delayed `action`... which would have the effect of running the lambda at the start and end of every slug of requests.

fommil:
I'm wondering if there is any way we could have caught this problem with an automated test... maybe simulating a full recompile of `ensime-server` itself? Or perhaps even spotting early warning signs in `CompileTimingTest`

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: I explained in another ticket. The primary reason is performance. Debounced calls are filtered instantaneously with almost zero allocations (except the stack frame allocation). This is meaningful when you consider that I have seen roughly 20,000 invocations / second on average filtered by this logic.

There are a lot of allocations and mutations involved with using an actor.

Also, this is typesafe.

Let me switch it: why use an Actor?

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: I can change it

util/src/main/scala/org/ensime/util/Debouncer.scala
fommil: Yes, I understood the reasons... could you please write that up in the comments so that anybody reading the code can understand why we did it this way. It is unexpected that we wouldn't just write this up as a reusable Actor since all the plumbing of the server is written in akka. It is very bad akka style to use closures because it allows state to leak across the concurrency model, so it requires justification / explanation. I think performance is a good justification.

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: I'll extract to method to make it easier to reason about

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: 👍

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: I don't think that helps; your call

util/src/test/scala/org/ensime/util/DebouncerSpec.scala
timcharper: It's stateless so should be able to adopt whatever style; I'll have a look at the others

core/src/main/scala/org/ensime/core/Project.scala
timcharper: This is why I did the maxDelay of 20.seconds; it'll get called, eventually. But, we can totally make them configurable before. It was hard coded before.

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: I see; we could write an `apply` function which does not takes an actorRef and a message to send.

fommil: I'm fine with this not being akka code, just so long as we document why. I tend to like comments to explain unexpected things, and performance optimisations are almost always unexpected things.

core/src/main/scala/org/ensime/core/Project.scala
fommil: the max delay is an interesting feature in its own right. Throw in batching and resetting of `debounceFirst` and you got yourself a completely general debouncer with every bell and whistle anybody could possibly want!

util/src/main/scala/org/ensime/util/Debouncer.scala
fommil: leave as is then

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: Worth noting... `compareAndSet` _IS_ slightly more efficient, as `getAndSet` simply calls `get`, then `compareAndSet`.

fommil: another way to do this could be to reverse the meaning of the initial boolean flag... then the logic stays the same without double negation. `runFirstMessage` with `debouncingActive = !runFirstMessage`

fommil: or maybe just drop that feature, since we don't seem to need it

timcharper:
@fommil how will the action run twice? I've carefully analyzed this code. It seems obviously correct to me. What am I missing that would allow an action to be run twice?

timcharper:
(Debouncing is activated permanently on first invocation... partly because I wasn't sure what the intended behavior was, and partly because this makes it more difficult to hold the execution guarantees)

core/src/main/scala/org/ensime/core/Project.scala
timcharper: I added the Timing.dilation adjustment

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: Done

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: Done; also added the `actorRef` function.

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: Done

util/src/main/scala/org/ensime/util/Debouncer.scala
timcharper: Done

util/src/test/scala/org/ensime/util/DebouncerSpec.scala
timcharper: Done

timcharper:
The appVeyor build failure seems unrelated

fommil:
I'll rerun it to be sure

timcharper:
Thanks!

PR #1777, opened 2017-05-17
Import type at point for objects without companion.

### LINKED ISSUE #1771: import type at point not working for objects
possibly a regression in #1769 @hzenginx it seems that import suggestions no longer work for `object`. I noticed this when trying to follow https://github.com/ensime/ensime-server/issues/1770#issuecomment-299314401 (but for `CreateRunSpec` not `Clock`) and noticed that import-type-at-point wasn't doing anything.

Maybe we are no longer stripping the final `$`? Mabye this has always been broken.

Maybe we should just do the search on the scala name now that we have the graphpocalypse instead of hacking it with bytecode fqn manipulation...

jozi-k:
Trying to resolve #1771 . I am not particularly happy with the implementation (mutation, too many ifs, magic numbers, list of suggestions shows $ at the end), but could not come with anything better.

Alternatively, @fommil mentioned that scala names can be looked up via graph, but I could not find any evidence of that in the current project code base. Was it part of so called graphocalypse? Maybe @sugakandrey can help to confirm it is possible. 

Will be happy to hear your comments.

### DIFF
--- a/core/src/main/scala/org/ensime/core/Indexer.scala
+++ b/core/src/main/scala/org/ensime/core/Indexer.scala
@@ -25,10 +25,8 @@ class Indexer(
   )
 
   def oldSearchTypes(query: String, max: Int): List[TypeSearchResult] = {
-    def strip(fqn: String): String =
-      if (fqn.endsWith("$")) fqn.replaceAll("[$]^", "")
-      else if (fqn.endsWith("$class")) fqn.replaceAll("[$class]^", "")
-      else fqn
+    // Remove $/$class from the end.
+    def strip(fqn: String): String = fqn.replaceAll("\\$(class)*$", "")
 
     import org.ensime.util.list._
 
--- a/core/src/main/scala/org/ensime/core/Indexer.scala
+++ b/core/src/main/scala/org/ensime/core/Indexer.scala
@@ -6,7 +6,7 @@ import akka.actor._
 import akka.event.LoggingReceive
 import org.ensime.api._
 import org.ensime.indexer.SearchService
-import org.ensime.indexer.graph.FqnSymbol
+import org.ensime.indexer.graph.{ ClassDef, FqnSymbol, Field, Method }
 import org.ensime.model._
 import org.ensime.vfs._
 
@@ -24,25 +24,22 @@ class Indexer(
     LineSourcePositionHelper.fromFqnSymbol(hit)(vfs)
   )
 
-  def oldSearchTypes(query: String, max: Int) = {
-    val scalaNames = scala.collection.mutable.Set.empty[String]
+  def oldSearchTypes(query: String, max: Int): List[TypeSearchResult] = {
+    def strip(fqn: String): String =
+      if (fqn.endsWith("$")) fqn.replaceAll("[$]^", "")
+      else if (fqn.endsWith("$class")) fqn.replaceAll("[$class]^", "")
+      else fqn
 
-    def isUnique(fqn: String): Boolean = {
-      val stripped =
-        if (fqn.endsWith("$")) fqn.dropRight(1)
-        else if (fqn.endsWith("$class")) fqn.dropRight(6)
-        else fqn
-
-      if (scalaNames.contains(stripped)) {
-        false
-      } else {
-        scalaNames.add(stripped)
-        true
-      }
-    }
+    import org.ensime.util.list._
 
     index.searchClasses(query, max)
-      .collect { case name if isUnique(name.fqn) => typeResult(name) }
+      .map {
+        case c: ClassDef => c.copy(fqn = strip(c.fqn))
+        case f: Field => f.copy(fqn = strip(f.fqn))
+        case m: Method => m.copy(fqn = strip(m.fqn))
+      }
+      .distinctBy(_.fqn)
+      .map(typeResult)
   }
 
   private val typeDecls: Set[DeclaredAs] = Set(DeclaredAs.Class, DeclaredAs.Trait, DeclaredAs.Object)
--- a/core/src/main/scala/org/ensime/core/Indexer.scala
+++ b/core/src/main/scala/org/ensime/core/Indexer.scala
@@ -24,10 +24,26 @@ class Indexer(
     LineSourcePositionHelper.fromFqnSymbol(hit)(vfs)
   )
 
-  def oldSearchTypes(query: String, max: Int) =
-    index.searchClasses(query, max).filterNot {
-      name => name.fqn.endsWith("$") || name.fqn.endsWith("$class")
-    }.map(typeResult)
+  def oldSearchTypes(query: String, max: Int) = {
+    val scalaNames = scala.collection.mutable.Set.empty[String]
+
+    def isUnique(fqn: String): Boolean = {
+      val stripped =
+        if (fqn.endsWith("$")) fqn.dropRight(1)
+        else if (fqn.endsWith("$class")) fqn.dropRight(6)
+        else fqn
+
+      if (scalaNames.contains(stripped)) {
+        false
+      } else {
+        scalaNames.add(stripped)
+        true
+      }
+    }
+
+    index.searchClasses(query, max)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/ensime/core/Indexer.scala
```diff
@@ -24,10 +24,26 @@ class Indexer(
...
+
+      if (scalaNames.contains(stripped)) {
+        false
+      } else {
+        scalaNames.add(stripped)
+        true
+      }
+    }
+
+    index.searchClasses(query, max)
```
iravid: Suggestion to avoid the mutable Set you are dissatisfied with: 

```
index.searchClasses(query, max)
  .map { fqn =>
    // fqn normalization here
  }
  .toSet
  .map(typeResult)
```

fommil:
looks like a failure in the build? This could have re-ordered results

```
[info] - should open the simple test project *** FAILED ***
586s
1021
[info]   java.lang.AssertionError: assertion failed: expected:  but got unexpected message SymbolSearchResults(List(TypeSearchResult(java.util.Random,Random,Class,Some(LineSourcePosition(ArchiveFile(/usr/lib/jvm/zulu-8-amd64/src.zip,/java/util/Random.java),106))), TypeSearchResult(scala.util.Random,Random,Class,Some(LineSourcePosition(ArchiveFile(/root/.coursier/cache/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.12.2/scala-library-2.12.2-sources.jar,/scala/util/Random.scala),21)))))
```

core/src/main/scala/org/ensime/core/Indexer.scala
fommil: yeah, I'd prefer this... I don't think performance is too much of a concern here.  But instead of a Set (which reorders things) maybe just `.distinct` ?

core/src/main/scala/org/ensime/core/Indexer.scala
```diff
@@ -24,10 +24,26 @@ class Indexer(
...
-  def oldSearchTypes(query: String, max: Int) =
-    index.searchClasses(query, max).filterNot {
-      name => name.fqn.endsWith("$") || name.fqn.endsWith("$class")
-    }.map(typeResult)
+  def oldSearchTypes(query: String, max: Int) = {
+    val scalaNames = scala.collection.mutable.Set.empty[String]
+
+    def isUnique(fqn: String): Boolean = {
+      val stripped =
+        if (fqn.endsWith("$")) fqn.dropRight(1)
```
fommil: how about just using `.replaceAll` with a regex? Believe it or not, it's probably faster than this because of the scala/java-isms going on.

core/src/main/scala/org/ensime/core/Indexer.scala
jozi-k: I believe that is not possible, `typeResult` expects `FqnSymbol` not `String`.

fommil: right, I see. We'd need a `.distinctBy`... lucky we have one of those on `List` ;-)

```
import org.ensime.util.list._
```

and it'll appear by magic.

jozi-k: Guys, you think much faster than me, let's slow down for a moment :D 
I do not get what you want to do in first `map`. I get `FqnSymbol` and need to convert it to `ClassDef` with changed `ClassDef#fqn`. So basically I would remap trait to case class in this case. Would you really implement it like that?

fommil: :smile: something like this (just doing `$` in example, not `$class`, apologies for typos)?

```scala
    import org.ensime.util.list._

    index.searchClasses(query, max).map(_.replaceAll("[$]^", "")).distinctBy(_.fqn).map(typeResult)
```

fommil: oh, sorry the `.map(_.replaceAll)` needs to be on the fqn so it's a little messier.

jozi-k: Ufff... I thought I missed something obvious, but now I see the misunderstanding. `name` is not case class and as far as I know there is no macro which defines `copy` for traits. This leads us to my previous post where I complained about converting trait to case class. I could do that with reflection, but I do not want to (I already broke most of FP good practices :D) Can it be done via new graph stuff?

jozi-k:
The test if failing because it expects both `scala.util.Random` and `scala.util.Random$`. But my change will return only scala.util.Random in this case. On the other hand, when you have object but no companion class/trait it would return only `path.to.your.Object$`

core/src/main/scala/org/ensime/core/Indexer.scala
jozi-k: Will change it. Btw, would you mind sharing the source of this information? Will be very happy to broaden my horizons.

core/src/main/scala/org/ensime/core/Indexer.scala
fommil: Aah, I see. Hmm, I'll need to think about that.

core/src/main/scala/org/ensime/core/Indexer.scala
fommil: Just experience. replaceAll is java method where's drop etc are scala EnrichedString methods that treat String as Seq[Char]

core/src/main/scala/org/ensime/core/Indexer.scala
jozi-k: I ended up with pattern matching.

fommil:
Cool! You worked it out, but no need to find and then replace... these are regexps so they optimise that already. Also your second replaceAll you only put the dollar in the square brackets... this is a "character class" and will match any of the characters inside. There is really good docs on the Java Pattern class about how regexes work in Java. A language within the language!

fommil:
You've definitely earned a few ensime stickers, if you want to claim them privately email me your address (assuming you're somewhere I can post to)

core/src/main/scala/org/ensime/core/Indexer.scala
```diff
@@ -24,10 +24,21 @@ class Indexer(
...
-    index.searchClasses(query, max).filterNot {
-      name => name.fqn.endsWith("$") || name.fqn.endsWith("$class")
-    }.map(typeResult)
+  def oldSearchTypes(query: String, max: Int): List[TypeSearchResult] = {
+    // Remove $/$class from the end.
+    def strip(fqn: String): String = fqn.replaceAll("\\$(class)*$", "")
+
+    import org.ensime.util.list._
+
+    index.searchClasses(query, max)
```
jozi-k: You already closed the PR but I want to discuss the return type of `searchClasses`. I think it is much more understandable to have List[ClassDef] or even Vector[ClassDef]. I would be surprised to get Method/Field instances from `searchClasses`, so code below would be simplified.

jozi-k:
Just out of curiosity, when commenting on the issue you mentioned 
> Actually we really want to search by scala name, which is now possible (I think) in the graph.

Can you share more detail about it? I asked for it previously but it was probably overlooked in the review/code stuff.

core/src/main/scala/org/ensime/core/Indexer.scala
fommil: This method is implementing a legacy feature that has been in the wire protocol for years. It might be possible to change the type signature to `List[ClassDef]` without impact, but it could also impact the wire format of the object being serialised... it's the difference between the client expecting a `case class` or a `sealed trait` (which encodes which `case class` is used). You could try changing this signature and then using `: List[TheSuperType]` in the actor response and then finally removing the type assertion, and see if it breaks anything in the clients.

fommil:
gladly! Start by following the "even more recent one" conference talk in http://ensime.org/contributing/ which describes the Graphpocalypse. It has since been merged. Andrey probably won't be able to respond for a few weeks as he is sitting his final exams.

PR #1781, opened 2017-05-26
Replaced usages of EnsimeModule with EnsimeProject. Deleted EnsimeModule

VC1995:
Deleted `EnsimeModule`.
`modules` in `EnsimeConfig` now correspond to a `List[EnsimeProject]`.

I have assumed the following equivalents :
`compileJars` in EnsimeModule <--> `libraryJars` in EnsimeProject
`referenceSourceJars` in EnsimeModule <--> `librarySources` in EnsimeProject
`docJars` in EnsimeModule <--> `libraryDocs` in EnsimeProject
Others are pretty straightforward.

The hack for handing `TypeCheckModule(moduleName)` and `UnloadModule(moduleName)` in `Analyzer.scala` needs review. Using `EnsimeProject` breaks files that would have been included in one `EnsimeModule` into two EnsimeProjects with `compile` and `test` configurations(probably more).  Hence we now need to filter EnsimeProject with `EnsimeProject.id.project == moduleName`, instead of a simple get.

If there's anything wrong, please let me know and I'll try my best to correct it. :)

### DIFF
--- a/api/src/main/scala/org/ensime/api/config.scala
+++ b/api/src/main/scala/org/ensime/api/config.scala
@@ -18,7 +18,6 @@ final case class EnsimeConfig(
     scalaVersion: String,
     compilerArgs: List[String],
     javaSources: List[File],
-    subprojects: List[EnsimeModule],
     projects: List[EnsimeProject],
     javaLibs: List[File]
 ) {
@@ -29,58 +28,31 @@ final case class EnsimeConfig(
   /* Proposed alternatives to the legacy wire format field names */
   def root = rootDir
   val referenceSourceJars =
-    (javaSources ++ subprojects.flatMap(_.referenceSourceRoots)).toSet
+    (javaSources ++ projects.flatMap(_.librarySources)).toSet
 
   // some marshalling libs (e.g. spray-json) might not like extra vals
-  val modules = subprojects.map { module => (module.name, module) }.toMap
+  val modules = projects.map { module => (module.id, module) }.toMap
 
   def compileClasspath: Set[File] = modules.values.toSet.flatMap {
-    m: EnsimeModule => m.compileDeps ++ m.testDeps
+    m: EnsimeProject => m.libraryJars
   } ++ (if (propOrFalse("ensime.sourceMode")) List.empty else targetClasspath)
 
   def targetClasspath: Set[File] = modules.values.toSet.flatMap {
-    m: EnsimeModule => m.targets ++ m.testTargets
+    m: EnsimeProject => m.targets
   }
 
   def allJars: Set[File] = {
     modules.values.flatMap { m =>
-      m.compileDeps ::: m.testDeps
+      m.libraryJars
     }.toSet
   } ++ javaLibs
 
   val allTargets: Set[File] =
-    subprojects.flatMap(sm => sm.targets ::: sm.testTargets).toSet
+    projects.flatMap(_.targets).toSet
 
-  def allDocJars: Set[File] = modules.values.flatMap(_.docJars).toSet
+  def allDocJars: Set[File] = modules.values.flatMap(_.libraryDocs).toSet
 
   def scalaLibrary: Option[File] = allJars.find(_.getName.startsWith("scala-library"))
-}
-
-final case class EnsimeModule(
-    name: String,
-    targets: List[File],
-    testTargets: List[File],
-    dependsOnModules: List[String],
-    compileDeps: List[File],
-    testDeps: List[File],
-    sourceRoots: List[File],
-    docJars: List[File],
-    referenceSourceRoots: List[File]
-) {
-  // only check the files, not the directories, see below
-  (compileDeps ::: testDeps ::: referenceSourceRoots).foreach { f =>
-    require(f.exists, "" + f + " is required but does not exist")
-  }
-
-  /*
-   Proposed alternatives to the legacy wire format field names:
-   */
-  def compileJars = compileDeps
-  def testJars = testDeps
-  def referenceSourceJars = referenceSourceRoots
-
-  def dependencies(implicit config: EnsimeConfig): List[EnsimeModule] =
-    dependsOnModules.map(config.modules)
 }
 
 final case class EnsimeProjectId(
@@ -100,4 +72,7 @@ final case class EnsimeProject(
     libraryDocs: Set[File]
 ) {
   sources.foreach(f => require(f.exists, "" + f + " is required but does not exist"))
+
+  def dependencies(implicit config: EnsimeConfig): List[EnsimeProject] =
+    depends.toList.map(config.modules)
 }
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -53,7 +53,7 @@ final case class TypecheckFileReq(fileInfo: SourceFileInfo) extends RpcAnalyserR
 /**
  * Response with a `VoidResponse`.
  */
-final case class UnloadModuleReq(module: String) extends RpcAnalyserRequest
+final case class UnloadModuleReq(moduleId: EnsimeProjectId) extends RpcAnalyserRequest
 
 /**
  * Responds with a `VoidResponse`
@@ -63,7 +63,7 @@ final case class UnloadFileReq(fileInfo: SourceFileInfo) extends RpcAnalyserRequ
 /**
  * Response with a `VoidResponse`.
  */
-final case class TypecheckModule(module: String) extends RpcAnalyserRequest
+final case class TypecheckModule(moduleId: EnsimeProjectId) extends RpcAnalyserRequest
 
 /**
  * Responds with a `VoidResponse`.
--- a/core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
+++ b/core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
@@ -30,7 +30,7 @@ trait EnsimeConfigFixture {
 
   // convenience method
   def main(lang: String)(implicit config: EnsimeConfig): File =
-    config.subprojects.head.sourceRoots.filter { dir =>
+    config.projects.head.sources.filter { dir =>
       val sep = JFile.separator
       dir.getPath.endsWith(s"${sep}main$sep$lang")
     }.head
@@ -38,7 +38,7 @@ trait EnsimeConfigFixture {
   def javaMain(implicit config: EnsimeConfig): File = main("java")
 
   def mainTarget(implicit config: EnsimeConfig): File =
-    config.subprojects.head.targets.head
+    config.projects.head.targets.head
 }
 
 object EnsimeConfigFixture {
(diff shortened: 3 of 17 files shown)


api/src/main/scala/org/ensime/api/config.scala
```diff
@@ -29,60 +28,33 @@ final case class EnsimeConfig(
...
-    (javaSources ++ subprojects.flatMap(_.referenceSourceRoots)).toSet
+    (javaSources ++ projects.flatMap(_.librarySources)).toSet
 
   // some marshalling libs (e.g. spray-json) might not like extra vals
-  val modules = subprojects.map { module => (module.name, module) }.toMap
+  val modules = projects.map { module => (module.id, module) }.toMap
 
   def compileClasspath: Set[File] = modules.values.toSet.flatMap {
-    m: EnsimeModule => m.compileDeps ++ m.testDeps
+    m: EnsimeProject => m.libraryJars
```
yyadavalli: @VC1995 with the work you are doing in creating an Analyzer per module, I don't think we will need these any longer.
Same for `targetClasspath`, `allJars`, `allTargets` and `allDocJars`.

core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
```diff
@@ -52,6 +52,8 @@ object EnsimeConfigFixture {
     )
     System.err.flush()
     sys.exit(1)
+  } else {
+    println(dotEnsime.getAbsolutePath)
```
yyadavalli: Why is this needed here?

core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
```diff
@@ -141,18 +143,17 @@ object EnsimeConfigFixture {
       if (copyTargets) renameAndCopy(from)
       else rename(from)
 
-    def cloneModule(m: EnsimeModule): EnsimeModule = m.copy(
+    def cloneModule(m: EnsimeProject): EnsimeProject = m.copy(
```
yyadavalli: You might want to rename it to `cloneProject` for consistency

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -28,7 +28,7 @@
  * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
  * PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE REGENTS OR
  * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
- * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DsubpAMAGES (INCLUDING, BUT NOT
```
yyadavalli: Committed a typo

core/src/main/scala/org/ensime/indexer/SearchService.scala
```diff
@@ -137,7 +137,7 @@ class SearchService(
     def findBases(): (Set[FileObject], Map[FileName, Set[FileObject]]) = {
       val (jarFiles, dirs) = config.modules.flatMap {
         case (name, m) =>
```
yyadavalli: The match is on a `id`, you can rename it for consistency, even though its not relevant here. You can also ignore it with `_`

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -185,13 +185,14 @@ class Analyzer(
...
-      config.modules get (moduleName) foreach {
+      //hack: to typecheck both compile and test configurations
+      (config.modules filter { case (moduleId, module) => moduleId.project == moduleName }).values foreach {
         module =>
           val files: Set[SourceFileInfo] = module.scalaSourceFiles.map(s => SourceFileInfo(EnsimeFile(s), None, None))(breakOut)
           sender ! scalaCompiler.handleReloadFiles(files)
       }
     case UnloadModuleReq(moduleName) =>
-      config.modules get (moduleName) foreach {
+      (config.modules filter { case (moduleId, module) => moduleId.project == moduleName }).values foreach {
```
yyadavalli: You can use `filterKeys` to simplify the code a bit. `filterKeys(id => id.project == moduleName)`.

core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
VC1995: I was doing that for testing something. I'm so sorry, I'll correct it.

yyadavalli: Okay, also you could squash your commits (if you like) for a cleaner history.

core/src/main/scala/org/ensime/core/Analyzer.scala
VC1995: Yep, I'll use it instead.

api/src/main/scala/org/ensime/api/config.scala
VC1995: Will it be okay if we remove these as we bring in code for Analyzer per module? I am just a little hesistant to remove all of it now as they are being used in a few classes and i am not sure about refactoring them.

yyadavalli: Sure, just wanted to point that out as you do the cleanup.

core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
VC1995: I'll remove the comment on TypeCheckModule(moduleName). Did I leave out more comments?

VC1995:
BasicWorkflow test now fails as I get two void responses on Type checking a module as projects with `EnsimeProjectId("testing_simple","compile")` and `EnsimeProjectId("testing_simple","test")` are loaded.
This was expected. Should I add an extra expectMsg(VoidResponse) statement? or Refactor code to send only a single response?

yyadavalli:
I don't think adding an extra `expectMsg(VoidResponse)` will help. You should either aggregate the result for all modules, or change it to accept a `EnsimeProjectId`, you can assume the default configuration of `compile`, if multiple are found and one is not provided.

VC1995:
For TypeCheckModule
I aggregated the EnsimeServerErrors into one with a description formed by combining all the descriptions inserting newline between them. If no error is obtained I return a Void Response.
For UnloadModule
I simply return a Void Response after unloading all modules with module.id.project == moduleName.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -185,18 +185,29 @@ class Analyzer(
...
       }
     case UnloadModuleReq(moduleName) =>
-      config.modules get (moduleName) foreach {
+      (config.modules filterKeys (_.project == moduleName)).values foreach {
         module =>
           val files = module.scalaSourceFiles.toList
           files.foreach(scalaCompiler.askRemoveDeleted)
-          sender ! VoidResponse
       }
+      sender ! VoidResponse
```
yyadavalli: I think you need to enclose them in a block

VC1995: That'll send multiple Void Responses. We wanted to send just one, right?

VC1995: This is kind of a hack though. Currently if we received a request and we had no module named `moduleName`, we wouldn't send back a Void Response. Otherwise, it's same.
Will that be a problem? If so, we can check if we get some EnsimeProject with `module.id.project` equal to moduleName and send back a Void Response only then.
But does this really matter?

yyadavalli: I think you are fine

yyadavalli: But for `TypeCheckModule` we maybe better off, accepting an `id`, it would be much simpler to change the call site to send the configuration.

VC1995: Yep. I think so too. I'll make that change. :)

VC1995:
I changed `TypeCheckModule(moduleId)` and `UnloadModuleReq(moduleId)`. They now accept an EnsimeProjectId as parameter.

yyadavalli:
@VC1995 Looks good. We need to watch out for any failures with clients using `TypecheckModule`

VC1995:
Where do I look for making the required changes?
We need to pass an EnsimeProjectId instead of a string.
Also, having an implicit function to convert a `String` to `EnsimeProjectId` with config as `"compile"` in the companion object of `EnsimeProjectId` could work for now, right?

yyadavalli:
I don't think you need to make any changes on the server side. 

The clients such as emacs need to make sure they are passing correct info such that the requests are unmarshalled correctly. So if we ever get issues regarding failure of typecheck module we need to update the client as needed.

We can add an implicit conversion if we are worried about backward compatibility, but we are not. If you are using Developer edition, you can expect to see stuff breaking.

.gitignore
```diff
@@ -8,3 +8,4 @@ ensime.sbt
 #
 # If you want to add something from your personal setup,
 # add it to your ~/.gitignore instead.
+.idea
```
fommil: please follow the instructions.

api/src/main/scala/org/ensime/api/config.scala
```diff
@@ -29,60 +28,33 @@ final case class EnsimeConfig(
...
   def allJars: Set[File] = {
     modules.values.flatMap { m =>
-      m.compileDeps ::: m.testDeps
+      m.libraryJars
     }.toSet
   } ++ javaLibs
 
   val allTargets: Set[File] =
-    subprojects.flatMap(sm => sm.targets ::: sm.testTargets).toSet
+    projects.flatMap(m => m.targets).toSet
```
fommil: you can write `_.targets`

api/src/main/scala/org/ensime/api/incoming.scala
```diff
@@ -53,7 +53,7 @@ final case class TypecheckFileReq(fileInfo: SourceFileInfo) extends RpcAnalyserR
 /**
  * Response with a `VoidResponse`.
  */
-final case class UnloadModuleReq(module: String) extends RpcAnalyserRequest
+final case class UnloadModuleReq(moduleId: EnsimeProjectId) extends RpcAnalyserRequest
```
fommil: these changes to the API will show up in the wire protocol. You'll need to add a custom marshaller for `EnsimeProjectId` in jerk(y) and swank(y), e.g. see how we add a custom `DebugThreadIdFormat` (you'll need to add that custom formatter in 3 places, and it wouldn't be a bad idea to add tests)

fommil:
almost there, just a few things. Unfortunately the changes to the wire protocol mean we can't merge just yet.

api/src/main/scala/org/ensime/api/incoming.scala
fommil: actually, I've changed my mind... let's just change the protocol message. @ktonga isn't this used in vim? This message will change to require a richer message.

In that case, bump the `1.9.1` to `1.9.2` in `outgoing.scala`

fommil: or was it @hedefalk who wanted it in atom...

VC1995: so should I simply bump the version to 1.9.2 ?

.gitignore
VC1995: I'm sorry I missed the instructions. I have corrected it.

api/src/main/scala/org/ensime/api/incoming.scala
VC1995: If we added a custom marshaller we would have the configuration default to `"compile"`.
There could be cases when user might want to typecheck files in the `"test"` configuration as well and might fail later. (say) during resolution related to `ScalaTest` which wasn't among the `libraryJars` in the `"compile"` configuration. Currently since `compileDeps` and `testDeps` are in the same `EnsimeModule` identifiable through `moduleName`it wasn't be a problem. Right?

ktonga: @fommil no, we're not using that req in vim ATM

fommil:
cool, could you please just push small commits to a PR with the changes? We can squash when we merge the PR but it makes it easier to see the changes in response to the review.

VC1995:
Yes sure. Sorry about that. I'll take care from now on.

fommil:
first merge, w00t!

PR #1786, opened 2017-06-04
Analyser per Module : Stage 1

VC1995:
The main introduction is the `AnalyzerManager.scala` class. Except the `AnalyzerManager` class itself it has `ModuleFinder` trait and `LoadedFilesData` class.
I extracted `ModuleFinder` as a trait as it might be used elsewhere. (though I am not sure that is the best idea).
The history of an analyzer necessary to initialize the project-wide analyzer when required is stored with the help of `LoadedFilesData`.

I added a test to check if requests that would go to different analyzers were handled correctly. Hence introduced `AnalyzerManagerSpec` and `AnalyzerManagerFixture`. `BasicWorkflow` checked that the basic working remained intact.

Currently the project-wide analyser that handles the refactor request isn't closed and I need to come up with a way to do that. Any suggestions would really help. 
Also I used the `collector` method for collecting results of various requests(`TypeByName`, `InspectPackageByPath`) and initially made it generic. Now it's only used for `TypeCheckFilesReq`, so should I change it?

Please allow me to make whatever changes you think are needed. I'll try my best to solve any issues and improve the code.

### DIFF
--- a/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
@@ -7,10 +7,13 @@ import java.nio.charset.Charset
 import java.text.SimpleDateFormat
 import java.util.Date
 
-import org.ensime.api._
+import org.ensime.api.{ AddImportRefactorDesc, ExpandMatchCasesDesc, OrganiseImportsRefactorDesc, RawFile, RefactorDiffEffect, RefactorReq, RenameRefactorDesc, RpcResponse, SourceFileInfo }
+
+//import org.ensime.api._
 import org.ensime.fixture._
 import org.ensime.util.EnsimeSpec
 import org.scalatest.Assertions
+
 import scala.concurrent.ExecutionContext.Implicits.global
 
 class RefactoringHandlerSpec extends EnsimeSpec
@@ -22,7 +25,6 @@ class RefactoringHandlerSpec extends EnsimeSpec
   def original = EnsimeConfigFixture.EmptyTestProject.copy(
     compilerArgs = List("-encoding", encoding)
   )
-
   // transitionary methods
   def ContentsSourceFileInfo(file: File, contents: String) =
     SourceFileInfo(RawFile(file.toPath), Some(contents))
--- a/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
@@ -7,13 +7,10 @@ import java.nio.charset.Charset
 import java.text.SimpleDateFormat
 import java.util.Date
 
-import org.ensime.api.{ AddImportRefactorDesc, ExpandMatchCasesDesc, OrganiseImportsRefactorDesc, RawFile, RefactorDiffEffect, RefactorReq, RenameRefactorDesc, RpcResponse, SourceFileInfo }
-
-//import org.ensime.api._
+import org.ensime.api._
 import org.ensime.fixture._
 import org.ensime.util.EnsimeSpec
 import org.scalatest.Assertions
-
 import scala.concurrent.ExecutionContext.Implicits.global
 
 class RefactoringHandlerSpec extends EnsimeSpec
@@ -25,6 +22,7 @@ class RefactoringHandlerSpec extends EnsimeSpec
   def original = EnsimeConfigFixture.EmptyTestProject.copy(
     compilerArgs = List("-encoding", encoding)
   )
+
   // transitionary methods
   def ContentsSourceFileInfo(file: File, contents: String) =
     SourceFileInfo(RawFile(file.toPath), Some(contents))
--- a/core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
+++ b/core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
@@ -17,7 +17,17 @@ object AnalyzerFixture {
   private[fixture] def create(search: SearchService)(implicit system: ActorSystem, config: EnsimeConfig, vfs: EnsimeVFS): TestActorRef[Analyzer] = {
     val indexer = TestProbe()
     val projectActor = TestProbe()
-    TestActorRef(Analyzer(projectActor.ref, indexer.ref, search))
+    TestActorRef(Analyzer(projectActor.ref, indexer.ref, search, EnsimeProject(
+      EnsimeProjectId("testProject", "compile"),
+      Seq(),
+      config.projects.flatMap(_.sources).toSet,
+      config.projects.flatMap(_.targets).toSet,
+      config.compilerArgs,
+      config.projects.flatMap(_.javacOptions).distinct,
+      config.projects.flatMap(_.libraryJars).toSet,
+      config.projects.flatMap(_.librarySources).toSet,
+      config.projects.flatMap(_.libraryDocs).toSet
+    )))
   }
 }
 
--- a/core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
+++ b/core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
@@ -14,20 +14,10 @@ trait AnalyzerFixture {
 }
 
 object AnalyzerFixture {
-  private[fixture] def create(search: SearchService)(implicit system: ActorSystem, config: EnsimeConfig, vfs: EnsimeVFS): TestActorRef[Analyzer] = {
+  private[fixture] def create(search: SearchService, id: EnsimeProjectId)(implicit system: ActorSystem, config: EnsimeConfig, vfs: EnsimeVFS): TestActorRef[Analyzer] = {
     val indexer = TestProbe()
     val projectActor = TestProbe()
-    TestActorRef(Analyzer(projectActor.ref, indexer.ref, search, EnsimeProject(
-      EnsimeProjectId("testProject", "compile"),
-      Seq(),
-      config.projects.flatMap(_.sources).toSet,
-      config.projects.flatMap(_.targets).toSet,
-      config.compilerArgs,
-      config.projects.flatMap(_.javacOptions).distinct,
-      config.projects.flatMap(_.libraryJars).toSet,
-      config.projects.flatMap(_.librarySources).toSet,
-      config.projects.flatMap(_.libraryDocs).toSet
-    )))
+    TestActorRef(Analyzer(projectActor.ref, indexer.ref, search, config.modules(id)))
   }
 }
 
--- a/core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
+++ b/core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
@@ -114,10 +114,10 @@ object EnsimeConfigFixture {
 
     def rename(from: File): File = {
       val toPath = from.getAbsolutePath.replace(
-        source.root.getAbsolutePath,
+        source.rootDir.getAbsolutePath,
         target.getAbsolutePath
       )
-      require(toPath != from.getAbsolutePath, s"${source.root.getAbsolutePath} ${target.getAbsolutePath} in ${from.getAbsolutePath}")
+      require(toPath != from.getAbsolutePath, s"${source.rootDir.getAbsolutePath} ${target.getAbsolutePath} in ${from.getAbsolutePath}")
       File(toPath)
     }
 
@@ -160,7 +160,7 @@ object EnsimeConfigFixture {
       file.writeLines(file.readLines())
     }
 
-    if (preWarm && (config.compileClasspath ++ config.javaLibs).nonEmpty)
+    if (preWarm && (config.classpath ++ config.javaLibs).nonEmpty)
       EnsimeCacheProject.foreach { cacheProject =>
         log.info(s"copying ${cacheProject.cacheDir}")
         cacheProject.cacheDir.toPath.copyDirTo(config.cacheDir.toPath)
--- a/core/src/it/scala/org/ensime/intg/JarTargetTest.scala
+++ b/core/src/it/scala/org/ensime/intg/JarTargetTest.scala
@@ -88,7 +88,7 @@ class MissingJarTargetTest extends EnsimeSpec
           mainTarget should be a 'file
 
           // means the file addition was detected
-          asyncHelper.expectMsg(10 seconds, CompilerRestartedEvent)
+          asyncHelper.expectMsg(10 seconds, AnalyzerReadyEvent)
 
           eventually(timeout(scaled(10 seconds)), interval(scaled(1 second))) {
             project ! PublicSymbolSearchReq(List("Foo"), 5)
--- a/core/src/main/scala/org/ensime/core/Analyzer.scala
+++ b/core/src/main/scala/org/ensime/core/Analyzer.scala
@@ -160,9 +160,9 @@ class Analyzer(
   }
 
   def ready: Receive = withLabel("ready") {
-    case ReloadExistingFilesEvent if allFilesMode =>
+    case AskReTypecheck if allFilesMode =>
       log.info("Skipping reload, in all-files mode")
-    case ReloadExistingFilesEvent =>
+    case AskReTypecheck =>
       restartCompiler(keepLoaded = true)
 
     case FullTypeCheckCompleteEvent =>
@@ -207,13 +207,6 @@ class Analyzer(
         module =>
           val files: Set[SourceFileInfo] = module.scalaSourceFiles.map(s => SourceFileInfo(EnsimeFile(s), None, None))(breakOut)
           sender ! scalaCompiler.handleReloadFiles(files)
-      }
-    case UnloadModuleReq(moduleId) =>
-      config.modules get (moduleId) foreach {
-        module =>
-          val files = module.scalaSourceFiles.toList
-          files.foreach(scalaCompiler.askRemoveDeleted)
-          sender ! VoidResponse
       }
     case TypecheckFileReq(fileInfo) =>
       sender ! scalaCompiler.handleReloadFiles(Set(fileInfo))
--- a/core/src/main/scala/org/ensime/core/Analyzer.scala
+++ b/core/src/main/scala/org/ensime/core/Analyzer.scala
@@ -185,6 +185,16 @@ class Analyzer(
     case RemoveFileReq(file: File) =>
       scalaCompiler.askRemoveDeleted(file)
       sender ! VoidResponse
+    case UnloadAllReq =>
+      if (propOrFalse("ensime.sourceMode")) {
+        log.info("in source mode, will reload all files")
+        scalaCompiler.askRemoveAllDeleted()
+        restartCompiler(keepLoaded = true)
+      } else {
+        allFilesMode = false
+        restartCompiler(keepLoaded = false)
+      }
+      sender ! VoidResponse
     case TypecheckModule(moduleId) =>
       //consider the case of a project with no modules
       config.modules get (moduleId) foreach {
--- a/core/src/main/scala/org/ensime/core/Analyzer.scala
+++ b/core/src/main/scala/org/ensime/core/Analyzer.scala
@@ -54,6 +54,7 @@ class Analyzer(
     broadcaster: ActorRef,
     indexer: ActorRef,
     search: SearchService,
+    module: EnsimeProject,
     implicit val config: EnsimeConfig,
     implicit val vfs: EnsimeVFS
 ) extends Actor with Stash with ActorLogging with RefactoringHandler {
@@ -80,8 +81,8 @@ class Analyzer(
       case Some(scalaLib) => settings.bootclasspath.value = scalaLib.getAbsolutePath
       case None => log.warning("scala-library.jar not present, enabling Odersky mode")
     }
-    settings.classpath.value = config.compileClasspath.mkString(JFile.pathSeparator)
-    settings.processArguments(config.compilerArgs, processAll = false)
+    settings.classpath.value = module.compileClasspath.mkString(JFile.pathSeparator)
+    settings.processArguments(module.scalacOptions, processAll = false)
     presCompLog.debug("Presentation Compiler settings:\n" + settings)
 
     reporter = new PresentationReporter(new ReportHandler {
@@ -98,11 +99,14 @@ class Analyzer(
     reporter.disable() // until we start up
 
     scalaCompiler = makeScalaCompiler()
-
     broadcaster ! SendBackgroundMessageEvent("Initializing Analyzer. Please wait...")
 
     scalaCompiler.askNotifyWhenReady()
-    if (propOrFalse("ensime.sourceMode")) scalaCompiler.askReloadAllFiles()
+
+    // each analyzer must load files  of its module
+    if (propOrFalse("ensime.sourceMode"))
+      scalaCompiler.askReloadAllFiles(module)
+
   }
 
   protected def makeScalaCompiler() = new RichPresentationCompiler(
@@ -134,8 +138,8 @@ class Analyzer(
       reporter.enable()
       // legacy clients expect to see AnalyzerReady and a
       // FullTypeCheckCompleteEvent on connection.
-      broadcaster ! Broadcaster.Persist(AnalyzerReadyEvent)
-      broadcaster ! Broadcaster.Persist(FullTypeCheckCompleteEvent)
+      //      broadcaster ! Broadcaster.Persist(AnalyzerReadyEvent)
+      //      broadcaster ! Broadcaster.Persist(FullTypeCheckCompleteEvent)
       context.become(ready)
       unstashAll()
 
@@ -151,6 +155,7 @@ class Analyzer(
 
     case FullTypeCheckCompleteEvent =>
       broadcaster ! FullTypeCheckCompleteEvent
+      context.parent ! FullTypeCheckCompleteEvent
 
     case req: RpcAnalyserRequest =>
       // fommil: I'm not entirely sure about the logic of
@@ -164,6 +169,9 @@ class Analyzer(
   }
 
   def allTheThings: PartialFunction[RpcAnalyserRequest, Unit] = {
+    case RemoveFilesReq(files) =>
+      files.foreach(scalaCompiler.askRemoveDeleted)
+      sender ! VoidResponse
     case RemoveFileReq(file: File) =>
       scalaCompiler.askRemoveDeleted(file)
       sender ! VoidResponse
@@ -209,7 +217,6 @@ class Analyzer(
         reporter.disable()
         scalaCompiler.askCompletionsAt(pos(fileInfo, point), maxResults, caseSens)
       } pipeTo sender
-
     case UsesOfSymbolAtPointReq(file, point) =>
       import context.dispatcher
       val response = if (toSourceFileInfo(file).exists()) {
@@ -328,10 +335,11 @@ object Analyzer {
   def apply(
     broadcaster: ActorRef,
     indexer: ActorRef,
-    search: SearchService
+    search: SearchService,
+    module: EnsimeProject
   )(
     implicit
     config: EnsimeConfig,
     vfs: EnsimeVFS
-  ) = Props(new Analyzer(broadcaster, indexer, search, config, vfs))
+  ) = Props(new Analyzer(broadcaster, indexer, search, module, config, vfs))
 }
--- a/core/src/main/scala/org/ensime/core/Analyzer.scala
+++ b/core/src/main/scala/org/ensime/core/Analyzer.scala
@@ -81,17 +81,17 @@ class Analyzer(
       case None => log.warning("scala-library.jar not present, enabling Odersky mode")
     }
 
+    val projects = scoped.map(config.modules)
+
     settings.classpath.value = {
       for {
-        scope <- scoped
-        project = config.modules(scope)
+        project <- projects
         entry <- project.classpath
       } yield entry
     }.distinct.mkString(JFile.pathSeparator)
 
     // arbitrarily pick the first project when there are multiple
-    // projects, because we don't know what else to do
-    settings.processArguments(config.projects.head.scalacOptions, processAll = false)
+    settings.processArguments(projects.head.scalacOptions, processAll = false)
 
     presCompLog.debug("Presentation Compiler settings:\n" + settings)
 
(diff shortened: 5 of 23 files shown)


VC1995:
Looks like I'm failing at `should index jar targets that appear after startup` because currently I'm forwarding the `ReloadExistingFilesEvent` to all the active analyzers. Since none are active we get no `CompilerRestartedEvent` as expected by the test.

fommil:
even a void response would be ok here.

VC1995:
`askReTypeCheck` gets called due to the file being added which leads to the `Project` actor sending a `ReloadExistingFilesEvent` to the `AnalyzerManager`. If I send a `VoidResponse` back to the `Project` actor, how will I detect this in the `JarTargetTest`? 
Did you mean I should broadcast a `VoidResponse` ?

core/src/it/scala/org/ensime/intg/JarTargetTest.scala
```diff
@@ -88,7 +88,7 @@ class MissingJarTargetTest extends EnsimeSpec
           mainTarget should be a 'file
 
           // means the file addition was detected
-          asyncHelper.expectMsg(10 seconds, CompilerRestartedEvent)
+          asyncHelper.expectMsg(10 seconds, AnalyzerReadyEvent)
```
fommil: :+1:

core/src/it/scala/org/ensime/core/AnalyzerManagerSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+
+import org.ensime.api
+import org.ensime.api._
+import org.ensime.fixture.{ EnsimeConfigFixture, IsolatedAnalyzerManagerFixture }
+import org.ensime.model.BasicTypeInfo
+import org.ensime.util.EnsimeSpec
+import org.ensime.util.ensimefile.EnsimeFile
+import org.ensime.util.file._
+class AnalyzerManagerSpec extends EnsimeSpec
+    with IsolatedAnalyzerManagerFixture {
```
fommil: I'm not sure you need to go to the effort of creating a fixture for this... you should be able to just write mock actors for the underlying actors. That might mean you could make this a unit test instead of an integration test (because you won't need to run the test on a real scala project)

core/src/it/scala/org/ensime/core/AnalyzerManagerSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+          }
+
+          analyzerManager ! InspectPackageByPathReq("org.example")
+          val packageInfo = expectMsgType[PackageInfo]
+          packageInfo.name shouldBe "example"
+          packageInfo.fullName shouldBe "org.example"
+
+          packageInfo.members.collect {
+            case b: api.BasicTypeInfo => b.copy(pos = None)
+          } should contain theSameElementsAs List(
```
fommil: I'm not sure the exact contents matter... it's enough to assert that the list is non-empty. The exact details can change between scala versions so over-specifying tests is often be a burden.

core/src/it/scala/org/ensime/core/AnalyzerManagerSpec.scala
VC1995: Yes, I'll do that.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
```diff
@@ -19,9 +19,7 @@ class RefactoringHandlerSpec extends EnsimeSpec
     with RefactoringHandlerTestUtils {
 
   val encoding = "UTF-16"
-  def original = EnsimeConfigFixture.EmptyTestProject.copy(
-    compilerArgs = List("-encoding", encoding)
```
fommil: I'm not sure we should remove this

core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
```diff
@@ -17,7 +17,17 @@ object AnalyzerFixture {
   private[fixture] def create(search: SearchService)(implicit system: ActorSystem, config: EnsimeConfig, vfs: EnsimeVFS): TestActorRef[Analyzer] = {
     val indexer = TestProbe()
     val projectActor = TestProbe()
-    TestActorRef(Analyzer(projectActor.ref, indexer.ref, search))
+    TestActorRef(Analyzer(projectActor.ref, indexer.ref, search, EnsimeProject(
+      EnsimeProjectId("testProject", "compile"),
+      Seq(),
+      config.projects.flatMap(_.sources).toSet,
+      config.projects.flatMap(_.targets).toSet,
+      List("-encoding", "UTF-16", "-feature", "-deprecation", "-language:postfixOps", "-language:implicitConversions", "-Xlint:-unused,_", "-Yno-adapted-args", "-Ywarn-dead-code", "-Xfuture", "-Ywarn-unused:implicits,imports,-locals,-params,-patvars,-privates", "-Xlint", "-Ywarn-numeric-widen", "-Ywarn-unused-import", "-Ymacro-expand:discard"),
```
fommil: I don't think we want all these settings by default for all analyzer tests.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
fommil: or, at least, we want to change it in the `projects` field

core/src/it/scala/org/ensime/core/AnalyzerManagerSpec.scala
VC1995: `org.example2` is in two modules. I wanted to make sure I was getting all the members. 
Like `org.ensime.util` in ensime-server.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -151,6 +155,7 @@ class Analyzer(
 
     case FullTypeCheckCompleteEvent =>
       broadcaster ! FullTypeCheckCompleteEvent
+      context.parent ! FullTypeCheckCompleteEvent
```
fommil: this doesn't seem right, shouldn't the broadcaster be taking care of this?

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,385 @@
...
+
+class LoadedFilesData(module: EnsimeProject) {
+  /**
+   * typeCheckedAll  flag is true when a TypecheckModule or TypecheckAllReq was more recent than an UnloadAllReq or UnloadModuleReq
+   * unloadedAll     flag is true when a UnloadAllReq or UnloadModuleReq was more recent than an TypecheckModule or TypecheckAllReq
+   * If we haven't seen any, both flag are false
+   * files           means different when different flags are set. Either contains the list od files to be loaded or the files to be unloaded
+   * tempFiles       used when both flags are false; gives list of files that were loaded
+   */
+  private var typeCheckedAll = false
```
fommil: I'm not keen on non-actor code storing state like this. Can't you refactor this is as an immutable `case class` that lives in an actor?

core/src/main/scala/org/ensime/util/Reporter.scala
```diff
@@ -33,6 +33,7 @@ class PresentationReporter(handler: ReportHandler) extends Reporter with Positio
   }
 
   override def info0(pos: Position, msg: String, severity: Severity, force: Boolean): Unit = {
+    //    println(s"called: with message $msg")
```
fommil: stray comment

core/src/main/scala/org/ensime/util/FileUtil.scala
```diff
@@ -14,6 +16,15 @@ object FileUtils {
   implicit def toSourceFileInfo(f: Either[File, SourceFileInfo]): SourceFileInfo =
     f.fold(l => SourceFileInfo(RawFile(l.toPath), None, None), r => r)
 
+  implicit def convertToPath(f: File): Path = f.toPath
```
fommil: I don't think we want implicit conversions for these.

core/src/main/scala/org/ensime/util/FileUtil.scala
```diff
@@ -14,6 +16,15 @@ object FileUtils {
...
+
+  implicit def convertToPath(f: SourceFileInfo): Path = {
+    val file: RichEnsimeFile = f.file
+    file.toPath
+  }
+
+  implicit def convertToPath(f: Either[File, SourceFileInfo]): Path = convertToPath(toSourceFileInfo(f))
```
fommil: matching on `Left` vs `Right` would seem more efficient

project/EnsimeBuild.scala
```diff
@@ -286,6 +287,10 @@ object EnsimeTestingBuild {
     libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.1" % Test intransitive ()
   )
 
+  lazy val testingSimple2 = testingProject("testing/simple2") settings (
```
fommil: is there any reason you need to create a new test project instead of re-using the existing one?

core/src/main/scala/org/ensime/core/Analyzer.scala
VC1995: It does. But I needed to use this in `AnalyzerManager` to handle the requests which didn't have a module to them.

core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
VC1995: Actually I did this because we'll be using `scalacOptions` instead of `compiler-args`.
I either had to change a few things in `RefactoringHandlerSpec` or do this here.

util/src/main/scala/org/ensime/util/ensimefile.scala
```diff
@@ -77,6 +78,7 @@ package object ensimefile {
     override def readStringDirect()(implicit cs: Charset): String = withEntry(_.readString())
     override def uri: URI = URI.create(s"jar:${archive.jar.toUri}!${archive.entry}") // path is null (opaque)
     override def canon: ArchiveFile = ArchiveFile(archive.jar.canon, archive.entry)
+    override def toPath: Path = archive.jar
```
fommil: this won't work... it's supposed to represent a file within an archive, not the archive itself.

project/EnsimeBuild.scala
VC1995: I used this for a sample multi-module project.

util/src/main/scala/org/ensime/util/ensimefile.scala
```diff
@@ -26,7 +26,7 @@ package ensimefile {
 
     /** Direct access contents: not efficient for streaming. */
     def readStringDirect()(implicit cs: Charset): String
-
+    def toPath: Path
```
fommil: I'm not keen on this change. It is very difficult to deal with `File` vs "file in an archive" and the point of this sealed family was to force us to deal with both cases separately when it comes up. You may have to rethink all the callers of `toPath` to see what you really want out of the API. Do you perhaps want a string identifier? In which case `uriString` is perhaps what you want.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
VC1995: Shouldn't we move from using `compiler-args` to `scalacOptions` ?

core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
```diff
@@ -29,13 +29,18 @@ trait EnsimeConfigFixture {
   def withEnsimeConfig(testCode: EnsimeConfig => Any): Any
 
   // convenience method
-  def main(lang: String)(implicit config: EnsimeConfig): File =
-    config.projects.head.sources.filter { dir =>
+
+  def mains(lang: String)(implicit config: EnsimeConfig): List[File] =
```
fommil: the convenience part here was in returning a single dir from a project definition where we only expected one `src/main/scala` directory. Has that changed?

core/src/main/scala/org/ensime/core/Analyzer.scala
fommil: I'm not sure I understand, we can come back to this one.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
fommil: yes, but we're not doing anything now

core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
fommil: we don't want to hand code the flags here, they should be read from the `.ensime`. We might need to add some (the minimal!) flags to the testproject settings in project/EnsimeBuild.scala if omitting these causes a problem.

project/EnsimeBuild.scala
fommil: ok, but you can do that already by filtering the `.ensime` to include two named projects.

core/src/it/scala/org/ensime/fixture/EnsimeConfigFixture.scala
VC1995: No, it hasn't. The `scalaMain` and `javaMain` functions still get one src/main/scala directory. I extracted out this `mains` function as it could be used to get multiple src/main/scala directories for a multi-module testing project.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -34,97 +34,47 @@ trait ModuleFinder {
...
-  def getModule(fileInfo: SourceFileInfo): Option[EnsimeProject] = getModule(richEnsimeFile(fileInfo.file))
-  def getModule(file: File): Option[EnsimeProject] = getModule(richEnsimeFile(RawFile(file.toPath)))
+  def getModule(fileInfo: SourceFileInfo): Option[EnsimeProject] = fileInfo.file match {
+    case RawFile(f) => getModule(f)
+    case ArchiveFile(f, _) => getModule(f)
+  }
+  def getModule(file: File): Option[EnsimeProject] = getModule(file.toPath)
   def getModule(file: Either[File, SourceFileInfo]): Option[EnsimeProject] = file match {
     case Left(f) => getModule(f)
     case Right(sourceFileInfo) => getModule(sourceFileInfo)
```
fommil: you can pattern match within the `SourceFileInfo` btw.

```scala
case Right(RawFile(f))      =>
case Right(ArchiveFile(f)) =>
```

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -155,7 +105,7 @@ class AnalyzerManager(
     broadcaster ! Broadcaster.Persist(AnalyzerReadyEvent)
     broadcaster ! Broadcaster.Persist(FullTypeCheckCompleteEvent)
     // initialize loadedFiles
-    config.projects foreach (p => historyOfModule += (p -> new LoadedFilesData(p)))
+    config.projects foreach (p => historyOfModule += (p -> LoadedFilesData(Map(), NotLoaded)))
```
fommil: `Map.empty` is slightly better performance than `Map()` because the latter will create a zero element sequence. Only usually relevant for code in a tight loop, but worth you knowing about.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -165,7 +115,7 @@ class AnalyzerManager(
       for {
         (module, analyzer) <- analyzers
       } analyzer forward ReloadExistingFilesEvent
-      if(analyzers.isEmpty)
+      if (analyzers.isEmpty)
```
fommil: I'd have done it as a `if (analyzers.isEmpty) ... else for ...` to avoid reasoning about the implied no-op in the `for` comprehension.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -183,7 +133,23 @@ class AnalyzerManager(
 
     }
 
-  private def withExistingModuleFor(file: Path, req: RpcAnalyserRequest)(f: (RpcAnalyserRequest, EnsimeProject) => Unit): Unit = {
+  def toFile(f: SourceFileInfo) = f.file match {
+    case RawFile(path) => path.toFile
+    case ArchiveFile(path, _) => path.toFile
```
fommil: I'm not sure if `.toFile` is correct for `ArchiveFile`. We may be missing a test case here. How come you want to get the `File` out?

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -275,7 +266,7 @@ class AnalyzerManager(
         }
         def receive: Receive = {
           case FullTypeCheckCompleteEvent =>
-            analyzer ! RemoveFilesReq(filesToBeDeleted.toList.map(f => richEnsimeFile(f.file).toPath.toFile))
+            analyzer ! RemoveFilesReq(filesToBeDeleted.map(toFile))
```
fommil: this definitely isn't correct if the `File` is a `jar`. Perhaps you want to `collect` on `RawFile` ? I don't really understand why we're sending `RemoveFilesReq`, it's a bit weird.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: Well because `RemoveFileReq` accepts a parameter of type `File` and here I have a list of `SourceFileInfo` . I'm not sure what I should do?

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: Because `TypeCheckFilesReq` would enter symbols into the `symsByFile` map and then I can delete those with `RemoveFilesReq`. 
I am actually wanting to delete the symbols that were deleted before we get a request for the project-wide actor.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: :+1:

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
fommil: hmm, can you please explain the steps for me?

VC1995: First I get the files whose symbols I must remove using `removeSymbolsOf` on all modules, and I pass a `TypeCheckFilesReq(_)` to the project-wide analyzer to type-check these files.
Then, when I get a `FullTypeCheckCompleteEvent`, I remove the symbols unsing `RemoveFilesReq`.
Now I forward the request (the one not specific to any module that I received) to this analyzer.

core/src/main/scala/org/ensime/core/FqnToSymbol.scala
```diff
@@ -50,9 +50,13 @@ trait FqnToSymbol { self: Global with SymbolToFqn =>
       segToSym(nme.segments(scalaName, assumeTerm = assumeTerm), rootSymbol)
     }
   }.getOrElse {
-    val term = segToSym(nme.segments(scalaName, assumeTerm = true), rootSymbol)
+    val nameSegments = nme.segments(scalaName, assumeTerm = true)
+    val term = segToSym(nameSegments, rootSymbol)
+    //    println(s"term is $term")
```
jozi-k: forgotten println

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,381 @@
...
+class AnalyzerManager(
+    broadcaster: ActorRef,
+    indexer: ActorRef,
+    search: SearchService,
+    implicit val config: EnsimeConfig,
+    implicit val vfs: EnsimeVFS
+) extends Actor with ActorLogging with ModuleFinder with Stash {
+
+  // maps the active modules to their analyzers
+  private var analyzers: mutable.Map[EnsimeProject, ActorRef] = mutable.Map()
```
fommil: or a `var` of an immutable `Map` (which I find better because then it's explicit). But doing both is very unusual.

project/EnsimeBuild.scala
VC1995: I was initially did it so that I could have members of `org.example2` in two modules `testing_simple` and `testing_simple2`. I could have used existing but I was hesitant to change the existing testing projects. 
But I am not using it now so I'll revert it. :)

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -98,25 +82,24 @@ class AnalyzerManager(
...
   override def receive: Receive = ready
 
   private def ready: Receive = withLabel("ready") {
     case ReloadExistingFilesEvent =>
-      for {
-        (module, analyzer) <- analyzers
-      } analyzer forward ReloadExistingFilesEvent
       if (analyzers.isEmpty)
-        broadcaster ! AnalyzerReadyEvent
+        context.parent ! AnalyzerReadyEvent
```
fommil: try to avoid using `context.parent` when sending messages. Try to use the sender or the `broadcaster` where possible. Baking in the `context.parent` implies that the actor hierarchy is rigid and usually indicates a design flaw.

VC1995: Okay. Should I pass the `broadcaster` to the `AnalyzerManager` then?

fommil: definitely

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,362 @@
...
+import scala.collection.{ breakOut, mutable }
+
+trait ModuleFinder {
+  implicit val config: EnsimeConfig
+
+  def getModule(path: Path): Option[EnsimeProject] = {
+    config.projects.collectFirst {
+      case p if (p.sources.exists(f => path.toString.startsWith(f.toString))) =>
+        p
+    }
```
jozi-k: I know I proposed the change to `collectFirst`, but now I see you rearranged code a bit and it is possible to use `find`.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,362 @@
...
+  }
+
+  def getModule(fileInfo: SourceFileInfo): Option[EnsimeProject] = fileInfo match {
+    case SourceFileInfo(RawFile(f), _, _) => getModule(f)
+    case SourceFileInfo(ArchiveFile(f, _), _, _) => getModule(f)
+  }
```
jozi-k: could be deleted as you can call `getModule(fileInfo.file)` or even `getModule(fileInfo.file.toPath)`

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,362 @@
...
+
+class AnalyzerManager(
+    broadcaster: ActorRef,
+    analyzerCreator: EnsimeProject => Props,
+    implicit val config: EnsimeConfig
+) extends Actor with ActorLogging with ModuleFinder with Stash {
+
+  // maps the active modules to their analyzers
+  private var analyzers: Map[EnsimeProject, ActorRef] = Map.empty
+  private val historyOfModule: mutable.Map[EnsimeProject, LoadedFilesData] = mutable.Map.empty
```
jozi-k: This seems bit confusing for me. I would not mix usage of immutable/var and mutable/val, but rather pick one and stick with it.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,362 @@
...
+      Seq(),
+      config.projects.flatMap(_.sources).toSet,
+      config.projects.flatMap(_.targets).toSet,
+      config.projects.flatMap(_.scalacOptions).distinct,
+      config.projects.flatMap(_.javacOptions).distinct,
+      config.projects.flatMap(_.libraryJars).toSet,
+      config.projects.flatMap(_.librarySources).toSet,
+      config.projects.flatMap(_.libraryDocs).toSet
+    )
+  )
```
jozi-k: What do you think about creating factory for `EnsimeProject` from `config`?

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,362 @@
...
+      else
+        for {
+          (module, analyzer) <- analyzers
+        } analyzer forward ReloadExistingFilesEvent
+    case req: RpcAnalyserRequest =>
+      forwardToAnalyzer(req)
+  }
+
+  // helper method
+  private def getAnalyzerFor(module: EnsimeProject): ActorRef =
```
jozi-k: Suggesting to rename this method to `getOrSpawnNew`

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,362 @@
...
+        val newAnalyzer = context.actorOf(analyzerCreator(module))
+        analyzers = analyzers + (module -> newAnalyzer)
+        newAnalyzer
+
+    }
+
+  def toFile(f: SourceFileInfo) = f.file match {
+    case RawFile(path) => path.toFile
+    case ArchiveFile(path, _) => path.toFile
+  }
```
jozi-k: Does it make sense to put `path` to `trait EnsimeFile`?

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,362 @@
...
+      var missingSource: Boolean = files.exists(getModule(_).isEmpty)
+      if (missingSource)
+        sender ! EnsimeServerError("Update .ensime file.")
+      else {
+        val filesPerModule: Map[EnsimeProject, List[Either[File, SourceFileInfo]]] = files.groupBy(getModule).map(x => x._1.get -> x._2)
+        for ((module, list) <- filesPerModule) {
+          val analyzer = getAnalyzerFor(module)
+          analyzer ! TypecheckFilesReq(list)
+          //        if (!fs.map(toSourceFileInfo).exists(!_.file.exists()))
+          //          fs.map(toSourceFileInfo).foreach(f => filesStatus += f -> filesStatus.getOrElse(f, default).append(Loaded))
```
jozi-k: forgotten comments

fommil:
thanks @jozi-k ! I will have time to look at this PR later on today and may push commits onto here.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: By implementing this in the apply method of the companion?

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: I'm always seem to do that :/ . Thanks for telling, I'll remove them.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: Yes, it does.
We could use the already existing `fullPath` method in `RichArchiveFile` and add it in `RichRawFile`

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: Is a mutable map more performant in the case of updates to the map?

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: I actually used this in `getModule(file: Either[File, SourceFileInfo])`

VC1995: to call `getModule(fileInfo.file)` I will have to write `getModule(file: EnsimeFile)` and the latter isn't possible because `trait EnsimeFile` doesn't have toPath

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: I am handling the case for `ArchiveFile` incorrectly here. Using `fullPath` method on `RichArchiveFile` to initialize a `File` for the case of an `ArchiveFile` might be the correct way to go.

VC1995: I though of adding `path` to `trait EnsimeFile` but I am not sure how to implement that for `ArchiveFile` though I think we'll have to use `fullPath`

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -80,8 +81,8 @@ class Analyzer(
       case Some(scalaLib) => settings.bootclasspath.value = scalaLib.getAbsolutePath
       case None => log.warning("scala-library.jar not present, enabling Odersky mode")
     }
-    settings.classpath.value = config.compileClasspath.mkString(JFile.pathSeparator)
-    settings.processArguments(config.compilerArgs, processAll = false)
+    settings.classpath.value = module.compileClasspath.mkString(JFile.pathSeparator)
```
fommil: this doesn't include all the transitive dependencies of the module. I think we need some coverage around this if it's currently passing. e.g. what if module A depends on B, and A is using a dependency that is pulled in by B?

fommil: I'll update this

core/src/it/scala/org/ensime/fixture/AnalyzerFixture.scala
```diff
@@ -17,7 +17,17 @@ object AnalyzerFixture {
   private[fixture] def create(search: SearchService)(implicit system: ActorSystem, config: EnsimeConfig, vfs: EnsimeVFS): TestActorRef[Analyzer] = {
     val indexer = TestProbe()
     val projectActor = TestProbe()
-    TestActorRef(Analyzer(projectActor.ref, indexer.ref, search))
+    TestActorRef(Analyzer(projectActor.ref, indexer.ref, search, EnsimeProject(
```
fommil: this doesn't look right. Was this a workaround to the transitive classpath problem?

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,339 @@
...
+          if (!list.map(toSourceFileInfo).exists(!_.file.exists)) {
+            val previousData = historyOfModule(module)
+            historyOfModule += module -> update(previousData) { (statusOfFile, default) =>
+              val addToStatusOfFile = list.map(toSourceFileInfo).map(f => f -> statusOfFile.getOrElse(f, default).append(Loaded))
+              LoadedFilesData(statusOfFile ++ addToStatusOfFile, default)
+            }
+          }
+
+        }
+        context.become(collector[List[String]](filesPerModule.size, Nil, sender)((newResponse, aggregate) => {
```
fommil: a concern I have about all these `become` calls is that it means ensime becomes effectively single threaded... the analyzer manager holds back all calls until this one responds, instead of being able to process many at a time. We'll need to think of another way of doing the results collecting.

fommil: probably we want to spawn an anonymous actor to listen for aggregated responses.

VC1995: Yes, I'll do that. I was doing it initially.

fommil:
let's open up a fresh PR rebased against #1792, from an appropriately named branch.

Issue #1790, opened 2017-06-11
pre-analyzer per module possibilities


fommil:
before #1786 we should consider if we want to do the following (with client breaking consequences)

- [x] cleanups around config / classpath handling
- [x] remove loadall
- [x] deprecate unloadall
- [x] remove sourceMode
- [x] add a "restart PC" request with an `Option[EnsimeProjectId]` (ignored initially) and a parameter allowing three kinds of restart: unloading everything, keep loaded files, load everything. This will actually bounce the `scalaCompiler` (more reliable than loading / unloading)
- [x] consolidate RemoveFileReq, UnloadFileReq into UnloadFilesReq (retained server side for a while)
- [x] `SourceFileInfo` take an `Option[EnsimeProjectId]`
- [x] drop `AstAtPointReq` (better implemented by the build tool, e.g. `ensimeCompileOnly`, or a pcplod test)

fommil:
@rorygraves @dragos @yazgoo @ches @ktonga @ChrisCoffey @jvican @hedefalk @d6y @dickwall @VC1995 any client side concerns? Perhaps @VC1995 could address them shortly after the server changes get merged.

fommil:
also, is anything other than emacs using the `by-name` API calls? I'd like to remove them #1787

It would also be good to get your thoughts on #1789

PR #1792, opened 2017-06-11
Stage 0: Analyzer per Project

fommil:
most of #1790


fommil:
@VC1995 following this, the Stage 1 should rethink how reloading of scala compilers is handled... perhaps we should add support to an `Analyzer` to go "dormant" for a while (i.e. persist its list of open files and allow the `scalaCompiler` to be garbage collected)

fommil:
what the hell? On Windows we're getting a compiler warning compling from `Predef` 

```
NewScalaNotesEvent(false,List(Note(C:\Users\appveyor\.coursier\cache\v1\https\repo1.maven.org\maven2\org\scala-lang\scala-library\2.12.2\scala-library-2.12.2-sources.jar!/scala/Predef.scala,Procedure syntax is deprecated. Convert procedure `assume` to method by adding `: Unit =`.,NoteWarn,11861,11861,233,35)))
```

in `BasicWorkflow` instead of the final `FullTypeCheckCompleteEvent`.

fommil:
@sugakandrey do you think it might be possible that reverse lookup is loading Predef on Windows but not on any other platform?

sugakandrey:
@fommil no idea honestly, I don't have a windows machine, so I've never really gotten to testing windows shenanigans properly

Issue #1787, opened 2017-06-10
Graphpocalypse browser


fommil:
The current implementation of the RPC calls used by the emacs inspector require use of the analyzer. We should rethink / redesign / rewrite the Emacs inspector to make use of the graphdb only, with appropriate and simpler API calls.

Using the global (all-seeing) Analyzer poses significant problems for #1152 which brings huge benefits for macros and compiler plugin usage.

fommil:
@timcharper said he uses the Inspector... which may be a reason not to remove this. Would you be interested in implementing an all-new set of endpoints that could be easily implemented in Emacs (and other clients)?

My problem with the Emacs Inspector is that it's just so complicated and no other client has bothered to implement it.

On the server side, it is using the PC rather than the graph. I should imagine that anything in this area would be the super-set of https://github.com/ensime/ensime-server/issues/1754 and #1753 and #1742

fommil:
@timcharper my recommended alternative to your usecase is to use `C-c C-v d` to get the documentation for the type at point, which is usually a far superior way to browse the methods of a class (`sbt myproject/doc` if it's in one of your own to generate the docs). Can you please try that out this week because it would be a *huge* win if we could remove these unreliable old queries... and then longer term move to Graphpocalypse-based features.

fommil:
sorry I meant @mpollmeier ^^^

mpollmeier:
I know `ensime-show-doc-for-symbol-at-point` and occasionally use it (however I'd prefer not to leave emacs for this). My main use case for the inspector is actually much simper: I just want to know the compiler-inferred type at point. If it doesn't list all available functions and inheritance, that's fine. 

Is there something simpler that achieves that? If not, what's a good place to start reading/investigating what's necessary to rebuild the inspector from the graph?

fommil:
LOL you've been using it for that? You'll kick yourself... just use `C-c C-v t` use `T` for full path and prefix with the universal argument `C-u` to add to the kill ring.

fommil:
To build a better inspector, start by watching my Scala Sphere thank (contributing page) about the Graphpocalypse, it introduces all the apis and let's you know what's there.

mpollmeier:
Hehe, well I'm not really kicking myself now, the inspector was pretty fast for me. 
And you shouldn't have pointed me to `ensime-type-at-point`, now I have little to no incentive to rebuild the inspector :) Maybe it's actually best to be dropped.

Issue #1152, opened 2015-11-15
Analyzer per module


fommil:
I'm working on a project that has different `scalacOptions` per module, and some modules have compiler plugins and others don't. To fix this (and get limited scope contextual completions) it would be good if each module had its own analyzer.

However, some problems with this:
1. potentially huge drain on heap usage
2. how do we handle cross-module refactorings? Maybe each refactor needs its own throwaway presentation compiler which loads all necessary source files

fommil:
scala-ide say they use a watchdog that closes presentation compilers if they have not been used for 20 seconds and it works well. That could work.

balajisivaraman:
As discussed on Gitter with @fommil, I'll be happy to try and fix this issue once I'm comfortable enough with the Ensime Server codebase.

fommil:
an interesting impact of fixing this would be that macros should be a lot better in ensime! Because macros don't work in the same presentation compiler that loads them.

rorygraves:
Unless one of the PCs is doing full compilation to bytecode how is the second PC going to apply the macro?  Am I missing something?

fommil:
Sbt does the compiling, second one picks it up

fommil:
another thing that we could do as part of this refactor (perhaps we'll create a new ticket) would be to monitor PC actors with heartbeats and basically throw them away / kill them if they disappear for more than 30 seconds. We would need to keep a count of how many PCs are still "being killed" so that we don't blow up the heap. This could be accompanied with an async message to users to let them know that they maybe need to introduce some `cachedImplicit`s or something.

PR #1795, opened 2017-06-11
An Inspector Doesn't Call

fommil:
Contentious PR because this removes the support for the Emacs' Inspector feature, which would be better served with a rewrite as per #1787 

The reasons to remove these calls are numerous:

- it relies on using an "all seeing" presentation compiler, which doesn't have a well defined set of compiler parameters
- it is pretty buggy
- it is very very slow on larger projects
- if you navigate to the wrong place, it can just take down the server (I've done this many times, especially the tab complete) and quite often cause Emacs to crash. On Windows, the fix is a reboot, nothing else will do.
- it's existence dramatically complicates #1152 and could mean that it is all our sponsored dev gets to do. That feature is highly sought after by many users and will allow ensime to "just work" on projects that use macros and compiler plugins. Right now, it's the biggest turn-off for people who've tried ensime and got lots of red squigglies, so left.

I am veering towards finishing off this PR and merging, but I am open to being convinced by fans of the Inspector, e.g. @mpollmeier @timcharper (although I really think it's less effort to just implement the same functionality with the graph and / or documentation browser)

### DIFF
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -155,17 +155,6 @@ final case class DocUriAtPointReq(
 ) extends RpcAnalyserRequest
 
 /**
- * Responds with a `StringResponse` for the URL of the documentation if valid,
- * or `FalseResponse`.
- */
-@deprecating("https://github.com/ensime/ensime-server/issues/1787")
-final case class DocUriForSymbolReq(
-  typeFullName: String,
-  memberName: Option[String],
-  signatureString: Option[String]
-) extends RpcAnalyserRequest
-
-/**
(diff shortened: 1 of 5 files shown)


fommil:
@VC1995 FYI... feel free to finish this off

fommil:
https://twitter.com/fommil/status/874023033513684997

prassee:
does this block auto complete feature ...

fommil:
@prasse ?

mpollmeier:
as discussed on https://github.com/ensime/ensime-server/issues/1787 I don't mind seeing Inspector Columbo leave the room.

fommil:
moved to #1796

fommil:
(trying hard not to make a "one more question" gag)...

PR #1796, opened 2017-06-12
An Inspector Doesn't Call: Deleting the server side support for the Emacs Inspector

VC1995:
picked up from #An Inspector Doesn't Call #1795

### DIFF
--- a/core/src/main/scala/org/ensime/core/Completion.scala
+++ b/core/src/main/scala/org/ensime/core/Completion.scala
@@ -342,27 +342,6 @@ object Keywords {
   }
 }
 
-trait Completion { self: RichPresentationCompiler =>
-
-  def completePackageMember(path: String, prefix: String): List[CompletionInfo] = {
-    toSymbol(PackageName(path.split('.').toList)) match {
-      case NoSymbol => List.empty
-      case sym =>
-        val memberSyms = packageMembers(sym).filterNot { s =>
-          s == NoSymbol || s.nameString.contains("$")
-        }
-        memberSyms.flatMap { s =>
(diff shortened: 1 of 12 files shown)


fommil:
you can also delete all the methods in the RichPresentationCompiler that are no longer used.

VC1995:
Okay :+1:

core/src/main/scala/org/ensime/core/Completion.scala
```diff
@@ -39,19 +39,18 @@
...
-
 import akka.actor.ActorRef
 import akka.pattern.Patterns
 import akka.util.Timeout
 import org.ensime.api._
-import org.ensime.indexer.PackageName
 import org.ensime.indexer.lucene.SimpleLucene
 import org.ensime.util.Timing.dilation
 
+import scala.collection.mutable
```
fommil: import weirdness

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -38,13 +38,22 @@
...
+import org.ensime.indexer._
+import org.ensime.model._
+import org.ensime.util.ensimefile._
+import org.ensime.util.file._
+import org.ensime.util.sourcefile._
+import org.ensime.vfs._
+import org.slf4j.LoggerFactory
+
 import scala.collection.immutable.{ Set => SCISet }
+import scala.collection.mutable
```
fommil: more import weirdness

VC1995: fixing real quick

fommil:
one more question

VC1995:
haha,yes?

Issue #1793, opened 2017-06-11
AppVeyor CI Broken: Windows is not unloading files


fommil:
in https://github.com/ensime/ensime-server/pull/1792 we had a weird error that we just worked around.

To reproduce, remove the hack in `ProjectFixture.scala` that ignores `Predef.scala` scala notes. Then run `BasicWorkflow` on a windows box. If you don't have access to a Windows box, edit appveyor.yml to only run this one test.

Some debugging to output the loaded file list after each stage of the test would be useful, or to print out which messages are causing which files to be loaded. e.g. add `logger.debug` info to `askReloadFiles`

fommil:
still seeing problems, e.g. now at the end of BasicWorkflow we are getting spurious NewScalaNotesEvent from a file that's been unloaded. There is something funky going on.

fommil:
@VC1995 if you could take a look at this one that would be good, as a break when the python and rebasing gets too much for you :smile:

VC1995:
@fommil Yes, I was looking already. I got the that `NewScalaNotesEvent` on my linux machine as well at somepoint so this caught my eye. We also get these due the `SymbolAtPontReq` in the `BasicWorkflow` test but those are filtered out using `asyncHelper.fishForMessage`. I am going to try fixing this today :)

fommil:
Removing fishForMessage would be a good first step.

PR #1797, opened 2017-06-12
Fix incorrect assertions in BasicWorkflow

VC1995:
#1793 
Removed `asyncHelper.fishForMessage` and tried to determine `NewScalaNotesEvent` at different points.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -26,7 +26,7 @@ pipeline:
         fi
       - if [ -n "$DRONE_PULL_REQUEST" ] ; then cd testing/cache && sbt ++$SCALA_VERSION ensimeConfig ensimeServerIndex && cd ../.. ; fi
       - if [ -n "$DRONE_PULL_REQUEST" ] ; then sbt ++$SCALA_VERSION "testOnly -- -l tags.IgnoreOnDrone" ; fi
-      - if [ -n "$DRONE_PULL_REQUEST" ] ; then sbt ++$SCALA_VERSION "it:testOnly -- -l tags.IgnoreOnDrone"; fi
+      - if [ -n "$DRONE_PULL_REQUEST" ] ; then sbt ++$SCALA_VERSION "it:testOnly -- -l tags.IgnoreOnDrone" ; fi
       - if [ -z "$DRONE_PULL_REQUEST" ] ; then sbt ++$SCALA_VERSION publish ; fi
       - if [ -z "$DRONE_PULL_REQUEST" ] ; then
           echo "$SSH_KEY" > ~/.ssh/id_rsa ; chmod 600 ~/.ssh/id_rsa ;
--- a/.drone.yml
+++ b/.drone.yml
@@ -25,8 +25,7 @@ pipeline:
           exit 1 ;
         fi
       - if [ -n "$DRONE_PULL_REQUEST" ] ; then cd testing/cache && sbt ++$SCALA_VERSION ensimeConfig ensimeServerIndex && cd ../.. ; fi
-      - if [ -n "$DRONE_PULL_REQUEST" ] ; then sbt ++$SCALA_VERSION "testOnly -- -l tags.IgnoreOnDrone" ; fi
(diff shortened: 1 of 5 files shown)


.drone.yml
```diff
@@ -25,7 +25,7 @@ pipeline:
           exit 1 ;
         fi
       - if [ -n "$DRONE_PULL_REQUEST" ] ; then cd testing/cache && sbt ++$SCALA_VERSION ensimeConfig ensimeServerIndex && cd ../.. ; fi
-      - if [ -n "$DRONE_PULL_REQUEST" ] && [ "$SCALA_VERSION" = "2.12.2" ] ; then sbt ++$SCALA_VERSION "core/it:testOnly *BasicWorkflow*" ; fi
+      - if [ -n "$DRONE_PULL_REQUEST" ] ; then sbt ++$SCALA_VERSION "core/it:testOnly *BasicWorkflow*" ; fi
```
fommil: to control the version of scala, remove entries from the matrix at the bottom of the file.

VC1995: okay, but I was only testing a theory. No longer needed. :)

VC1995:
So it happened because we get a `FullTypeCheckCompleteEvent` when we restart the compiler with non-empty files and the analyzer is in "ready" state (here https://github.com/ensime/ensime-server/blob/2.0/core/src/main/scala/org/ensime/core/Analyzer.scala#L134 and here https://github.com/ensime/ensime-server/blob/2.0/core/src/main/scala/org/ensime/core/Analyzer.scala#L162)
we were only receiving `CompilerRestartedEvent` in `BasicWorkflow` and the unaccounted message kept creeping down unless we saw some unpredicted message later on.

fommil:
@VC1995 you've still got the temporary comments in the CI definition files.

core/src/it/scala/org/ensime/fixture/ProjectFixture.scala
```diff
@@ -3,12 +3,11 @@
 package org.ensime.fixture
 
 import akka.actor.{ ActorRef, ActorSystem }
-import org.scalatest._
 import akka.testkit._
-
 import org.ensime.api._
```
fommil: could you please back out the import rearrangements?

Probably you need to change your IntelliJ sorting options so they go

```
java.

scala.

<other>
```

I'm not sure how you do that

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -2,18 +2,18 @@
 // License: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.intg
 
-import scala.concurrent.duration._
```
fommil: same here

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -37,24 +37,25 @@ class BasicWorkflow extends EnsimeSpec
 
           project ! TypecheckModule(EnsimeProjectId("testing_simple", "compile"))
           expectMsg(VoidResponse)
-          all(asyncHelper.receiveN(2)) should matchPattern {
+          all(asyncHelper.receiveN(3)) should matchPattern {
```
fommil: heh, nice catch. I actually tried something like this lower down, but it was after the fishing so it was missed. We should probably rationalise our use of `fish` everywhere.

core/src/it/scala/org/ensime/fixture/ProjectFixture.scala
VC1995: I'm sorry that I keep messing it up. I'll fix this. :)

fommil: heh, don't worry... everybody does.

fommil:
you forgot to compile it in `sbt` before pushing so the formatting rules didn't kick in :smile:

Issue #1791, opened 2017-06-11
config should use RawFile, not File


fommil:
Pretty simple refactor, will touch lots of code.

It would be good to take the opportunity to push for more nio / `Path` usage instead of legacy IO as per #1437

fommil:
and while at it, removing `javaLibs` from the `EnsimeConfig` would be good.

zaynetro:
I'd like to work on this one. First, I'll submit a pull request with `javaLibs` removed and then take a look at using `RawFile`.

zaynetro:
When I start using `RawFile` from `api/common.scala` this part breaks immediately:

```
[error] ensime-server/core/src/main/scala/org/ensime/config/EnsimeConfigProtocol.scala:26: Cannot find SexpFormat for org.ensime.api.EnsimeConfig
[error]   private implicit val configFormat: SexpFormat[EnsimeConfig] = cachedImplicit
```

I suspect that `cachedImplicit` that is coming from shapeless can find implicit value for java's `File` class but doesn't know about `RawFile` class. Shall I implement `SexpFormat` for `RawFile`?

Implicit values are always a pain to me. Please, provide some guidance about this issue.

fommil:
yes that's correct. In fact the problem is that there is no `Path` like there is for `File` in `StandardFormats.scala`, but that would break binary compatibility.

What you could do is something like what we do in `SwankyFormats.scala` but just for `RawFile`

```scala
  object Protocol extends DefaultSexpProtocol
    with OptionAltFormat
    with CamelCaseToDashes {
        implicit object EnsimeFileFormat extends SexpFormat[RawFile] {
    def write(f: RawFile): Sexp = SexpString(path.toString)
    def read(sexp: Sexp): RawFile = sexp match {
      case SexpString(file) => RawFile(file)
      case got => deserializationError(got)
    }
  }
  }
```

You may need to create a `Path` out of the `file`

zaynetro:
Thanks for the answer. I will have time on the weekend to work on this issue again.

PR #1799, opened 2017-06-13
Remove javaLibs from EnsimeConfig

zaynetro:
Fixes part of #1791

### DIFF
--- a/api/src/main/scala/org/ensime/api/config.scala
+++ b/api/src/main/scala/org/ensime/api/config.scala
@@ -12,8 +12,7 @@ final case class EnsimeConfig(
   scalaVersion: String,
   @deprecating("each project will have a compiler") compilerArgs: List[String],
   javaSources: List[File],
-  projects: List[EnsimeProject],
-  @deprecating("leaked SearchService impl detail") javaLibs: List[File]
+  projects: List[EnsimeProject]
 )
 
 final case class EnsimeProjectId(
--- a/api/src/main/scala/org/ensime/api/config.scala
+++ b/api/src/main/scala/org/ensime/api/config.scala
@@ -3,26 +3,17 @@
 package org.ensime.api
 
 import java.io.File
-import java.nio.file.Files
-
-import scala.collection.JavaConverters._
 
 final case class EnsimeConfig(
-    @deprecating("rootDir is no longer used except in testing") rootDir: File,
-    cacheDir: File,
-    javaHome: File,
-    name: String,
-    scalaVersion: String,
-    @deprecating("each project will have a compiler") compilerArgs: List[String],
-    javaSources: List[File],
-    projects: List[EnsimeProject]
-) {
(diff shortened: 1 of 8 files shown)


fommil:
thanks for looking into this... see I told you all it was easy :smile:

But don't forget to address all warnings, they are errors in CI.

api/src/main/scala/org/ensime/api/config.scala
```diff
@@ -3,17 +3,26 @@
...
+    @deprecating("rootDir is no longer used except in testing") rootDir: File,
+    cacheDir: File,
+    javaHome: File,
+    name: String,
+    scalaVersion: String,
+    @deprecating("each project will have a compiler") compilerArgs: List[String],
+    javaSources: List[File],
+    projects: List[EnsimeProject]
+) {
+  lazy val javaRunTime: List[File] = tree(javaHome).filter(_.getName == "rt.jar").toList
```
fommil: we're trying to remove methods from case classes to keep them as pure data definitions. Also, the `rt.jar` is not the only jar we care about.

core/src/main/scala/org/ensime/config/EnsimeConfigProtocol.scala
```diff
@@ -29,22 +27,14 @@ object EnsimeConfigProtocol {
...
+    validated(raw)
   }
 
-  // there are lots of JRE libs, but most people only care about
-  // rt.jar --- this could be parameterised.
-  private def inferJavaLibs(javaHome: File): List[File] =
-    // WORKAROUND https://github.com/ensime/ensime-server/issues/886
-    // speeds up the emacs integration tests significantly,
-    if (Properties.envOrNone("ENSIME_SKIP_JRE_INDEX").isDefined) Nil
-    else javaHome.tree.filter(_.getName == "rt.jar").toList
```
fommil: I stand corrected, the `rt.jar` *is* the only jar we care about... but the logic should live here, not in the `case class`

zaynetro: Do we want to keep supporting `ENSIME_SKIP_JRE_INDEX` env variable?

api/src/main/scala/org/ensime/api/config.scala
zaynetro: Instead of traversing files recursively do we know for a fact where `rt.jar` is located?

`$JAVA_HOME/jre/lib/rt.jar` is this location only for my machine?

core/src/main/scala/org/ensime/config/EnsimeConfigProtocol.scala
fommil: Nope

api/src/main/scala/org/ensime/api/config.scala
zaynetro: @fommil `tree` method is deprecated and is suggesting using paths instead. Could you elaborate?

Sorry, for too many questions

fommil:
kiitos!

PR #1800, opened 2017-06-14
[Sponsored] Analyzer per Project, better macro/plugin support

VC1995:
Used `Debouncer` in `AnalyzerManager` to generate `SuspendAnalyzer` messages every 5 minutes which is then passed to all analyzers. 
`Analyzer`s on receiving `SuspendAnalyzer` check if they have been inactive for too long and shutdown the scala compiler if so.
p.s. WIP

### DIFF
--- a/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
@@ -22,9 +22,10 @@ class RefactoringHandlerSpec extends EnsimeSpec
   val encoding = "UTF-16"
   def original = {
     val empty = EnsimeConfigFixture.EmptyTestProject
+    val origScalacOptions = empty.projects.head.scalacOptions
     empty.copy(
       projects = List(empty.projects.head.copy(
-        scalacOptions = List("-encoding", encoding)
+        scalacOptions = origScalacOptions ::: List("-encoding", encoding)
       ))
     )
   }
--- a/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
@@ -20,12 +20,14 @@ class RefactoringHandlerSpec extends EnsimeSpec
     with RefactoringHandlerTestUtils {
 
   val encoding = "UTF-16"
-  val ensimeProject = EnsimeConfigFixture.EmptyTestProject.projects.head.copy(
-    scalacOptions = List("-encoding", encoding)
-  )
-  def original = EnsimeConfigFixture.EmptyTestProject.copy(
-    projects = ensimeProject :: Nil
-  )
+  def original = {
+    val empty = EnsimeConfigFixture.EmptyTestProject
+    empty.copy(
+      projects = List(empty.projects.head.copy(
+        scalacOptions = List("-encoding", encoding)
+      ))
+    )
+  }
 
   // transitionary methods
   def ContentsSourceFileInfo(file: File, contents: String) =
--- a/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
@@ -7,11 +7,12 @@ import java.nio.charset.Charset
 import java.text.SimpleDateFormat
 import java.util.Date
 
+import scala.concurrent.ExecutionContext.Implicits.global
+
 import org.ensime.api._
 import org.ensime.fixture._
 import org.ensime.util.EnsimeSpec
 import org.scalatest.Assertions
-import scala.concurrent.ExecutionContext.Implicits.global
 
 class RefactoringHandlerSpec extends EnsimeSpec
     with IsolatedAnalyzerFixture
@@ -19,8 +20,11 @@ class RefactoringHandlerSpec extends EnsimeSpec
     with RefactoringHandlerTestUtils {
 
   val encoding = "UTF-16"
+  val ensimeProject = EnsimeConfigFixture.EmptyTestProject.projects.head.copy(
+    scalacOptions = List("-encoding", encoding)
+  )
   def original = EnsimeConfigFixture.EmptyTestProject.copy(
-    compilerArgs = List("-encoding", encoding)
+    projects = ensimeProject :: Nil
   )
 
   // transitionary methods
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -45,6 +45,10 @@ class BasicWorkflow extends EnsimeSpec
 
           project ! UnloadAllReq
           expectMsg(VoidResponse)
+          all(asyncHelper.receiveN(2)) should matchPattern {
+            case CompilerRestartedEvent =>
+            case FullTypeCheckCompleteEvent =>
+          }
 
           all(asyncHelper.receiveN(2)) should matchPattern {
             case CompilerRestartedEvent =>
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -209,6 +209,7 @@ class BasicWorkflow extends EnsimeSpec
                 )
               ) =>
           }
+
           // expand selection around "seven" in `foo.testMethod` call
           project ! ExpandSelectionReq(fooFile, 215, 215)
           val expandRange1 = expectMsgType[FileRange]
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -45,11 +45,12 @@ class BasicWorkflow extends EnsimeSpec
 
           project ! UnloadAllReq
           expectMsg(VoidResponse)
-          
+
           all(asyncHelper.receiveN(2)) should matchPattern {
             case CompilerRestartedEvent =>
             case FullTypeCheckCompleteEvent =>
           }
+
 
           // trigger typeCheck
           project ! TypecheckFilesReq(List(Left(fooFile), Left(barFile)))
@@ -209,7 +210,6 @@ class BasicWorkflow extends EnsimeSpec
                 )
               ) =>
           }
-          
           // expand selection around "seven" in `foo.testMethod` call
           project ! ExpandSelectionReq(fooFile, 215, 215)
           val expandRange1 = expectMsgType[FileRange]
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -2,17 +2,17 @@
 // License: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.intg
 
-import scala.concurrent.duration._
-
 import org.ensime.api
 import org.ensime.api.{ BasicTypeInfo => _, EnsimeFile => _, _ }
 import org.ensime.core._
 import org.ensime.fixture._
 import org.ensime.model.BasicTypeInfo
 import org.ensime.util.EnsimeSpec
-import org.ensime.util.ensimefile._
 import org.ensime.util.ensimefile.Implicits.DefaultCharset
+import org.ensime.util.ensimefile._
 import org.ensime.util.file._
+
+import scala.concurrent.duration._
 
 class BasicWorkflow extends EnsimeSpec
     with IsolatedEnsimeConfigFixture
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -2,17 +2,17 @@
 // License: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.intg
 
+import scala.concurrent.duration._
+
 import org.ensime.api
 import org.ensime.api.{ BasicTypeInfo => _, EnsimeFile => _, _ }
 import org.ensime.core._
 import org.ensime.fixture._
 import org.ensime.model.BasicTypeInfo
 import org.ensime.util.EnsimeSpec
-import org.ensime.util.ensimefile.Implicits.DefaultCharset
 import org.ensime.util.ensimefile._
+import org.ensime.util.ensimefile.Implicits.DefaultCharset
 import org.ensime.util.file._
-
-import scala.concurrent.duration._
 
 class BasicWorkflow extends EnsimeSpec
     with IsolatedEnsimeConfigFixture
@@ -45,17 +45,10 @@ class BasicWorkflow extends EnsimeSpec
 
           project ! UnloadAllReq
           expectMsg(VoidResponse)
-
           all(asyncHelper.receiveN(2)) should matchPattern {
             case CompilerRestartedEvent =>
             case FullTypeCheckCompleteEvent =>
           }
-
-          all(asyncHelper.receiveN(2)) should matchPattern {
-            case CompilerRestartedEvent =>
-            case FullTypeCheckCompleteEvent =>
-          }
-
 
           // trigger typeCheck
           project ! TypecheckFilesReq(List(Left(fooFile), Left(barFile)))
@@ -127,10 +120,11 @@ class BasicWorkflow extends EnsimeSpec
             ERangePosition(`packageFilePath`, 94, 80, 104)
           )
 
-          all(asyncHelper.receiveN(2)) should matchPattern {
-            case note: NewScalaNotesEvent =>
-            case FullTypeCheckCompleteEvent =>
+          asyncHelper.fishForMessage() {
+            case FullTypeCheckCompleteEvent => true
+            case _ => false
           }
+
           // note that the line numbers appear to have been stripped from the
           // scala library classfiles, so offset/line comes out as zero unless
           // loaded by the pres compiler
@@ -308,8 +302,7 @@ class BasicWorkflow extends EnsimeSpec
 
           project ! TypecheckFilesReq(List(Left(bazFile), Right(toBeUnloaded)))
           expectMsg(VoidResponse)
-
-          all(asyncHelper.receiveN(3)) should matchPattern {
+          all(asyncHelper.receiveN(2)) should matchPattern {
             case note: NewScalaNotesEvent =>
             case FullTypeCheckCompleteEvent =>
           }
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -45,6 +45,7 @@ class BasicWorkflow extends EnsimeSpec
 
           project ! UnloadAllReq
           expectMsg(VoidResponse)
+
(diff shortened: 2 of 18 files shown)


VC1995:
oh I need to remove the FIXMEs

fommil:
and rebase

VC1995:
argh! yes rebase

VC1995:
@fommil What exactly did you want to do here : https://github.com/VC1995/ensime-server/blob/961e3e1b37124d2142d5720b5610152dc663893b/core/src/main/scala/org/ensime/core/Analyzer.scala#L132

fommil:
`scoped: Option[EnsimeProjectId]` isn't needed becuase the actor already has a `scoped: List[EnsimeProjectId]` which should *always* be used (not optional).

VC1995:
Oh, okay. I'll redo that. I did something very stupid, haha.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
```diff
@@ -7,20 +7,24 @@ import java.nio.charset.Charset
...
 class RefactoringHandlerSpec extends EnsimeSpec
     with IsolatedAnalyzerFixture
     with RichPresentationCompilerTestUtils
     with RefactoringHandlerTestUtils {
 
   val encoding = "UTF-16"
+  val ensimeProject = EnsimeConfigFixture.EmptyTestProject.projects.head.copy(
+    scalacOptions = List("-encoding", encoding)
+  )
   def original = EnsimeConfigFixture.EmptyTestProject.copy(
```
fommil: this could be a bit cleaner. e.g.

```scala
def original = {
  val empty = EnsimeConfigFixture.EmptyTestProject
  empty.copy(empty.projects.head.copy(... ))
}
```

you could even use a shapeless lens, if you wanted to avoid the nesting of `copy` calls. If we depended on Monocle it'd be even cleaner (but we can't).

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,181 @@
...
+
+import akka.actor._
+import akka.event.LoggingReceive.withLabel
+import org.ensime.api._
+import org.ensime.config.richconfig._
+import org.ensime.util.{ Debouncer, Timing }
+import org.ensime.util.FileUtils.toSourceFileInfo
+import org.ensime.util.file._
+
+final case object SuspendAnalyzer
```
fommil: this can be private inside the class

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,181 @@
...
+
+class AnalyzerManager(
+    broadcaster: ActorRef,
+    analyzerCreator: List[EnsimeProjectId] => Props,
+    implicit val config: EnsimeConfig
+) extends Actor with ActorLogging with Stash {
+
+  // is this okay ?
+  private val suspendAnalyzer = Debouncer.forActor(
```
fommil: you don't want a debouncer. You want to a `scheduler` to send the `Suspend` message in 5 minutes, but cancel it every time a message comes in.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
fommil: this should be inside `Analyzer`, not the manager. The manager doesn't need to know anything about the analyzers suspending themselves.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -179,9 +214,22 @@ class Analyzer(
       // disable it when we explicitly want it to be quiet, instead of
       // enabling on every incoming message.
       reporter.enable()
+      lastActivity = clock.millis()
       allTheThings(req)
   }
 
+  def suspended: Receive = withLabel("suspended") {
```
fommil: this looks good

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: But canceling would cancel it for all analyzers out which some might have been inactive for more than the threshold period ?

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -163,12 +189,21 @@ class Analyzer(
     case RestartScalaCompilerReq(id, strategy) =>
       restartCompiler(strategy, id)
 
+    case SuspendAnalyzer =>
+      val now = clock.millis()
```
fommil: I don't know what the time tracking is all about... just ask to shutdown and then become suspended.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: Maybe I could pass it to all analyzers and let them decide whether to go dormant or not based on the time of their last request ?

fommil: see other comment, this shouldn't be in the manager

VC1995: Yes okay, that'll do it.
I just thought, since I was using Debouncer, it might be expensive to have one per analyzer and didn't know if it was wise.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
VC1995: Okay :+1:

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -169,43 +161,29 @@ class Analyzer(
...
     case UnloadAllReq =>
-      restartCompiler(ReloadStrategy.UnloadAll, None)
+      restartCompiler(ReloadStrategy.UnloadAll)
       sender ! VoidResponse
-
     case TypecheckModule(id) =>
-      restartCompiler(ReloadStrategy.LoadProject, Some(id))
-      lastActivity = clock.millis()
+      countdown.cancel()
+      countdown = setCountdown()
```
fommil: you don't need to do it here, look up to the method that calls `doAllTheThings`, you can put the logic in there or there might be a better way to run a side effect before running a partial function.

VC1995: Yes, I don't need it here. Should I simply do this just before calling `allTheThings(req)`?

VC1995:
@fommil Currently we have `RestartCompilerReq` which takes an option of a project id.
One use case I know is that it's used whenever a file is added, changed, or deleted. But currently we pass a None as the project id which causes all the alive(not suspended) PCs to be restarted. (here: https://github.com/VC1995/ensime-server/blob/analyzer-per-project/core/src/main/scala/org/ensime/core/Project.scala#L55) . Can we change this somehow to restart only the pc for that module? Will that be better?

fommil:
btw all `SourceFileInfo` now have an `Option[EnsimeProjectId]` so make sure to look there before trying to infer it (currently clients don't use it but they will)

fommil:
if the restart message only has one message in it, then yeah do that.

But it's more complex for file watching because all downstream analyzers need to be restarted. You could add that logic if you want to send a seq of restart messages.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
```diff
@@ -7,21 +7,27 @@ import java.nio.charset.Charset
...
 
   val encoding = "UTF-16"
-  def original = EnsimeConfigFixture.EmptyTestProject.copy(
-    compilerArgs = List("-encoding", encoding)
-  )
+  def original = {
+    val empty = EnsimeConfigFixture.EmptyTestProject
+    empty.copy(
+      projects = List(empty.projects.head.copy(
+        scalacOptions = List("-encoding", encoding)
```
fommil: adding to the existing ones may be slightly more future proof and unexpected

core/src/main/scala/org/ensime/config/richconfig.scala
```diff
@@ -34,17 +34,19 @@ package object richconfig {
...
       }
 
-    def findProject(path: Path): Option[EnsimeProject] = {
+    def findProject(path: Path): Option[EnsimeProjectId] = {
       // should use NIO relations instead of string comparison...
       // needs https://github.com/ensime/ensime-server/issues/1791
       val p = path.toFile.getPath
-      c.projects.find(_.sources.exists(f => p.startsWith(f.getPath)))
+      c.projects collectFirst {
+        case project if project.sources.exists(f => p.startsWith(f.getPath)) => project.id
```
fommil: since we have `Path` objects, there is no reason to drop down into `String`, we can use `nio` to do a (much more efficient) ancestor check.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -51,17 +52,22 @@ class Analyzer(
     broadcaster: ActorRef,
     indexer: ActorRef,
     search: SearchService,
+    scoped: List[EnsimeProjectId],
```
fommil: I wonder if perhaps we shouldn't just create a version of the `EnsimeConfig` with the `projects` filtered to include only the scoped projects. It would reduce the risk of accidental typechecking of something that we haven't been scoped to see. Principle of least power, and all.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -145,32 +161,27 @@ class Analyzer(
...
-      restartCompiler(ReloadStrategy.UnloadAll, None)
+      restartCompiler(ReloadStrategy.UnloadAll)
       sender ! VoidResponse
-
     case TypecheckModule(id) =>
-      restartCompiler(ReloadStrategy.LoadProject, Some(id))
+      restartCompiler(ReloadStrategy.LoadProject)
       sender ! VoidResponse
-
     case req: RpcAnalyserRequest =>
```
fommil: this means that none of the messages above will cancel the 5 minute timebomb... but since they are not direct user requests I think that's ok.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -179,9 +190,23 @@ class Analyzer(
...
       // enabling on every incoming message.
       reporter.enable()
+      countdown.cancel()
+      countdown = setCountdown()
       allTheThings(req)
   }
 
+  def suspended: Receive = withLabel("suspended") {
+    case req: RpcAnalyserRequest =>
+      // re-start the pc
```
fommil: this comment serves no purpose. A good rule with code comments is that they should only be used to explain something that is not obvious from reading the code.

core/src/main/scala/org/ensime/core/FqnToSymbol.scala
```diff
@@ -50,9 +50,12 @@ trait FqnToSymbol { self: Global with SymbolToFqn =>
       segToSym(nme.segments(scalaName, assumeTerm = assumeTerm), rootSymbol)
     }
   }.getOrElse {
-    val term = segToSym(nme.segments(scalaName, assumeTerm = true), rootSymbol)
```
fommil: I don't think these changes are doing anything... in a medium/large codebase it's best not to change code for stylistic reasons because a) your name will show up in the git history b) it mixes unrelated changes with your feature c) it can cause unexpected regressions that are super hard to track down.

util/src/main/scala/org/ensime/util/Debouncer.scala
```diff
@@ -63,8 +63,8 @@ class Debouncer(
   /** See if it's time to invoke the debounce action */
   private[this] def tryActionOrPostpone(): Unit = {
     val now = clock.millis()
-    val delaySurpassed = ((now - lastCallAttempt) >= delayMs)
-    val maxDelaySurpassed = ((now - lastInvocation) >= maxDelayMs)
+    val delaySurpassed = (now - lastCallAttempt) >= delayMs
```
fommil: unrelated change

util/src/main/scala/org/ensime/util/ensimefile.scala
```diff
@@ -26,7 +26,6 @@ package ensimefile {
 
     /** Direct access contents: not efficient for streaming. */
     def readStringDirect()(implicit cs: Charset): String
-
```
fommil: unrelated change

core/src/main/scala/org/ensime/core/Analyzer.scala
VC1995: Yes, that'll be safer. But `config` should also have `projects` on which the `scoped` projects depend.

core/src/main/scala/org/ensime/config/richconfig.scala
VC1995: Okay

core/src/main/scala/org/ensime/core/FqnToSymbol.scala
VC1995: Okay, I'll remove these changes.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
VC1995: I'm sorry I didn't understand.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -53,7 +53,7 @@ class AnalyzerManager(
     case req @ RestartScalaCompilerReq(id, _) =>
       id match {
         case Some(projectId) =>
-          getOrSpawnNew(projectId) forward req
+          analyzers.getOrElse(projectId, context.system.deadLetters) forward req
```
fommil: `analyzers.get(projectId).foreach(_ forward req)` is better, no need to get `deadLetters` involved

util/src/main/scala/org/ensime/util/Debouncer.scala
```diff
@@ -22,7 +22,7 @@ import scala.concurrent.duration._
  * @param maxDelay The maximum time to delay the invocation of action()
  * @param action The action to invoke
  * @param debounceFirst If false, then the first invocation of call() calls action exactly once. Subsequent calls are debounced.
- * @param Clock The source from which we get the current time. This input should use the same source. Specified for testing purposes
+ * @param clock The source from which we get the current time. This input should use the same source. Specified for testing purposes
```
fommil: irrelevant change

core/src/main/scala/org/ensime/util/Reporter.scala
```diff
@@ -2,13 +2,12 @@
...
 package org.ensime.util
 
-import org.slf4j.LoggerFactory
-
 import scala.reflect.internal.util.Position
 import scala.tools.nsc.reporters.Reporter
 
 import org.ensime.api._
 import org.ensime.core.PositionBackCompat
+import org.slf4j.LoggerFactory
```
fommil: best not change files you're not otherwise touching (on larger projects these kinds of changes can be a real nightmare when you have 50 open PRs with conflicts across 10 team branches, because somebody's IDE "optimised" some imports in a file they weren't even changing)

core/src/main/scala/org/ensime/model/ModelBuilders.scala
```diff
@@ -11,9 +14,6 @@ import org.ensime.util.ensimefile._
 import org.ensime.util.fileobject._
 import org.ensime.vfs._
 
-import scala.collection.mutable
-import scala.reflect.internal.util.{ NoPosition, Position, RangePosition }
```
fommil: imports in untouched file

core/src/main/scala/org/ensime/core/Refactoring.scala
```diff
@@ -221,7 +221,7 @@ trait RefactoringImpl {
           doInlineLocal(procId, tpe, file, start, end)
         }
       case RenameRefactorDesc(newName, file, start, end) =>
-        import scala.reflect.internal.util.{ RangePosition, OffsetPosition }
+        import scala.reflect.internal.util.{ OffsetPosition, RangePosition }
```
fommil: import changes in untouched file

fommil:
@VC1995 almost there, just a few stylistic changes to go! In general it's best not to reformat or change imports in files that you're not touching because it makes later git bisect analysis more complex (e.g. if those files have a regression in them). A good rule of thumb is to try to minimise your footprint, so you don't get pulled into conversations because git says you touched it.

VC1995:
Yes okay, I'm sorry about that. I'll keep it mind. :)

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
VC1995: Okay :+1:

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
fommil: I mean `scalacOptions = "-encoding" :: encoding :: origScalacOptions`

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
```diff
@@ -7,21 +7,28 @@ import java.nio.charset.Charset
...
   val encoding = "UTF-16"
-  def original = EnsimeConfigFixture.EmptyTestProject.copy(
-    compilerArgs = List("-encoding", encoding)
-  )
+  def original = {
+    val empty = EnsimeConfigFixture.EmptyTestProject
+    val origScalacOptions = empty.projects.head.scalacOptions
+    empty.copy(
+      projects = List(empty.projects.head.copy(
+        scalacOptions = origScalacOptions.patch(origScalacOptions.indexOf("-encoding"), List("-encoding", encoding), 2)
```
fommil: this is not what you mean to do

```
List("foo", "bar", "baz", "gaz").patch(-1, List("FOO", "BAR"), 2)
```

you're erasing the first two elements if there is no match for `-encoding`.

Just add to the end of the original list, no need to be fancy.

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -209,7 +209,6 @@ class BasicWorkflow extends EnsimeSpec
                 )
               ) =>
           }
-
```
fommil: changed in otherwise untouched file

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -0,0 +1,171 @@
...
+        }
+        override def receive = {
+          case res: RpcResponse =>
+            original ! res
+            context.stop(self)
+        }
+      }))
+    case req @ CompletionsReq(fileInfo, _, _, _, _) =>
+      withExistingModuleFor(fileInfo, req)((req, moduleId) =>
+        getOrSpawnNew(moduleId) forward req)
```
fommil: this pattern of getting the analyzer and forwarding the request seems like it could be abstracted instead of repeated everywhere.

core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
VC1995: Okay, so we consider the rightmost value of a property if repeated ?

fommil: yes that's right

fommil:
There is a bug in `sbt-ensime` which shows up as lots of red squigglies in test projects. Reverting until it is fixed.

PR #1807, opened 2017-06-20
[Sponsored] Analyzer per Project

VC1995:
Extended `BasicWorkflow` to show that symbol tables are not shared between presentation compilers.

### DIFF
--- a/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
+++ b/core/src/it/scala/org/ensime/core/RefactoringHandlerSpec.scala
@@ -22,9 +22,10 @@ class RefactoringHandlerSpec extends EnsimeSpec
   val encoding = "UTF-16"
   def original = {
     val empty = EnsimeConfigFixture.EmptyTestProject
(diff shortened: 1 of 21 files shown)


VC1995:
I wonder why I sometimes fail "should shutdown if preferred TCP port is not available" on Windows.
Also the added tests work as long as the sbt-ensime fix doesn't come in. Then the test configuration will automatically load the symbols of "compile" configuration we will no longer get the scala notes saying "not found: value Bloo"

fommil:
@VC1995 please upgrade `sbt-ensime` and we'll see how it does

fommil:
of course, I can do it too :smile: let's see how that does

VC1995:
@fommil As expected : `expected class org.ensime.api.NewScalaNotesEvent, found class org.ensime.api.FullTypeCheckCompleteEvent`

VC1995:
I am guessing now the "test" configuration is depending on "compile". So we find the Blue class definition in the "compile" configuration and don't get the `NewScalaNotesEvent` as we earlier expected.
Should I modify the BasicWorkflow according to what we expect now ? :-)

PR #1809, opened 2017-06-20
.ensime-server.conf for user preferences

hardmettle:
Started a new PR for #1346 . Only the preliminary work for server config file. Pending is to : 

- complete the parse method for server config.
- complete testing and integration testing.

@fommil and @yyadavalli please approve the approach.

### DIFF
--- a/core/src/main/scala/org/ensime/config/Environment.scala
+++ b/core/src/main/scala/org/ensime/config/Environment.scala
@@ -31,8 +31,4 @@ object Environment {
 
   private def ensimeVersion: String =
     BuildInfo.version
-
-  def shutdownOnDisconnectFlag: Boolean = {
-    Option(System.getProperty("ensime.explode.on.disconnect")).isDefined
-  }
 }
--- a/core/src/main/scala/org/ensime/core/Canon.scala
+++ b/core/src/main/scala/org/ensime/core/Canon.scala
@@ -5,8 +5,7 @@ package org.ensime.core
 import java.io.File
 
 import scala.util.Try
-
-import org.ensime.api.{ EnsimeFile, EnsimeConfig }
+import org.ensime.api._
 import org.ensime.util.LegacyArchiveExtraction
 import org.ensime.util.ensimefile._
 import org.ensime.util.file._
@@ -22,13 +21,14 @@ object Canon extends Poly1 {
   // people extend File, so we have to handle subtypes
   implicit def caseFile[F <: File]: Case[F] { type Result = File } = at[F](f => f.canon)
 
-  private val legacyJarUrls: Boolean = sys.props.getOrElse("ensime.legacy.jarurls", "true").toBoolean
+  var serverConfig: EnsimeServerConfig = null
+
   var config: EnsimeConfig = null // yes, I know...
   // we really want extractor to be a constructor parameter to a Canon instance
-  private def extractor: Option[LegacyArchiveExtraction] =
-    if (!legacyJarUrls || config == null) None
-    else Some(new LegacyArchiveExtraction(config.cacheDir.file))
 
+  private def extractor: Option[LegacyArchiveExtraction] =
+    if (serverConfig == null || !serverConfig.legacyJarUrls || config == null) None
+    else Some(new LegacyArchiveExtraction(config.cacheDir.file))
   implicit def caseEnsimeFile[EF <: EnsimeFile]: Case[EF] { type Result = EnsimeFile } = at[EF] { f =>
     extractor.flatMap { extractor => Try(extractor.write(f)).toOption }.getOrElse(f).canon
   }
--- a/core/src/main/scala/org/ensime/core/Canon.scala
+++ b/core/src/main/scala/org/ensime/core/Canon.scala
@@ -27,7 +27,7 @@ object Canon extends Poly1 {
   // we really want extractor to be a constructor parameter to a Canon instance
 
   private def extractor: Option[LegacyArchiveExtraction] =
-    if (serverConfig == null || !serverConfig.legacyJarUrls || config == null) None
+    if ((serverConfig != null && !serverConfig.legacy.jarurls) || config == null) None
     else Some(new LegacyArchiveExtraction(config.cacheDir.file))
   implicit def caseEnsimeFile[EF <: EnsimeFile]: Case[EF] { type Result = EnsimeFile } = at[EF] { f =>
     extractor.flatMap { extractor => Try(extractor.write(f)).toOption }.getOrElse(f).canon
--- a/server/src/main/scala/org/ensime/server/Server.scala
+++ b/server/src/main/scala/org/ensime/server/Server.scala
@@ -4,6 +4,7 @@ package org.ensime.server
 
 import java.io._
 import java.net.InetSocketAddress
+
 import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.util._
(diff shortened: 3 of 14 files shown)


core/src/main/scala/org/ensime/config/Environment.scala
```diff
@@ -32,7 +34,5 @@ object Environment {
   private def ensimeVersion: String =
     BuildInfo.version
 
-  def shutdownOnDisconnectFlag: Boolean = {
-    Option(System.getProperty("ensime.explode.on.disconnect")).isDefined
-  }
+  def shutdownOnDisconnectFlag(implicit serverConfig: EnsimeServerConfig): Boolean = serverConfig.shutDownOnDisconnect
```
fommil: there is no reason to have this method anymore if the caller can read it off the config object

core/src/main/scala/org/ensime/core/Canon.scala
```diff
@@ -22,7 +21,8 @@ object Canon extends Poly1 {
   // people extend File, so we have to handle subtypes
   implicit def caseFile[F <: File]: Case[F] { type Result = File } = at[F](f => f.canon)
 
-  private val legacyJarUrls: Boolean = sys.props.getOrElse("ensime.legacy.jarurls", "true").toBoolean
+  var serverConfig: EnsimeServerConfig = null
+  private val legacyJarUrls: Boolean = serverConfig.legacyJarUrls
```
fommil: no need to have this `val` if it can be read from the config

server/src/main/scala/org/ensime/server/Server.scala
```diff
@@ -115,13 +117,23 @@ object Server extends AkkaBackCompat {
   Slf4jSetup.init()
 
   val log = LoggerFactory.getLogger("Server")
+  val XDG_CONFIG_HOME = " ~/.config/ensime-server"
+
+  lazy val shutDownOnDisconnect = Option(System.getProperty("ensime.explode.on.disconnect")).isDefined
```
fommil: if we're using typesafe config you can move all of these into a `reference.conf`

server/src/main/scala/org/ensime/server/Server.scala
```diff
@@ -115,13 +117,23 @@ object Server extends AkkaBackCompat {
   Slf4jSetup.init()
 
   val log = LoggerFactory.getLogger("Server")
+  val XDG_CONFIG_HOME = " ~/.config/ensime-server"
```
fommil: we want to look for files called `.ensime-server.conf` in the project root and then merge in (with lower priority) a `~/.ensime-server.conf`

server/src/main/scala/org/ensime/server/Server.scala
hardmettle: A bit confused here. You mean to fetch ``XDG_CONFIG_HOME``  value from ``reference.conf`` ? At the moment we dont have a ``reference.conf`` , does that have to be created ?

server/src/main/scala/org/ensime/server/Server.scala
hardmettle: Yes, this has been taken care of , just need to change the typo for the file name .

server/src/main/scala/org/ensime/server/Server.scala
fommil: I mean all defaults

server/src/main/scala/org/ensime/server/Server.scala
yyadavalli: @hardmettle we have three file to check for with priority order

1. project-root/.ensime-server.conf
2. $XDG_CONFIG_HOME(or ~/.config/)/ensime/server.conf -- this is in compliance with xdg directory specification we can deviate if we choose to.
3. ~/.ensime-server.conf

server/src/main/scala/org/ensime/server/Server.scala
yyadavalli: `XDG_CONFIG_HOME` is an environment variable, if it is not present the fallback should be `~/.config/`, which can be defined in our `reference.conf`

server/src/main/scala/org/ensime/server/Server.scala
fommil: I'm ok with the xdg home, then we don't need 3.

This does mean the project root becomes relevant again, so you can remove the deprecation warning / comment that I added in `config.scala`

core/src/main/scala/org/ensime/core/Canon.scala
fommil: indeed, should be a `def` because `serverConfig` should be set before `Canon` is accessed.

server/src/main/scala/org/ensime/server/Server.scala
```diff
@@ -115,13 +117,24 @@ object Server extends AkkaBackCompat {
   Slf4jSetup.init()
 
   val log = LoggerFactory.getLogger("Server")
+  val appConf = ConfigFactory.load("ensime")
+  val serverAppConf = appConf.getConfig("server-conf")
+
+  lazy val XDG_CONFIG_HOME = appConf.getString("xdg-conf-home")
```
yyadavalli: This should be fall back to an environment variable `XDG_CONFIG_HOME`

application.conf
```diff
@@ -0,0 +1,12 @@
...
+  server-conf = {
+    shutDownOnDisconnect = true
+    test = true
+    sys-protocol = "swank"
+    exitAfterIndex = true
+    disableSourceMonitoring = true
+    disableClassMonitoring = true
+    legacyJarUrls = true
+  }
+  xdg-conf-home = "~/.config/.ensime-server.conf"
```
yyadavalli: Leave `xdg-conf-home` as a directory and append filename later just for consistency.

yyadavalli:
@hardmettle Just have a couple of small changes. 

I think its better to use a `Sexp` formatted file for server configuration as well, with that you can use the current `SexpFormat` format to implement `parse`

The way I see it, using typesafe config is a different concern and we can handle it later, once we write a converter between `Config` and our case classes.

yyadavalli:
@hardmettle  you need to update the tests

fommil:
@yyadavalli the problem with an sexp config is it can't be a subset. If we use typesafe config then the user can override one thing and it won't break on upgrades.

We can have a handwritten converter.

yyadavalli:
@fommil I did not think about that, just wanted to make it easier to merge this piece in. We can have a handwritten converter for now.

server/src/main/scala/org/ensime/server/Server.scala
hardmettle: @yyadavalli whats the command to fetch it from env variable ?

hardmettle:
@fommil @yyadavalli All the failing tests are from `FileWatcherSpec.scala` . I cant exactly find the root cause of it . Any leads on it would help me find the issue here.

yyadavalli:
@hardmettle you can use `scala.util.Properties.envOrElse("HOME", "/myhome" )`

server/src/main/scala/org/ensime/server/Server.scala
yyadavalli: you can use `scala.util.Properties.envOrElse("HOME", "/myhome" )`

yyadavalli:
For the test failures, you need to add implicit variable in `ProjectFixture.startup()`

hardmettle:
Hi @yyadavalli adding the implicit `serverConfig:EnsimeServerConfig` to `ProjectFixture.startup()` isnt helping. `FileWatcherSpec` are still failing.

yyadavalli:
@hardmettle You are just adding the parameter, you have to make sure an implicit value is available when the function is being called or you have to pass the parameter

hardmettle:
Just wanted a bit clarification for understanding . `ProjectFixture.scala` is not being used by `FileWatcherSpec.scala` so why should that variable be added to it. Also  I made the implicit value available as well and it's still failing for `FileWatcherSpec.scala`

fommil:
I'm consistently getting this locally

```
[error] Failed tests:
[error] 	org.ensime.intg.CompileTimingTest
```

and a few others. I can't think what could be the problem.

yyadavalli:
@fommil I will take a look and try to fix it next week. I will be done with finals by then.

_Sent from my OnePlus ONEPLUS A5000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

fommil:
@yyadavalli I got this :smile: good luck!

fommil:
I'm expecting green lights this time.

yyadavalli:
Thanks @fommil, sorry I couldn't finish this off earlier.

Issue #1812, opened 2017-06-22
third party scala files no longer typechecked


fommil:
regression in #1807 we no longer handle requests for scala files from the cache (or jar urls). This is the typical log message when jumping to source for `scala.Int` (e.g. in the ensime-server project itself)

```
10:20:39.685 DEBUG akka://ENSIME/user/ensime-main/project/scalac o.e.c.AnalyzerManager - received handled message SymbolDesignationsReq(Right(SourceFileInfo(RawFile(/tmp/fommil-ensime/home/fommil/Projects/drone-dynamic-agents/.ensime_cache/dep-src/source-jars/scala/Int.scala),None,None)),0,16415,List(VarSymbol, ValSymbol, VarFieldSymbol,
10:20:39.686 DEBUG akka://ENSIME/user/ensime-main/tcp-server/con1/21 o.e.s.RequestHandler - received handled message EnsimeServerError(Couldn't find the project for RawFile(/tmp/fommil-ensime/home/fommil/Projects/drone-dynamic-agents/.ensime_cache/dep-src/source-jars/scala/Int.scala)) from Actor[akka://ENSIME/user/ensime-main/project/scalac
```

we should probably forward these to Sauron unless the client requested to use a particular ensime project id.

VC1995:
#1814

jozi-k:
probably fixed

PR #1814, opened 2017-06-22
Send requests outside of any project to sauron

VC1995:
#1812 
@fommil Is this okay?

### DIFF
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -97,7 +97,7 @@ class BasicWorkflow extends EnsimeSpec
           val sourceFileLocation = symInfo.`type`.pos.get.asInstanceOf[OffsetSourcePosition].file
           project ! SymbolDesignationsReq(Right(SourceFileInfo(sourceFileLocation, None, None)), -1, 9999, SourceSymbol.allSymbols)
           val symbolDesignations = expectMsgType[SymbolDesignations]
-          symbolDesignations.syms should not be List.empty
+          symbolDesignations.syms should not be empty
 
           //-----------------------------------------------------------------------------------------------
           // public symbol search - java.io.File
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -94,6 +94,11 @@ class BasicWorkflow extends EnsimeSpec
             case Some(OffsetSourcePosition(_, _)) =>
           }
 
+          val sourceFileLocation = symInfo.`type`.pos.get.asInstanceOf[OffsetSourcePosition].file
+          project ! SymbolDesignationsReq(Right(SourceFileInfo(sourceFileLocation, None, None)), -1, 9999, SourceSymbol.allSymbols)
+          val symbolDesignations = expectMsgType[SymbolDesignations]
+          symbolDesignations.syms should not be List.empty
+
           //-----------------------------------------------------------------------------------------------
           // public symbol search - java.io.File
 
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -86,7 +86,23 @@ class BasicWorkflow extends EnsimeSpec
           //-----------------------------------------------------------------------------------------------
           // symbolAtPoint
           project ! SymbolAtPointReq(Left(fooFile), 128)
-          val symbolAtPointOpt: SymbolInfo = expectMsgType[SymbolInfo]
+          expectMsgType[SymbolInfo]
+
+          project ! SymbolAtPointReq(Left(fooFile), 68) // scala.Int symbol
+          val sourcePosition = expectMsgType[SymbolInfo].`type`.pos.get
+          val sourceFile = sourcePosition match {
+            case OffsetSourcePosition(file, _) =>
+              file
+            case _ =>
+              EnsimeFile("empty")
+          }
+
+          project ! TypecheckFilesReq(List(Right(SourceFileInfo(sourceFile, None, None))))
+          expectMsg(VoidResponse)
+          asyncHelper.expectMsg(FullTypeCheckCompleteEvent)
+
+          project ! SymbolAtPointReq(Right(SourceFileInfo(sourceFile, None, None)), 720)
+          expectMsg(SymbolInfo("scala", "scala", None, BasicTypeInfo("scala", DeclaredAs.Object, "scala")))
 
           //-----------------------------------------------------------------------------------------------
           // public symbol search - java.io.File
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -89,20 +89,10 @@ class BasicWorkflow extends EnsimeSpec
           expectMsgType[SymbolInfo]
 
           project ! SymbolAtPointReq(Left(fooFile), 68) // scala.Int symbol
-          val sourcePosition = expectMsgType[SymbolInfo].`type`.pos.get
-          val sourceFile = sourcePosition match {
-            case OffsetSourcePosition(file, _) =>
-              file
-            case _ =>
-              EnsimeFile("empty")
+          val symInfo = expectMsgType[SymbolInfo]
+          symInfo.`type`.pos should matchPattern {
+            case Some(OffsetSourcePosition(_, _)) =>
           }
-
-          project ! TypecheckFilesReq(List(Right(SourceFileInfo(sourceFile, None, None))))
-          expectMsg(VoidResponse)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -44,7 +45,7 @@ class AnalyzerManager(
       case Some(moduleId) =>
         f(req, moduleId)
       case None =>
-        sender ! EnsimeServerError(s"Couldn't find the project for ${fileInfo.file}")
+        sauron ! EnsimeServerError(s"Couldn't find the project for ${fileInfo.file}")
```
fommil: this doesn't look right

fommil:
I don't think this is the right way to do it... I actually missed that we dropped `sauron`, I thought we were sending refactor requests to the "all seeing" `Analyzer`?

I think it's as simple as having a fallback analyzer in the case that the `EnsimeProjectId` cannot be resolved.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -45,7 +45,7 @@ class AnalyzerManager(
       case Some(moduleId) =>
         f(req, moduleId)
       case None =>
-        sauron ! EnsimeServerError(s"Couldn't find the project for ${fileInfo.file}")
+        sauron forward req
```
fommil: pretty much :smile: but I think we need a test. Maybe send a request for semantic symbols for `/scala/Int.scala` in the scala-library source jar or something?

fommil: and if you put the logic in `getOrSpawnNew` then you don't even need to change most of the existing logic.

VC1995: I'll still need to make a few changes in the case statement of `TypeCheckFilesReq`.
I will add some test. :)

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -86,7 +86,23 @@ class BasicWorkflow extends EnsimeSpec
...
+              file
+            case _ =>
+              EnsimeFile("empty")
+          }
+
+          project ! TypecheckFilesReq(List(Right(SourceFileInfo(sourceFile, None, None))))
+          expectMsg(VoidResponse)
+          asyncHelper.expectMsg(FullTypeCheckCompleteEvent)
+
+          project ! SymbolAtPointReq(Right(SourceFileInfo(sourceFile, None, None)), 720)
```
fommil: this is not going to be stable across scala versions (even minor releases)

VC1995: I think it's okay till typechecking the source file. The `SymbolAtPointReq` won't work, right?

fommil: we shouldn't be considering the `empty` case... if that doesn't work something failed. I guess typecheck file should be enough. But maybe asking for the semantic information instead of the symbol-at-point would be better, and just assert we get an non-empty response.

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -73,44 +74,46 @@ class AnalyzerManager(
       withExistingModuleFor(fileInfo, req)((req, moduleId) =>
         getOrSpawnNew(moduleId) forward req)
     case req @ TypecheckFilesReq(files) =>
-      if (files.exists(config.findProject(_).isEmpty))
-        sender ! EnsimeServerError("Update .ensime file.")
-      else {
-        val original = sender
-        val filesPerProject = files.groupBy(config.findProject(_)).map(x => x._1.get -> x._2)
+      val original = sender
+      val filesPerProject = files.groupBy(config.findProject(_)).map(x => x._1 -> x._2)
```
fommil: I know this was in the original code, but `._` getters are discourage. Prefer ` { case (a, b) => } ` and give meaningful names to `a` and `b`.

Ummm, is `map(x => x._1 -> x._2)` even doing anything? Can't we just drop the `map`?

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
VC1995: What do you mean by asking for semantic information? 
I just added `case _ => EnsimeFile("empty")` to get rid of the warning. 
I kind of copied `GoToDefinition`'s idea. Getting location of source file from the `TypeInfo`.

fommil: if we don't get an `OffsetSourcePosition`, the test should fail

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -44,7 +45,7 @@ class AnalyzerManager(
       case Some(moduleId) =>
         f(req, moduleId)
       case None =>
-        sender ! EnsimeServerError(s"Couldn't find the project for ${fileInfo.file}")
+        sauron forward req
```
fommil: this is a bit hacky... the `f` doesn't get called if there is no `moduleId`. Perhaps it's time to rethink this borrow pattern?

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -73,44 +74,46 @@ class AnalyzerManager(
...
+        private var aggregate: List[String] = List.empty
 
-        context.actorOf(Props(new Actor {
-          private var remaining = filesPerProject.size
-          private var aggregate: List[String] = List.empty
+        override def preStart: Unit =
+          for ((optionalModuleId, list) <- filesPerProject)
+            optionalModuleId match {
+              case Some(moduleId) =>
+                getOrSpawnNew(moduleId) ! TypecheckFilesReq(list)
```
fommil: I still thing `getOrSpawnNew` should return `sauron` if it doesn't match, which means it doesn't return `Option` anymore.

fommil:
LOL I kept getting errors about this today and I was thinking "man, that fix really didn't work"... then I realised this hasn't been merged yet :smile:

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -86,7 +86,13 @@ class BasicWorkflow extends EnsimeSpec
...
           // symbolAtPoint
           project ! SymbolAtPointReq(Left(fooFile), 128)
-          val symbolAtPointOpt: SymbolInfo = expectMsgType[SymbolInfo]
+          expectMsgType[SymbolInfo]
+
+          project ! SymbolAtPointReq(Left(fooFile), 68) // scala.Int symbol
+          val symInfo = expectMsgType[SymbolInfo]
+          symInfo.`type`.pos should matchPattern {
+            case Some(OffsetSourcePosition(_, _)) =>
+          }
```
fommil: we're missing the rest of the test here... we still need to lead the `scala.Int` source code and then do something with it, e.g. a `SymbolDesignationsReq`

fommil:
needs a rebase, sorry

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -94,6 +94,11 @@ class BasicWorkflow extends EnsimeSpec
             case Some(OffsetSourcePosition(_, _)) =>
           }
 
+          val sourceFileLocation = symInfo.`type`.pos.get.asInstanceOf[OffsetSourcePosition].file
+          project ! SymbolDesignationsReq(Right(SourceFileInfo(sourceFileLocation, None, None)), -1, 9999, SourceSymbol.allSymbols)
+          val symbolDesignations = expectMsgType[SymbolDesignations]
+          symbolDesignations.syms should not be List.empty
```
fommil: in scalatest you can just say `should not be ('empty)` and it works for any collection (check my syntax)

core/src/main/scala/org/ensime/core/AnalyzerManager.scala
```diff
@@ -115,7 +115,7 @@ class AnalyzerManager(
     case req @ CompletionsReq(fileInfo, _, _, _, _) =>
       getOrSpawnNew(config.findProject(fileInfo)) forward req
     case req @ UsesOfSymbolAtPointReq(file, _) =>
-        getOrSpawnNew(config.findProject(file)) forward req
+      getOrSpawnNew(config.findProject(file)) forward req
     case req @ SymbolAtPointReq(file, point: Int) =>
       getOrSpawnNew(config.findProject(file)) forward req
```
fommil: this pattern is repeating a lot. I wonder if perhaps it would be cleaner to have a method (typeclass) that extracts the `SourceFileInfo` / `Option[EnsimeProjectId]` then we don't need to repeat the business logic.

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
VC1995: Okay I'll do that :-)

PR #1822, opened 2017-06-30
Fork spray json

### LINKED ISSUE #1821: internal fork of spray-json and spray-json-shapeless
- to generally reduce external dependencies, part of strategic #1398 
- to avoid the ScalaJSON clusterfuck https://github.com/spray/spray-json/issues/232

fommil:
close #1821 and avoid the absolute train wreck that is ScalaJSON.

@sirthias / @jrudolph if there were ever any binary / source breaking changes you wanted to make to `spray-json` please feel free to submit them here as therapy! :smile:

I noticed that `spray-json` doesn't have a `NOTICE` file (it really should, otherwise you don't get attribution as per Apache 2.0). I've taken the liberty of creating one for you and distributing it in the `resources` directory, which is an often overlooked requirement of the license (most people only distribute the `LICENSE` / `NOTICE` with the source and forget to include it in the binary).


json/src/main/scala/spray/json/JsonParser.scala
```diff
@@ -190,7 +190,7 @@ class JsonParser(input: ParserInput) {
 
   @tailrec private def ws(): Unit =
     // fast test whether cursorChar is one of " \n\r\t"
-    if (((1L << cursorChar) & ((cursorChar - 64) >> 31) & 0x100002600L) != 0L) { advance(); ws() }
+    if (((1L << cursorChar.toInt) & ((cursorChar - 64) >> 31) & 0x100002600L) != 0L) { advance(); ws() }
```
fommil: looks like scalac missed `(cursorChar - 64)` for numeric widening

json/src/main/scala/spray/json/JsonParser.scala
```diff
@@ -0,0 +1,347 @@
...
+    sb.setLength(0)
+    while (`char`()) cursorChar = input.nextUtf8Char()
+    require('"')
+    ws()
+  }
+
+  private def `char`() =
+    // simple bloom-filter that quick-matches the most frequent case of characters that are ok to append
+    // (it doesn't match control chars, EOI, '"', '?', '\', 'b' and certain higher, non-ASCII chars)
+    if (((1L << cursorChar.toInt) & ((31L - cursorChar.toLong) >> 31) & 0x7ffffffbefffffffL) != 0L) appendSB(cursorChar)
```
fommil: I'd be kidding if I said I understood this line.

fommil:
Damn, dupe license entries. Trivial fix incoming.

json/src/main/scala/spray/json/JsonParser.scala
sirthias: It's a micro-optimization reducing the most common code path to a single branch instruction. Its effect is probably barely noticeable, but you have to admit that it looks cool... :)

fommil:
Needs a fix for the micro optimisation and comment explaining how it works.

Issue #1802, opened 2017-06-18
file watchers should only restart relevant analyzers


fommil:
Once #1152 is merged, we should make our file watchers smarter. Source watchers should only restart the one module with the source in it, and classfile watchers should restart only the analyzer and its downstream dependencies.

fommil:
@VC1995 something to work on alongside the sublime work, so you don't run of Scala coding entirely? :smile:

VC1995:
Yep, okay :)

PR #1824, opened 2017-07-01
Smarter FileChangeListener in Project

VC1995:
#1802 
WIP

### DIFF
--- a/core/src/main/scala/org/ensime/config/richconfig.scala
+++ b/core/src/main/scala/org/ensime/config/richconfig.scala
@@ -34,7 +34,7 @@ package object richconfig {
 
     def findProject(path: Path): Option[EnsimeProjectId] = {
       c.projects collectFirst {
-        case project if project.sources.exists(f => path.startsWith(f.file)) => project.id
+        case project if (project.sources ++ project.targets).exists(f => path.startsWith(f.file)) => project.id
       }
     }
     def findProject(file: EnsimeFile): Option[EnsimeProjectId] = file match {
--- a/core/src/main/scala/org/ensime/config/richconfig.scala
+++ b/core/src/main/scala/org/ensime/config/richconfig.scala
@@ -7,10 +7,12 @@ import java.nio.file.Path
 
 import scala.collection.breakOut
 
+import org.apache.commons.vfs2.FileObject
 import org.ensime.api._
 import org.ensime.util.file._
 import org.ensime.util.path._
 import org.ensime.util.ensimefile._
+import org.ensime.vfs.`package`.EnsimeVFS
 
 package object richconfig {
 
@@ -31,6 +33,15 @@ package object richconfig {
         val name = f.getName
         name.startsWith("scala-library") && name.endsWith(".jar")
       }
+
+    def findProject(f: FileObject)(implicit vfs: EnsimeVFS) = {
+      val filePath = f.getName.getPath
+      c.projects collectFirst {
+        case project if (project.sources ++ project.targets).map(rf =>
+          vfs.toFileObject(rf.file.toFile).getName.getPath).exists(filePath.startsWith) =>
+          project.id
+      }
+    }
 
     def findProject(path: Path): Option[EnsimeProjectId] = {
       c.projects collectFirst {
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -9,6 +9,7 @@ import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.util._
 import scala.util.Properties._
+import scala.collection.mutable
 
 import akka.actor._
 import akka.event.LoggingReceive.withLabel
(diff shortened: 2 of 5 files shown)


fommil:
Oh awesome, are you getting bored of python?

VC1995:
Maybe a little. Then I realized I had this to do. :-)
Also, does the `SourceWatcher` restart the analyzers at all? I gets the resolver as the `FileChangeListener` which only calls `this.update()`  when a file is added, changed or removed.

fommil:
I think we should delete the source watchers. Feel free to do that.

VC1995:
#1825

fommil:
tests are failing

VC1995:
Yes, that's because I have assumed `File` instead of `FileObject` in `Project`. It should work once those changes are merged.

VC1995:
@fommil What do you think?

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -47,18 +48,38 @@ class Project(
...
+
+  private case object AskReTypeCheck
+
+  // map to get all projects that list the key project in 'depends'
+  private val dependentProjects: Map[EnsimeProjectId, List[EnsimeProjectId]] =
+    config.projects.map(
+      p => p.id -> config.projects.filter(_.depends.contains(p.id)).map(_.id)
+    ).toMap
+
+  val restartProjects: mutable.Queue[List[EnsimeProjectId]] = mutable.Queue.empty
```
fommil: I'm not keen on using mutable state to communicate the message.

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -49,29 +48,28 @@ class Project(
...
+        RestartScalaCompilerReq(None, ReloadStrategy.KeepLoaded),
         delay = (5 * Timing.dilation).seconds,
         maxDelay = (20 * Timing.dilation).seconds
       )
     def fileChanged(f: FileObject): Unit = {
       val projectId = config.findProject(f)
       projectId foreach { projectId =>
-        restartProjects.enqueue(projectId :: dependentProjects.getOrElse(projectId, Nil))
-        askReTypeCheck.call()
+        projectActor ! AskReTypeCheck(projectId :: dependentProjects.getOrElse(projectId, Nil))
```
fommil: this now bypasses the debouncer. It seems like we might want to debounce on individual projects.

VC1995: Yeah, I realise that. The roadblock was not being able to send any parameters to call method hence came up with the initial mutable queue.

fommil: we could have a debouncer for each project

VC1995: Yep! I'll do that :)

VC1995:
Appveyor is doing this again : should shutdown if preferred TCP port is not available *** FAILED ***

VC1995:
And travis-ci : The job exceeded the maximum time limit for jobs, and has been terminated.

fommil:
It's because USA is awake :-)

This is looking good now, I'll review again and merge after I restart the tests tomorrow.

core/src/main/scala/org/ensime/config/richconfig.scala
```diff
@@ -32,9 +34,18 @@ package object richconfig {
         name.startsWith("scala-library") && name.endsWith(".jar")
       }
 
+    def findProject(f: FileObject)(implicit vfs: EnsimeVFS) = {
```
fommil: why didn't you just convert the `FileObject` into a `Path` and call the existing method?

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -47,19 +47,35 @@ class Project(
   private implicit val vfs: EnsimeVFS = EnsimeVFS()
   private val resolver = new SourceResolver(config)
   private val searchService = new SearchService(config, resolver)
-  private val sourceWatcher = new SourceWatcher(config, resolver :: Nil)
+
+  // map to get all projects that list the key project in 'depends'
```
fommil: no need for the comment

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -47,19 +47,35 @@ class Project(
...
+  private val dependentProjects: Map[EnsimeProjectId, List[EnsimeProjectId]] =
+    config.projects.map(
+      p => p.id -> config.projects.filter(_.depends.contains(p.id)).map(_.id)
+    ).toMap
+
+  // FileChangeListener for ClassFileWatcher
```
fommil: no need for comment

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -47,19 +47,35 @@ class Project(
   private implicit val vfs: EnsimeVFS = EnsimeVFS()
   private val resolver = new SourceResolver(config)
   private val searchService = new SearchService(config, resolver)
-  private val sourceWatcher = new SourceWatcher(config, resolver :: Nil)
+
+  // map to get all projects that list the key project in 'depends'
+  private val dependentProjects: Map[EnsimeProjectId, List[EnsimeProjectId]] =
```
fommil: idiomatically, this would be done in `preStart` to avoid doing work on the `Actor` initialisation thread.

core/src/main/scala/org/ensime/config/richconfig.scala
VC1995: I tried that but it seems it returns a Unix type path, didn't work on Windows.

core/src/main/scala/org/ensime/core/Project.scala
VC1995: Should I make it a var and transfer the initialization to `preStart` ?

fommil: sadly, yes (I know these rules about `var` are inconsistent, but that's `Actor`s for you...)

VC1995: What about a lazy val and cal it in `preStart` ?

fommil: that's not really idiomatic to Akka

core/src/main/scala/org/ensime/config/richconfig.scala
fommil: ok, now **that** is the sort of thing that should go into a comment. Anything "unexpected" should be documented, otherwise the reader just thinks you missed something :wink:

Issue #1827, opened 2017-07-03
UsesOfSymbolAtPointReq should not try to solve symbols from classfiles


hzenginx:
When using UsesOfSymbolAtPoint feature on any symbol defined in classfile ensime hangs with NPE.

an example project can be any project using external libraries, including test projects in ensime.

Reproduction instructions:
 - Try to find uses of a function from standart libraries, example `"SomeString".length`
 - Kill ensime process from outside(`kill $PID`)
 - Check logs for a NPE 

Reason: Ensime is using `Symbol.sourceFile` to get source code, which returns `null` if symbol is defined in a classfile.

- [x] I have read http://ensime.org/getting_help/
- [x] I have followed the troubleshooting guide for my text editor
- [x] I have asked in http://gitter.im/ensime/ensime-server
- [x] I have included a link to an example project
- [x] I have included exact bug reproduction instructions

fommil:
the real problem is that this API is pretty rubbish and should be replaced by https://github.com/ensime/ensime-server/issues/1753

PR #1828, opened 2017-07-03
UsesOfSymbolAtPointReq, do not try to solve symbols from classfiles (#1827)

hzenginx:
I am not sure if there is a way to find usages of a symbol coming from classfiles, Intellij IDEA CE is also not able to do this.
This pr only prevents NPEs and returns a empty list as usages.

UsesOfSymbolAtPointReq, do not try to solve symbols from classfiles (#1827)

### DIFF
--- a/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
+++ b/core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
@@ -340,6 +340,14 @@ class BasicWorkflow extends EnsimeSpec
           expectMsg(VoidResponse)
 
           asyncHelper.expectMsg(FullTypeCheckCompleteEvent)
+
+          //-----------------------------------------------------------------------------------------------
+          // uses of symbol at point with a symbol defined in classfiles
+
+          project ! UsesOfSymbolAtPointReq(Left(fooFile), 162) // point on String.length
+          expectMsgType[ERangePositions].positions shouldBe empty
+
+          asyncHelper.expectMsg(FullTypeCheckCompleteEvent)
           asyncHelper.expectNoMsg(3 seconds)
         }
       }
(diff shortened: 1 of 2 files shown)


core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -146,6 +146,20 @@ class BasicWorkflow extends EnsimeSpec
...
           }
 
+          //-----------------------------------------------------------------------------------------------
+          // uses of symbol at point with a symbol defined in classfiles
+
+          project ! TypecheckFilesReq(List(Left(fooFile)))
+          expectMsg(VoidResponse)
+
+          project ! UsesOfSymbolAtPointReq(Left(fooFile), 162) // point on String.length
+          expectMsgType[ERangePositions].positions should contain theSameElementsAs List()
```
fommil: `should be empty` is cleaner

core/src/it/scala/org/ensime/intg/BasicWorkflow.scala
```diff
@@ -146,6 +146,20 @@ class BasicWorkflow extends EnsimeSpec
...
+
+          asyncHelper.fishForMessage() {
```
fommil: we've found these `fishForMessage` waiting blocks to be very unreliable. Please try to work out exactly what messages are being returned and wait for them. It is probably safer to put your test at the very end.

hzenginx: This test is really hard to extend :/ I moved the new code to end of the file

hzenginx:
Sure I will try to find all messages. Build is already failed so It makes sense to really figure out all messages 

I am interested in doing that but first I want to merge this because I am using emacs+ensime as  my primary environment at my job and this bug is annoying :P

fommil:
Yeah, I leave bugs in intentionally to trick people into contributing muahahaha!

(Did you see the instructions to use your own build? Really useful for when you have a fix in PR but not merged yet)

hzenginx:
Looks like your plan is working well, this is the second time I got trapped 😄 
I am using my build currently BTW but nothing is better than having it in regular ways.

I added another expected message, hopefully this should work but this test is too complicated to change :(

hzenginx:
Is AppVeyor broken?

fommil:
Yeah... it's increasingly getting stuck at the serverstartspec.

hzenginx:
Should I push some changes just to restart that one, otherwise the tests looks pretty fine to me now

fommil:
Thanks! But damn I meant to squash merge.

PR #1829, opened 2017-07-04
[WIP] Removing apache commons VFS

jozi-k:
This is just start, basically I want to know if I am not doing something horribly wrong as I have never met vfs before.

### DIFF
--- a/api/src/main/scala/org/ensime/api/common.scala
+++ b/api/src/main/scala/org/ensime/api/common.scala
@@ -3,7 +3,9 @@
 package org.ensime.api
 
 import java.io.File
-import java.nio.file.Path
+import java.nio.file.{ Files, FileSystem, FileSystems, Path, Paths }
+import java.net.{ URI, URL, URLDecoder }
+import java.util.HashMap
 
 import scala.annotation.StaticAnnotation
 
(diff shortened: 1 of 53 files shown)


jozi-k:
Forgot to compile (it)tests :(

fommil:
awesome, it would be really great if you worked on this! But instead of passing `Path` around, use `EnsimeFile` ... you'll constantly hit the problem of wondering if something is a file or a jarfile entry and that's exactly what `EnsimeFile` is for.

jozi-k:
Thanks for feedback and advice @fommil

api/src/main/scala/org/ensime/api/common.scala
```diff
@@ -129,10 +131,80 @@ object OffsetRange extends ((Int, Int) => OffsetRange) {
 // such as files/dirs, existance, content hints
 // (java/scala/class/resource) in the type, validated at construction
 // (and can be revalidated at any time)
-sealed trait EnsimeFile
-final case class RawFile(file: Path) extends EnsimeFile
+sealed trait EnsimeFile {
+  def path: Path
```
fommil: please no methods on this ADT. It is a pure data structure.

fommil: in fact, no code in `api` is the rule. This remains a clean data structure as it is exported and viewed across languages. It is also Apache2 licensed whereas the code you copied is GPLv3.

jozi-k: Hi @fommil, thanks for comments. You are apparently expert in licensing topic, but my observation is that 5-sigma of devs want to code.

Pure data structures cannot do anything useful (besides storing state), for "interesting" stuff we have methods. Or am I missing something? I think `ensimefile.scala` is overusing implicits, therefore I moved it to `EnsimeFile` trait. I believe same approach is used in scala itself, see `List` trait.
What kind of rule do you have in mind?
Is there any reason to use `EnsimeFile` outside of ensime-server?
What "languages" view `EnsimeFile`?
Is there any reason why `util` code uses different license than `api` code?
If some code is deleted from file using GPLv3, does it mean it cannot be added to file which uses Apache2?

Sorry for so many questions, but I always want to learn something new and having discussion is so far the best way of doing that.

fommil: you cannot relicense GPLv3 code under Apache 2.0, it would change the license of the `api` module and then our downstream developers would be very annoyed because they'd have to relicense their applications or fork ensime-server.

If you want to know about the benefits of separating data and functionality, read chapter 4 of my book. In this particular case, consider an Emacs, Atom, VSCode or Vim developer looking at our API, why would they care about 100 lines of distracting Scala code when they just want to see the data structure to be able to interpret the wire protocol format? Furthermore, within ensime itself if we move to `EnsimeFile` there is a **lot** of data conveyed when a method accepts an `EnsimeFile`, `ArchiveFile` or `RawFile`. Indeed, the problem with `FileObject` is that it tries to do exactly what you describe (have lots of methods) but what we really need is the contextual understanding of *what* we have in our hands.

Another example are the data structures we use for storage in the indexer. The actual implementation, which uses large OOP classes, are so complicated to explain that the only way I could describe them for the Scala Sphere talk was to rewrite them as pure data structures and type classes.

Basically splitting data structure and functionality leads to cleaner modular code that is easier to reason about.

fommil: Book link: https://leanpub.com/fp-scala-mortals

jozi-k:
I removed code from api. Can you @fommil please have a look at problems with `cachedImplicit`?

jozi-k:
I am still fighting with implicit cache:
```[error] /ensime-server/core/src/main/scala/org/ensime/indexer/graph/GraphService.scala:383: Could not find an implicit value of type org.ensime.util.stringymap.api.BigDataFormat[org.ensime.indexer.graph.FileCheck] to cache```

**Comment by fommil** [comment]
we only show one because often the others are lies and it keeps the noise down. You can easily change it locally, look for `showErrors`.

I think you actually need to provide an instance of `BigDataFormat` for whatever you replaced `FileObject` with in `FileCheck`. This is a terrible thing in scala implicit errors and I'm working on a fix: https://github.com/fommil/stalactite but will be a while before I have the time to bring it in to ensime.

Make sure it serialises out to exactly the same format as before or we'll have to change the db schema number.

File: core/src/test/scala/org/ensime/indexer/FileWatcherSpec.scala [hunk 2]
```diff
@@ -18,11 +19,11 @@ import org.scalatest.tagobjects.Retryable
 import org.scalatest.time._
 
 sealed trait FileWatcherMessage
-final case class Added(f: FileObject) extends FileWatcherMessage
-final case class Removed(f: FileObject) extends FileWatcherMessage
-final case class Changed(f: FileObject) extends FileWatcherMessage
-final case class BaseAdded(f: FileObject) extends FileWatcherMessage
-final case class BaseRemoved(f: FileObject) extends FileWatcherMessage
+final case class Added(f: EnsimeFile) extends FileWatcherMessage
```
fommil: these look like they should be `RawFile`.

Remember, `EnsimeFile` indicates that it is a file or a path within an archive. So we want to be as specific as possible where we can both in terms of what the inputs for a method are but also for the outputs. This helps us know what we actually need to implement.

jozi-k:
I am struggling a bit with how to handle correctly `FileSystem`. For instance, I need the `path` inside jar (which creates fs) and also working with iterator (I cannot close fs then), depicking class files (which tries to create new fs), etc. My last commit is just a bunch of ideas I tried, but none of it worked. I would be very happy to continue, but I will need some hints to move on.

fommil:
hmm, I think I'll need to check out the branch and have a deep dive before I can comment on anything.

The one thing I know from memory is that I thought the `EnsimeFile` implicits wrapped all of this so you didn't have to think about it.

fommil:
I'm at Scala World next week, I'd love to take some time and look at this. We'll see if I find the time or not... believe me I dream about **finally** removing VFS :grinning:

fommil:
can you please re-raise this against the 3.0 branch? You shouldn't have any conflicts

Issue #1753, opened 2017-04-18
alternative find usages query using only the indexer


fommil:
the current impl of find usages (with the graph) loads all files into the PC. It would be good to have a variant that just returns the files and lines for displaying in the client like a grep query.

fommil:
(some discussion in linked / closed tickets)

VC1995:
@fommil Hey, I was out of scala work so wanted to pick this up. Just needed to clear a few things :)

Currently we use the `GraphService` to get an Iterable of `FqnSymbol`. We then use the filename(uri) information to load the file in compiler and then later, using `usesOfSym(pos, files)` we return `ERangePosition`s which have `filename` and also the `offset`. Am I right?

If we already have the `source` and `line`, why are we loading the files in the compiler and then using `usesOfSym(pos, files)` ?

Now, do we simply want to return the `source` and `line` to the user t?(which I think is actually enough to open and scroll to that location in the editor as we can convert (row,col=1) to offset after we have loaded file in the editor)

fommil:
yes exactly. We just want to return a list of source and lines to the client.

VC1995:
Okay :+1:

PR #1830, opened 2017-07-08
[Sponsored] Find Usages and Implementing Classes

VC1995:
#1753 
Does this look like what we want?

### DIFF
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -137,21 +137,29 @@ final case class ImportSuggestionsReq(
   maxResults: Int
 ) extends RpcSearchRequest
 
+final case class FqnOfSymbolAtPointReq(file: SourceFileInfo, point: Int) extends RpcAnalyserRequest
+
+final case class FqnOfTypeAtPointReq(file: SourceFileInfo, point: Int) extends RpcAnalyserRequest
+
 /**
  * Responds with `SourcePositions`.
  */
 final case class UsesOfSymbolAtPointReq(
-  file: Either[File, SourceFileInfo],
+  file: SourceFileInfo,
   point: Int
-) extends RpcAnalyserRequest // will probably become a search request
+) extends RpcRequest
 
 /**
- * Responds with `SymbolTreeInfo`
+ * Responds with `SymbolHierarchyInfo`
  */
-final case class TreeOfSymbolAtPointReq(
-  file: Either[File, SourceFileInfo],
+final case class HierarchyOfTypeAtPointReq(
+  file: SourceFileInfo,
   point: Int
-) extends RpcAnalyserRequest
+) extends RpcRequest
+
+final case class FindUsages(fqn: String) extends RpcSearchRequest
+
+final case class FindHierarchy(fqn: String) extends RpcSearchRequest
 
 /**
  * Responds with a `StringResponse` for the URL of the documentation if valid,
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -138,13 +138,20 @@ final case class ImportSuggestionsReq(
 ) extends RpcSearchRequest
 
 /**
- * Responds with `ERangePositions`.
+ * Responds with `SourcePositions`.
  */
 final case class UsesOfSymbolAtPointReq(
   file: Either[File, SourceFileInfo],
   point: Int
 ) extends RpcAnalyserRequest // will probably become a search request
 
+/**
+ * Responds with `HierarchyInfo`
+ */
+/*final case class HierarchyOfSymbolAtPointReq(
+  file: Either[File, SourceFileInfo],
+  point: Int
+) extends RpcAnalyserRequest*/
 /**
  * Responds with a `StringResponse` for the URL of the documentation if valid,
  * or `FalseResponse`.
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -146,12 +146,13 @@ final case class UsesOfSymbolAtPointReq(
 ) extends RpcAnalyserRequest // will probably become a search request
 
 /**
- * Responds with `HierarchyInfo`
+ * Responds with `SymbolTreeInfo`
  */
-/*final case class HierarchyOfSymbolAtPointReq(
+final case class TreeOfSymbolAtPointReq(
   file: Either[File, SourceFileInfo],
   point: Int
-) extends RpcAnalyserRequest*/
+) extends RpcAnalyserRequest
+
 /**
  * Responds with a `StringResponse` for the URL of the documentation if valid,
  * or `FalseResponse`.
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -137,8 +137,14 @@ final case class ImportSuggestionsReq(
   maxResults: Int
 ) extends RpcSearchRequest
 
+/**
+ * Responds with `FullyQualifiedName`
+ */
 final case class FqnOfSymbolAtPointReq(file: SourceFileInfo, point: Int) extends RpcAnalyserRequest
 
+/**
+ * Responds with `FullyQualifiedName`
+ */
 final case class FqnOfTypeAtPointReq(file: SourceFileInfo, point: Int) extends RpcAnalyserRequest
 
 /**
@@ -150,15 +156,21 @@ final case class UsesOfSymbolAtPointReq(
 ) extends RpcRequest
 
 /**
- * Responds with `SymbolHierarchyInfo`
+ * Responds with `HierarchyInfo`
  */
 final case class HierarchyOfTypeAtPointReq(
   file: SourceFileInfo,
   point: Int
 ) extends RpcRequest
 
+/**
+ * Responds with `SourcePositions`.
+ */
 final case class FindUsages(fqn: String) extends RpcSearchRequest
 
+/**
+ * Responds with `HierarchyInfo`
+ */
 final case class FindHierarchy(fqn: String) extends RpcSearchRequest
 
 /**
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -237,11 +237,10 @@ final case class LineSourcePosition(file: EnsimeFile, line: Int) extends SourceP
 
 case class SourcePositions(positions: List[SourcePosition]) extends RpcResponse
 
-sealed trait HierarchyInfo
-final case class ClassInfo(fqn: String, sourcePosition: Option[SourcePosition]) extends HierarchyInfo
-final case class TreeInfo(classInfo: ClassInfo, hierarchy: Seq[HierarchyInfo]) extends HierarchyInfo
+// See if `TypeInfo` can be used instead
+final case class ClassInfo(fqn: String, sourcePosition: Option[SourcePosition])
 
-final case class SymbolTreeInfo(ancestors: Option[HierarchyInfo], inheritors: Option[HierarchyInfo]) extends RpcResponse
+final case class HierarchyInfo(ancestors: List[ClassInfo], inheritors: List[ClassInfo]) extends RpcResponse
 
 final case class PackageInfo(
     name: String,
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -237,6 +237,9 @@ final case class LineSourcePosition(file: EnsimeFile, line: Int) extends SourceP
 
 case class SourcePositions(positions: List[SourcePosition]) extends RpcResponse
 
+// think about what should these be, or if something could be reused here
+// final case class HierarchyInfo(ancestors: Any, inheritors: Any) extends RpcResponse
+
 final case class PackageInfo(
     name: String,
     fullName: String,
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -237,8 +237,11 @@ final case class LineSourcePosition(file: EnsimeFile, line: Int) extends SourceP
 
 case class SourcePositions(positions: List[SourcePosition]) extends RpcResponse
 
-// think about what should these be, or if something could be reused here
-// final case class HierarchyInfo(ancestors: Any, inheritors: Any) extends RpcResponse
+sealed trait HierarchyInfo
+final case class ClassInfo(fqn: String, sourcePosition: Option[SourcePosition]) extends HierarchyInfo
+final case class TreeInfo(classInfo: ClassInfo, hierarchy: Seq[HierarchyInfo]) extends HierarchyInfo
+
+final case class SymbolTreeInfo(ancestors: Option[HierarchyInfo], inheritors: Option[HierarchyInfo]) extends RpcResponse
 
 final case class PackageInfo(
     name: String,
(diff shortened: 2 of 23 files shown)


fommil:
file should be an EnsimeFile, Symbol Location looks good. Maybe best just to break the name of positions and call it something else.

Can't we use SourceFileInfo? That'd be cleaner.

fommil:
Perhaps we want uses of symbol at point AND type at point, two new calls. The type variant would have to ignore type parameters.

VC1995:
We may use `LineSourcePosition`?
`SourciFileInfo` doesn't have line information

fommil:
yeah `LineSourcePosition` is perfect :smile:

VC1995:
@fommil Could you guide me a liitle on how can I add an optional line number with the UsedIn edge

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -145,17 +145,28 @@ trait RichCompilerControl extends CompilerControl with RefactoringControl with C
   def askReloadAndTypeFiles(files: Iterable[SourceFile]) =
     askOption(reloadAndTypeFiles(files))
 
-  def askUsesOfSymAtPos(pos: Position)(implicit ec: ExecutionContext): Future[List[RangePosition]] = {
+  def askUsesOfSymAtPos(pos: Position)(implicit ec: ExecutionContext): Future[List[LineSourcePosition]] = {
```
fommil: we want to move this query into the `SearchService`, shouldn't need the presentation compiler at all

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -66,8 +66,8 @@ final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
       val interfaceNames: List[ClassName] = interfaces.map(ClassName.fromInternal)(collection.breakOut)
       val superClass = Option(superName).map(ClassName.fromInternal)
 
-      internalRefs.addAll(interfaceNames.asJava)
-      internalRefs.addAll(superClass.toList.asJava)
+      internalRefs.addAll(interfaceNames.map(FullyQualifiedReference(_, None)).asJava)
```
fommil: I think I need to sit down and read this file again when I'm not tired, sorry.

core/src/main/scala/org/ensime/indexer/domain.scala
```diff
@@ -22,6 +22,8 @@ object Access {
     else Default
 }
 
+final case class FullyQualifiedReference(fqn: FullyQualifiedName, line: Option[Int])
```
fommil: ok, this is probably a good thing. Is it always obvious which source file it refers to?

fommil: btw, you need to bump the schema version number in `SearchService`

VC1995: Yes, okay

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
VC1995: Yes, sure. I wanted to show the changes in `ClassFileIndexer` mainly. Hadn't sorted out stuff here yet.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -227,13 +226,28 @@ class Analyzer(
...
-      import context.dispatcher
-      val response = if (toSourceFileInfo(file).exists()) {
+      if (toSourceFileInfo(file).exists()) {
+        val p = pos(file, point)
+        scalaCompiler.askLoadedTyped(p.source)
+        scalaCompiler.askSymbolFqn(p) match {
+          case Some(fqn) =>
+            indexer forward FindUsages(fqn.fqnString)
+          case None =>
+            sender ! FalseResponse
```
fommil: this seems to be repeating a lot of code... can't we just treat this like a two part query as we do for the Doc requests? i.e. higher up sends a TypeAtPoint and then reroutes it to the indexer. Means the analyzer doesn't need to know that these requests exist.

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
```diff
@@ -145,21 +145,6 @@ trait RichCompilerControl extends CompilerControl with RefactoringControl with C
   def askReloadAndTypeFiles(files: Iterable[SourceFile]) =
     askOption(reloadAndTypeFiles(files))
 
-  def askUsesOfSymAtPos(pos: Position)(implicit ec: ExecutionContext): Future[List[RangePosition]] = {
```
fommil: I'm wondering if perhaps we should still augment the indexer-only results with the hits from the currently open source files. Part of me prefers the new cleaner "indexer only" impl, but the other half knows that somebody is going to expect it to show hits in their current file.

fommil: (we can maybe do that as a phase 2 refinement)

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
fommil: maybe `interfaceNames` should always have been a Java collection in the first place, to avoid the conversion.

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -101,7 +101,7 @@ final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
       super.visitField(access, name, desc, signature, value)
       new FieldVisitor(ASM5) with ReferenceInFieldHunter {
         override def visitEnd(): Unit = {
-          internalRefs.add(ClassName.fromDescriptor(desc))
+          internalRefs.add(FullyQualifiedReference(ClassName.fromDescriptor(desc), None))
```
fommil: why are there so many `None` uses here? Surely we know the most recently seen line number?

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
```diff
@@ -157,7 +158,7 @@ final class ClassfileIndexer(file: FileObject) extends SLF4JLogging {
 
     var clazz: RawClassfile
 
-    var internalRefs = new util.HashSet[FullyQualifiedName]()
+    var internalRefs = new util.HashSet[FullyQualifiedReference]()
```
fommil: `HashSet` could be something faster, like a linked list. It is unlikely that we'll have dupes anymore, and the few that we do have we can filter later.

core/src/main/scala/org/ensime/indexer/graph/GraphService.scala
```diff
@@ -251,6 +253,7 @@ class GraphService(dir: File) extends SLF4JLogging {
 
         case MethodSymbolInfo(_, source, refs, bs, scalap) =>
           val owner = classes(bs.name.owner.fqnString)
+          //          if (s.fqn.contains("testMethod")) println(s"${s.fqn} at ${bs.line}")
```
fommil: comments

core/src/main/scala/org/ensime/indexer/orientdb/SimpleOrient.scala
```diff
@@ -372,7 +373,7 @@ package object syntax {
       bdf: BigDataFormat[FqnSymbol],
       oid: OrientIdFormat[FqnSymbol, P],
       p: SPrimitive[P]
-    ): Seq[VertexT[FqnSymbol]] = {
+    ): Seq[(VertexT[FqnSymbol], VertexT[LineNumber])] = {
```
fommil: I don't like that this returns a Seq of Tuples. It should return a `Seq` of some domain object.

fommil: this also seems to be doing two jumps in one go.

It looks like the generalisability of this file went out the window a long time ago, it was never supposed to know anything about the domain object but it seems to be hardcoding `FqnSymbol` and `UsedIn`.

protocol-swanky/src/main/scala/org/ensime/server/protocol/swank/SwankFormats.scala
```diff
@@ -79,6 +79,14 @@ object SwankProtocolCommon {
     }
   }
 
+  implicit object HierarchyInfoFormat extends SexpFormat[HierarchyInfo] {
+    def write(hi: HierarchyInfo): Sexp = hi match {
+      case value: ClassInfo => ???
```
fommil: :+1: obvs still TODO

api/src/main/scala/org/ensime/api/incoming.scala
```diff
@@ -138,14 +138,22 @@ final case class ImportSuggestionsReq(
...
   point: Int
 ) extends RpcAnalyserRequest // will probably become a search request
 
 /**
+ * Responds with `SymbolTreeInfo`
+ */
+final case class TreeOfSymbolAtPointReq(
+  file: Either[File, SourceFileInfo],
+  point: Int
+) extends RpcAnalyserRequest
```
VC1995: That would be `FindHierarchy` ?

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
VC1995: I don't completely understand. We wouldn't  get anything extra using this, would we?
Since now we store `FullyQualifiedReference` references on different lines are different and we should get all of them.

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
VC1995: Yes sure.

protocol-swanky/src/main/scala/org/ensime/server/protocol/swank/SwankFormats.scala
VC1995: Yes, I wasn't sure if those were the right way to do it in the first place so kept it for later.

core/src/main/scala/org/ensime/core/Analyzer.scala
```diff
@@ -227,13 +226,28 @@ class Analyzer(
...
+      } else sender ! EnsimeServerError(s"File does not exist: ${file.file}")
+    case TreeOfSymbolAtPointReq(file, point) =>
+      if (toSourceFileInfo(file).exists()) {
         val p = pos(file, point)
-        val uses = scalaCompiler.askUsesOfSymAtPos(p)
-        uses.map(positions => ERangePositions(positions.map(ERangePositionHelper.fromRangePosition)))
-      } else Future.successful(EnsimeServerError(s"File does not exist: ${file.file}"))
-      pipe(response) to sender
+        scalaCompiler.askLoadedTyped(p.source)
+        scalaCompiler.askSymbolFqn(p) match {
```
VC1995: I'm sorry I don't know if it is. Maybe you could help me understand :)

core/src/main/scala/org/ensime/core/Analyzer.scala
VC1995: The `Project` actor should ask `Analyzer` the type at point and getting the fqn from the type forward the request to `Indexer`, is that what we want to do?

fommil: See how we do it for the docurl requests

api/src/main/scala/org/ensime/api/incoming.scala
fommil: It's not exposed

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
VC1995: `visitLine` is only on the `MethodVisitor`

fommil: are you sure we don't get the line numbers of fields somehow?

core/src/main/scala/org/ensime/indexer/ClassfileIndexer.scala
VC1995: Why do we want `interfaceNames` to be a Java collection?

fommil: This would be only if it is used internally to the classfile indexer.

Since we have to use mutable state when using the Visitor pattern (which is what the ASM library is doing), we may as well use the most efficient mutable collections. Scala's mutable collections have terrible performance.

fommil:
github is doing that weird thing again where I can see you've pushed commits but I can't see them. I think they must do location based sharding / caching.

fommil:
btw, I looked in the code again and multi-phase requests are done in `RequestHandler`. e.g. `DocUriAtPointReq`. This is isolated to a request and will not cause the `Project` actor or anything else to become unavailable.

fommil:
probably there are quite a few requests that could be refactored to be like `DocUriAtPointReq` instead of complicating `Analyzer` any further.

api/src/main/scala/org/ensime/api/outgoing.scala
```diff
@@ -235,6 +235,14 @@ final case class EmptySourcePosition() extends SourcePosition
 final case class OffsetSourcePosition(file: EnsimeFile, offset: Int) extends SourcePosition
 final case class LineSourcePosition(file: EnsimeFile, line: Int) extends SourcePosition
 
+case class SourcePositions(positions: List[SourcePosition]) extends RpcResponse
+
+sealed trait HierarchyInfo
+final case class ClassInfo(fqn: String, sourcePosition: Option[SourcePosition]) extends HierarchyInfo
+final case class TreeInfo(classInfo: ClassInfo, hierarchy: Seq[HierarchyInfo]) extends HierarchyInfo
+
+final case class SymbolTreeInfo(ancestors: Option[HierarchyInfo], inheritors: Option[HierarchyInfo]) extends RpcResponse
```
VC1995: To create `TypeInfo` from `ClassDef` :

- [x] name (optional)
- [x] declAs
- [x] fullName
- [ ] typeArgs
- [ ] members
- [x] pos (optional)
- [ ] typeParams

So we can set others as `Nil` and pass `TypeInfo` instead of `ClassInfo`

fommil: I think the members are available in the graph. I'm surprised that the typeargs and params are not in the graph.

core/src/main/scala/org/ensime/core/RichPresentationCompiler.scala
fommil: consider the case where the user has added a reference in their currentyl open file, but not yet compiled that file.

VC1995: Ah, yes.
Though augmenting the results is going to require some more work.
Since now we first query the analyzer to obtain the fqn and the reroute to indexer so that it can find it's usages, should we also pass along the results we have from the loaded files?

fommil:
almost there. A few more things to do in addition to outstanding comments above:

- bump the index/db version string
- bump the protocol version
- make sure deletion is deleting these new vertices

fommil:
```
[error] Failed tests:
[error] 	org.ensime.server.protocol.swank.SwankFormatsSpec
[error] 	org.ensime.swanky.SwankyFormatsSpec
```

fommil:
ok, this is because of file stuff. We have an evil evil hack to deal with this. Because you've used `"/abc/def"` by hand, it means that things don't work as expected. This must use the `s"$file"` string interpolator which (omg, it's terrible) does some nasty things to hijack how `s""` works for `File`, see `EscapingStringInterpolation`

fommil:
you know what, screw Travis... MERGE-APOCALYPSE!

PR #1831, opened 2017-07-14
Introduce a test explicitly checking your dev environment

raboof:
So I won't have to search the web to find out about
https://github.com/ensime/ensime-server/issues/1650


fommil:
it's a bit weird to do it this way. I thought you'd put a check into the build definition itself.

fommil:
also, it's not necessary that `JAVA_HOME` is defined... the important thing is that the current java has a `src.zip` in the expected location.

PR #1832, opened 2017-07-16
canon spec no longer depends on $JAVA_HOME or JDK locations:

fanatoly:
e.g. the spec should now consistently pass on OS X

### DIFF
--- a/core/src/test/scala/org/ensime/core/CanonSpec.scala
+++ b/core/src/test/scala/org/ensime/core/CanonSpec.scala
@@ -2,10 +2,9 @@
 // License: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.core
 
-import java.io.File
+import java.io.{ File, OutputStreamWriter }
 import java.nio.file._
-
-import scala.util.Properties.jdkHome
+import java.util.zip.{ ZipEntry, ZipOutputStream }
 
 import org.ensime.util.file._
 import org.ensime.util.path._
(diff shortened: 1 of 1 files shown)


fommil:
Hmm, thank you for taking the time to contribute to ensime but I don't think you understand why this test is here. The `src.zip` is *required* by other tests.

fommil:
what is it you're trying to achieve, perhaps I can explain or help you?

fanatoly:
Thanks! I wanted to debug an issue I was having with a specific .ensime
file and the current version of ensime 2.0 but I've rabbit holed into
trying to get the test suite to pass locally for me, on OS X. These src.zip
tests seem like the only tests that are stopping that from happening(I'm
not addressing FileWatcher stuff). I was planning to submit a separate PR
removing the dependence on jdkHome from EnsimeFileSpec.scala.

Please, let me know if I'm way off base here.

fommil:
the src.zip is necessary for us to be able to assert on source lookups to java files in other parts of the code. It's also necessary for the `jdk-home` to be in the `.ensime` and for it to have the standard jdk layout. Also, yes it needs to be a jdk not a jre.

Make sure you installed java / sbt correctly as per http://ensime.org/learning_scala/#install and I advise using jenv, since you're on OS X.

Before running the tests, make sure to follow the advise at http://ensime.org/contributing/#compiling-and-tests specifically the bit about following `.drone.yml`

fanatoly:
Am I correct in my understanding that the other places that rely on src.zip
use the location provided by the sbt plugin(
https://github.com/ensime/ensime-sbt/blob/2.0/src/main/scala/EnsimePlugin.scala#L600)?
If so, it might make sense to use the same code to derive jdkHome in the
tests(would be glad to send over a patch).

I'm also missing which part of .drone.yaml sets (JDK|JAVA)_HOME. I would
appreciate a more direct pointer there. It also doesn't look like I'm the
first to land here(https://github.com/ensime/ensime-server/pull/1831).
Would love to find some way to improve the situation.

Again, thanks for taking the time to explain.

fommil:
jenv sets JAVA_HOME, just install jenv.

PR #1837, opened 2017-07-22
canary tests for broken java lambdas

fommil:
#1764


fommil:
Now that's interesting

fommil:
Windows and OSX are behaving the same way: `StackOverflowError` on the first one and no exception on the second.

Issue #1785, opened 2017-06-01
exclusion list for completions


fommil:
very much the dual of #354 and mentioned in https://github.com/ensime/ensime-emacs/pull/623 by @yangwen0228

yangwen0228:
So, as you mentioned, where to do this? 😃

fommil:
The best place to start is the contributing page on the website. There are two videos there that explain the architecture of the server.

Then have a look for Analyzer and beneath that *Completion*. You should be able to find where to apply a regex filter easily enough. We can apply the $ exclusion for now but when the config PR is ready we can allow users to provide their own. I'm not sure if it applies to FQN symbols or relative.

yangwen0228:
Ok, thank you! Then, let me watch the videos first...

PR #1842, opened 2017-07-29
#1785: exclusion list for completions containing "$"

yangwen0228:
This change can exclude the type completions containing "$".

### DIFF
--- a/core/src/main/scala/org/ensime/core/Indexer.scala
+++ b/core/src/main/scala/org/ensime/core/Indexer.scala
@@ -31,7 +31,7 @@ class Indexer(
 
   def oldSearchTypes(query: String, max: Int): List[TypeSearchResult] = {
     // Remove $/$class from the end.
-    def strip(fqn: String): String = fqn.replaceAll("\\$(class)*$", "")
+    def strip(fqn: String): String = fqn.replaceAll("\\$.*$", "")
 
     import org.ensime.util.list._
 


fommil:
This seems like it might have some unexpected consequences, can you please add some extra tests to show why it might be good?

yangwen0228:
Sure! It can exclude 'CollectionMapper$ListSort' ...etc. Only keep the 'CollectionMapper'. I haven't found any unexpected sequences so far.

fommil:
Can you please add some automated tests to SearchServiceSpec? If it isn't tested, somebody will break it by accident....

fommil:
Would we not want to import the ListSort in that example?

fommil:
Btw in scala it's quite common to want to import inner classes, so I am concerned this will introduce a regression. The real fix, of course, is to present the Scala name (not the fqn) in scala import choices... but for now we might want to live with this and workaround it on the client side.

yangwen0228:
In Java, we use dot to represent inner class. And the $ is used to represent inner class when compile. So we avoid to use $ in practice. If we use $ a lot in Scala, maybe I should drop this commit, and just implement it in client side.

fommil:
yes, I think that would be the correct way to do it (for now). Or, you could pass the information about if it's a scala/java file down to this method and then have different filters based on language?

fommil:
or maybe client side should convert `$` into `.` (it's the same in scala... the `$` is only the bytecode name, not the language name).

yangwen0228:
@fommil Thank you so much! 😊 I think in most cases we don't need the inner classes. And the string-prefix matching strategy is enough and more convenient. So I will give up this commit, and use the client implementation instead.




Emacs implementation is like this:
```
(defun ensime-get-completions-async
      (max-results case-sense callback)
    (ensime-rpc-async-completions-at-point
     max-results case-sense
     (lexical-let ((continuation callback))
       (lambda (info)
         (let* (name
                (prefix (ensime-completion-prefix-at-point))
                (candidates (remove-if
                             '(lambda (candidate)
                                (setq name (getf candidate :name))
                                (or (string-match "\\$" name)
                                    (not (string-prefix-p prefix name))))
                             (plist-get info :completions)))
                (candidates (ensime--annotate-completions candidates)))
           (funcall continuation candidates))))))
```

Issue #1843, opened 2017-07-29
bugfixes for reverse lookup


fommil:
- exclude noise of self references (e.g. find usages of `SearchService` should not include usages within `SearchService`'s methods)
- include one line summary
- should work for methods

ktonga:
Hey guys, any chance to include the column of the usage along with the line?

fommil:
unfortunately not, it's not in the class files. See https://github.com/scalacenter/advisoryboard/pull/21 for some info. And https://github.com/lampepfl/dotty/issues/1523#issuecomment-255862771 and several discussions with the scalameta folks about putting this in the semanticdb.

ktonga:
Ok, thank you for the references.

VC1995:
Should we close this?

ktonga:
Last time I tried it wasn't working for methods. Should it be working already?

fommil:
Probably what you've done is looked for the uses of a method in an implementation of a trait or something... it will only find usages of that exact FQN. Look for usages of the trait method.

PR #1848, opened 2017-07-31
[Sponsored] Bug fixes for reverse lookup 

VC1995:
#1843

- [x] exclude noise of self references
- [x] include one line summary


VC1995:
@fommil What do you think?
And should we ignore the results with no line numbers? Are those useful?

api/src/main/scala/org/ensime/api/outgoing.scala
```diff
@@ -235,10 +235,8 @@ final case class EmptySourcePosition() extends SourcePosition
 final case class OffsetSourcePosition(file: EnsimeFile, offset: Int) extends SourcePosition
 final case class LineSourcePosition(file: EnsimeFile, line: Int) extends SourcePosition
 
-final case class SourcePositions(positions: List[SourcePosition]) extends RpcResponse
-
-/*case class UsageLocation(pos: SourcePosition, preview: Option[String])
-case class UsageLocations(locations: List[UsageLocation]) extends RpcResponse*/
+case class SourceHint(position: SourcePosition, preview: Option[String])
+case class SourceHints(sourceHints: List[SourceHint]) extends RpcResponse
```
fommil: It would be good if the protocol was backwards compatible.

fommil:
Yeah we should keep results with no numbers/source because we might do decompilation in the client.

Btw I wonder if perhaps we should limit results to 1,000 or something, or ban lookups on some key classes like String.

VC1995:
Currently I just did `.take(1000)`. But do you think we should add `maxResults: Option[Int]` to the `UsesOfSymbolAtPointReq` ?

api/src/main/scala/org/ensime/api/outgoing.scala
```diff
@@ -235,8 +235,7 @@ final case class EmptySourcePosition() extends SourcePosition
 final case class OffsetSourcePosition(file: EnsimeFile, offset: Int) extends SourcePosition
 final case class LineSourcePosition(file: EnsimeFile, line: Int) extends SourcePosition
 
-case class SourceHint(position: SourcePosition, preview: Option[String])
-case class SourceHints(sourceHints: List[SourceHint]) extends RpcResponse
+final case class SourcePositions(positions: List[SourcePosition], previews: List[String]) extends RpcResponse
```
fommil: you could just leave the field called `positions` and change the type. This is a bit messy.

fommil:
we should probably make this max a part of the EnsimeServerConfig with a default value in `application.conf` but let's leave it at 1000.

VC1995:
How do I do that?

fommil:
if you want to, it's not hard :smile:... add a new field to `EnsimeServerConfig` and then fix all the compile errors that it implies.

Then go into the `core/src/resources/application.conf` and add your new setting there up at the top (should be obvious).

fommil:
only problem is you might need to add it as a dependency to create an `Indexer` or `SearchService` or something

VC1995:
It'll be great if you can help with that :sweat_smile: 
I was trying this out on ensime-sublime and seems to be working fine :smile: 
Do the changes look okay?

fommil:
yeah, but you changed the protocol so you need to bump the protocol version. Maybe @bomgar will (pretty please with cherry on top!) update the emacs side? We also need to assert on the new server protocol string.

fommil:
I'd like to have the Emacs fix ready to release alongside this... I presume you have a sublime one sitting waiting?

VC1995:
Yep. :smile:

VC1995:
@fommil Don't merge just yet. I'm still getting duplicates (was trying on sublime). I returned distinct vertices but they might as well contain same info. I need to add `distinct` when I convert them to `LineSourcePositions`.

fommil:
We have the Emacs patch ready so I'll merge everything in the morning :-)

core/src/main/scala/org/ensime/core/Indexer.scala
```diff
@@ -73,16 +73,26 @@ class Indexer(
...
-              u.line.getOrElse(u.line.getOrElse(0))
+              u.line.getOrElse(0)
             )
           }
         }(collection.breakOut)
-        SourcePositions(results)
+        val files = positions.foldLeft[Set[EnsimeFile]](Set.empty) { (files, pos) =>
+          if (pos.line > 0) files + pos.file else files
+        }
+        val contents: Map[EnsimeFile, Array[String]] = files.map(f => f -> f.readAllLines.toArray)(collection.breakOut)
```
fommil: oh, I missed the `toArray` here, it is almost certainly slower to create an `Array` and lookup than the cost to lookup the line directly on the `List`.

fommil: or maybe the underlying `path.scala` impl should return a `Vector

VC1995: Since a file may contain multiple references, looking up an index multiple times in an array would be faster compared to a list. (constant and linear). That's why I did that. Do you think we  would have better performance with keeping it as a list?
Though ofcourse a file may only have a single reference where converting to an array would have gone waste.

ktonga:
@fommil WDYM?

>Yeah we should keep results with no numbers/source because we might do decompilation in the client.

fommil:
@ktonga I was thinking if there are results that don't have source info, it probably means the sources weren't attached. Then you can tell the client to load the relevant `.class` file, which can have a mode for piping it through a decompiler. But on looking at this again, I think we're only talking about source files and the information is just not available in the `.class` file about where it is defined in the source. I guess we must be filtering reverse lookups that don't have sources.

ktonga:
Gotcha. Thanks.

PR #1859, opened 2017-08-22
Prevent dealiasing when constructing type info for type aliases

### LINKED ISSUE #1858: Wrong response for type at point on values of type aliases
Given the following (already with clues for the tests):
```
object TypeAliases {
  trait State[S, A]
  type IntState[A] = State[Int, A]
  val intInt@intintstate@State: IntState[Int] = ???
  type IntList = List[Int]
  val int@intlist@List: IntList = ???
}
```

The returned name for `intintstate` would be `State[Int]` and `List` for `intlist`. This happens because `tpe.typeSymbol` is used for constructing the name, which normalizes the typeSymbol and dealiases it. 

This has been annoying me, so PR with a fix coming up.

iravid:
See commit message for details.

Resolves #1858.

### DIFF
--- a/core/src/it/scala/org/ensime/core/TypeToScalaNameSpec.scala
+++ b/core/src/it/scala/org/ensime/core/TypeToScalaNameSpec.scala
@@ -251,4 +251,69 @@ class TypeToScalaNameSpec extends EnsimeSpec
         }
       }
   }
+
+  it should "not dealias user-controlled type aliases" in withPresCompiler { (config, cc) =>
+    runForPositionInCompiledSource(
+      config, cc,
+      "package com.example",
+      "object TypeAliases {",
+      "  trait State[S, A]",
+      "  type IntState[A] = State[Int, A]",
+      "  val intInt@intintstate@State: IntState[Int] = ???",
+      "  type IntList = List[Int]",
+      "  val int@intlist@List: IntList = ???",
+      "  type IntMap[A] = Map[Int, A]",
+      "  val int@intmap@map: IntMap[Int] = ???",
+      "}"
+    ) { (p, label, cc) =>
+        withClue(label) {
+          val typeResult = cc.askTypeInfoAt(p).getOrElse(fail)
+          typeResult shouldEqual {
+            label match {
+              case "intintstate" =>
+                BasicTypeInfo("IntState[Int]", DeclaredAs.Nil, "com.example.TypeAliases.IntState[scala.Int]")
+              case "intlist" =>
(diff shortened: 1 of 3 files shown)


iravid:
I toyed around with excluding anything under `scala.Predef` from this change. It's definitely possible. Wondering if it's more correct.

fommil:
If we perhaps have a rule that we dealias anything matching `scala.*` or `.*Predef.*` (don't forget about the user's custom predefs) that would be a pretty good balance?

iravid:
Sounds good to me. I'll make that change.

Perhaps instead of `.*Predef.*` we can fetch the custom predef from the compiler? Does that exist?

fommil:
not really, there is a patch upstream but everybody who does this disables the system one and manually imports their own `Predef` (keeping the naming convention).

iravid:
@fommil what do you think should be done with values of type `type IntMap[A] = Map[Int, A]`? I'm currently going for full dealiasing: display `Map[Int, A]`

fommil:
I'd go for the rule of whatever the type alias's FQN is, rather than what it points at.

If somebody does `type Wibble = String` I'd want to see `Wibble`.

iravid:
That's a great idea. Thanks.

iravid:
Currently still breaking because the dealiasing is dropping some of the type parameters. Will continue tomorrow :-)

iravid:
The AppVeyor failure seems unrelated. @fommil could you restart it?

Other than that, ready for review :-)

fommil:
rebase needed

iravid:
Rebased, thanks!

api/src/main/scala/org/ensime/api/common.scala
```diff
@@ -23,9 +23,10 @@ object DeclaredAs {
   case object Object extends DeclaredAs('object)
   case object Class extends DeclaredAs('class)
   case object Field extends DeclaredAs('field)
+  case object TypeAlias extends DeclaredAs('typeAlias)
```
fommil: this needs a bump in the protocol version. I'm not sure we want to expose type aliases at the api level.

iravid: Aha, you're right. I don't think this is critical either TBH. It just seemed strange that the `DeclaredAs` value would be `Nil`.

What do you say? Shall we drop it?

fommil: yeah, drop it. This all has to do with values retrieved from the graph.

iravid: Done

PR #1861, opened 2017-08-23
Widen single types when formatting them for display

### LINKED ISSUE #1860: Type at point shows wrong result when checking the type of the qualifier in a method selection
```
object Test {
  val set: Set[(String, String)] = ???
  s@set@et.toMap
}
```

Returns `Set`. It appears that the type of the tree is `set.type`, which is why the type parameters are missing.

iravid:
Next up in the "scratch your itch" series...

Given the expression `set.toMap`, the tree for the qualifier is of type
`set.type`. This causes the name formatting to drop the type parameters.

This commit widens single types before formatting them for display.

Also added is coverage for formatting of constant types.

Resolves #1860.


fommil:
ignore appveyor. it gets like this sometimes.

PR #1867, opened 2017-09-01
Add dealiasing fixes when formatting tuples and function types

### LINKED ISSUE #1866: Type aliases for tuples and function types are dealiased when displayed
Given this:
```scala
type Env[E, A] = (E, A)
val env: Env[String, Int] = ???

type IntReader[A] = Int => A
val intreader: IntReader[String] = ???
```

`env` shows `(String, Int)` and `intreader` shows `() => String`. Again a dealiasing issue.

iravid:
This commit fixes two issues:
- when formatting type aliases pointing to tuples, dealiasing is
  avoided by using typeSymbolDirect when testing if the type's full
  name starts with `Tuple`.
- when formatting type aliases pointing to function types, dealiasing
  is similarly avoided in isArrowType.

Resolves #1866.

### DIFF
--- a/core/src/it/scala/org/ensime/core/TypeToScalaNameSpec.scala
+++ b/core/src/it/scala/org/ensime/core/TypeToScalaNameSpec.scala
@@ -264,6 +264,24 @@ class TypeToScalaNameSpec extends EnsimeSpec
       "  val int@intlist@List: IntList = ???",
       "  type IntMap[A] = Map[Int, A]",
       "  val int@intmap@map: IntMap[Int] = ???",
+      "  type Env[E, A] = (E, A)",
+      "  val en@env@v: Env[String, Int] = ???",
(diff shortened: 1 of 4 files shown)


iravid:
At this point, btw, I'm pretty certain that `TypeToScalaName` should be folded into `ModelBuilders`. The fact that they are separate modules makes us repeat a lot of work around using the right type symbols, etc. I'll try and take a stab at that later this month.

fommil:
I'd prefer to move it the other way... `ModelBuilders` is a travesty and that it has functions with the same names as the things it is building is purely a legacy thing.

fommil:
this is really nice stuff, btw

iravid:
Ah you're definitely right. `BasicTypeInfo` should just construct the data type and shouldn't be so damn smart :-)

iravid:
This is good for merging, right?

Issue #1553, opened 2016-07-09
async connection info


fommil:
as discussed in https://github.com/ensime/ensime-server/pull/1552 the first message that clients receive when connecting should be an _async_ connection info, with information about the protocol etc.

This is breaking, because right now it's a sync message in response to a request.

We can then remove the connection info request message.

This simplifies the client side logic.

fommil:
We can keep the sync message for backwards compat and deprecate it, for removal later.

jan0sch:
Is anyone working on this? I'm tempted to give it a shot when there is time.

Would this just involve removing the stashing and waiting stuff from the server?

fommil:
you can't break the current protocols, but you could add an async message that the clients could move to. Then we can remove the old request/response message in 3.0

jan0sch:
Ok, I'll try to give it a shot.

PR #1868, opened 2017-09-13
#1553 async connection info

jan0sch:
* add AsynConnectionInfoReq
* deprecate ConnectionInfoReq
* handle both messages the same way for now

### DIFF
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -19,8 +19,14 @@ sealed trait RpcRequest
 sealed trait RpcStartupRequest extends RpcRequest
 
 /**
+ * Respond with a `ConnectionInfo`.
+ */
+case object AsynConnectionInfoReq extends RpcStartupRequest
+
+/**
  * Responds with a `ConnectionInfo`.
  */
+@deprecating("prefer AsynConnectionInfoReq")
 case object ConnectionInfoReq extends RpcStartupRequest
 
 // related to managing the state of the analyser
--- a/api/src/main/scala/org/ensime/api/incoming.scala
+++ b/api/src/main/scala/org/ensime/api/incoming.scala
@@ -19,14 +19,9 @@ sealed trait RpcRequest
 sealed trait RpcStartupRequest extends RpcRequest
 
 /**
- * Respond with a `ConnectionInfo`.
- */
-case object AsynConnectionInfoReq extends RpcStartupRequest
-
-/**
  * Responds with a `ConnectionInfo`.
  */
-@deprecating("prefer AsynConnectionInfoReq")
+@deprecating("Please switch to asynchronous connection handling.")
 case object ConnectionInfoReq extends RpcStartupRequest
 
 // related to managing the state of the analyser
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -44,7 +44,7 @@ final case class DebugVmError(
 sealed trait GeneralSwankEvent extends EnsimeEvent
 sealed trait DebugEvent extends EnsimeEvent
 
-final case class AsyncConnectionInfo(
+final case class GreetingInfo(
   pid: Option[Int] = None,
   implementation: EnsimeImplementation = EnsimeImplementation("ENSIME"),
   version: String = "1.9.6"
--- a/api/src/main/scala/org/ensime/api/outgoing.scala
+++ b/api/src/main/scala/org/ensime/api/outgoing.scala
@@ -44,6 +44,12 @@ final case class DebugVmError(
 sealed trait GeneralSwankEvent extends EnsimeEvent
 sealed trait DebugEvent extends EnsimeEvent
 
+final case class AsyncConnectionInfo(
+  pid: Option[Int] = None,
+  implementation: EnsimeImplementation = EnsimeImplementation("ENSIME"),
+  version: String = "1.9.6"
+) extends EnsimeEvent
+
 /**
  * Generic background notification.
  *
--- a/core/src/main/scala/org/ensime/core/Project.scala
+++ b/core/src/main/scala/org/ensime/core/Project.scala
@@ -74,7 +74,16 @@ class Project(
   }
   context.actorOf(Props(new ClassfileWatcher(searchService :: reTypecheck :: Nil)), "classFileWatcher")
 
-  def receive: Receive = awaitingConnectionInfoReq
+  // If the system property `ensime.server.async` is missing or not set to 
+  // `true` the old behaviour will be used.
+  private val handleAsyncConnections: Boolean = scala.util.Try(System.getProperty("ensime.server.async") == "true").toOption.getOrElse(false)
+
+  def receive: Receive = {
+    if (handleAsyncConnections)
+      handleRequests
+    else
+      awaitingConnectionInfoReq
+  }
 
   private def handleConnectionInfoReq(s: ActorRef): Unit = {
     s ! ConnectionInfo()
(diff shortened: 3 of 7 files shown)


jan0sch:
I don't know if this is too simple. There is just another message right now which is handled the same way. Is that the way to go?

api/src/main/scala/org/ensime/api/incoming.scala
```diff
@@ -19,8 +19,14 @@ sealed trait RpcRequest
 sealed trait RpcStartupRequest extends RpcRequest
 
 /**
+ * Respond with a `ConnectionInfo`.
+ */
+case object AsynConnectionInfoReq extends RpcStartupRequest
```
fommil: oh, no I was thinking there would be no request for this. It would just be provided on connection.

We could have a system property or something (on by default) that does the current (legacy) behaviour so we don't break the clients.

jan0sch: Ah, ok. I'll rewrite it then. :)

jan0sch:
Does the code go in the right direction now? If yes, I'll try adding a test.

Besides that: Do you prefer the commits to be squashed before merging?

fommil:
We squash when we merge you don't need to do that.

I'll take a look later, thanks!

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -74,7 +74,23 @@ class Project(
   }
   context.actorOf(Props(new ClassfileWatcher(searchService :: reTypecheck :: Nil)), "classFileWatcher")
 
-  def receive: Receive = awaitingConnectionInfoReq
+  // If the system property `ensime.server.async` is missing or not set to 
+  // `true` the old behaviour will be used.
+  private val handleAsyncConnections: Boolean = scala.util.Try(System.getProperty("ensime.server.async") == "true").toOption.getOrElse(false)
```
fommil: can you please add this to the EnsimeServerConfig with default values in application.conf?

fommil: also, please invert the name. It should be something like `ensime.legacy.connectionInfoReq = true`

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -74,7 +74,23 @@ class Project(
...
+
+  def receive: Receive = {
+    if (handleAsyncConnections)
+      handleRequests
+    else
+      awaitingConnectionInfoReq
+  }
+
+  private def handleConnectionInfoReq(s: ActorRef): Unit = {
```
fommil: it'd be best to put this inline above

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -74,7 +74,23 @@ class Project(
...
   context.actorOf(Props(new ClassfileWatcher(searchService :: reTypecheck :: Nil)), "classFileWatcher")
 
-  def receive: Receive = awaitingConnectionInfoReq
+  // If the system property `ensime.server.async` is missing or not set to 
+  // `true` the old behaviour will be used.
+  private val handleAsyncConnections: Boolean = scala.util.Try(System.getProperty("ensime.server.async") == "true").toOption.getOrElse(false)
+
+  def receive: Receive = {
+    if (handleAsyncConnections)
+      handleRequests
```
fommil: wouldn't this be best as a `become`?

fommil: in fact, isn't this best in `preStart`?

protocol-swanky/src/main/scala/org/ensime/server/protocol/swank/SwankFormats.scala
```diff
@@ -357,6 +357,7 @@ object SwankProtocolResponse {
   implicit object EnsimeEventFormat extends TraitFormat[EnsimeEvent] {
     def write(ee: EnsimeEvent): Sexp = ee match {
       case e: AnalyzerReadyEvent.type => wrap(e)
+      case e: AsyncConnectionInfo => wrap(e)
```
jan0sch: I am not quite sure here. Is it okay that way?

api/src/main/scala/org/ensime/api/outgoing.scala
```diff
@@ -44,6 +44,12 @@ final case class DebugVmError(
 sealed trait GeneralSwankEvent extends EnsimeEvent
 sealed trait DebugEvent extends EnsimeEvent
 
+final case class AsyncConnectionInfo(
```
fommil: I wouldn't call it async... how about `GreetingInfo` ?

protocol-swanky/src/main/scala/org/ensime/server/protocol/swank/SwankFormats.scala
fommil: yeah I think that looks fine

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -74,22 +74,11 @@ class Project(
...
 
-  // If the system property `ensime.server.async` is missing or not set to 
-  // `true` the old behaviour will be used.
-  private val handleAsyncConnections: Boolean = scala.util.Try(System.getProperty("ensime.server.async") == "true").toOption.getOrElse(false)
-
   def receive: Receive = {
-    if (handleAsyncConnections)
-      handleRequests
-    else
+    if (serverConfig.legacy.connectionInfoReq)
```
fommil: I'm not too keen on this boolean check everywhere. Can you set the `receive` in the `preStart`?

jan0sch: I'm not sure if `context.become` is clean in `preStart`. But I'll check that and refactor accordingly.

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -180,7 +173,10 @@ class Project(
 
     // added here to prevent errors when client sends this repeatedly (e.g. as a keepalive
     case ConnectionInfoReq =>
-      sender() ! ConnectionInfo()
+      if (serverConfig.legacy.connectionInfoReq)
+        sender() ! ConnectionInfo()
+      else
+        sender() ! AsyncConnectionInfo()
```
fommil: this should always return the `ConnectionInfo`.

The `Greeting` should be sent as soon as the actor becomes `handleRequests`, which could be on startup if the legacy setting is disabled.

jan0sch: Ah, ok. But to whom do I send it?

core/src/main/scala/org/ensime/core/Project.scala
fommil: this is akka... nothing is clean :smile:

core/src/main/scala/org/ensime/core/Project.scala
```diff
@@ -139,6 +134,10 @@ class Project(
     }
     debugger = context.actorOf(DebugActor.props(delayedBroadcaster, searchService), "debugging")
     docs = context.actorOf(DocResolver(), "docs")
+    if (serverConfig.legacy.connectionInfoReq)
+      context.become(awaitingConnectionInfoReq)
```
fommil: this branch is redundant, but I guess it doesn't matter

jan0sch:
Yep, I don't know where/when to send it. Upon which message shall I send back the `GreetingInfo`?

fommil:
it's an async `Message`, it doesn't need a request. You just send it when the actor starts (only for the new version, not the legacy mode)

jan0sch:
Excuse my stupidity but I don't see how. I only know of sending via tell (`ref ! message`) or using things like pubsub mediator. ❓ 🤔

fommil:
`broadcaster ! <message>` :smile:

jan0sch:
Oh my... stupid me. I should have guessed that. Coding right away... :-)

fommil:
I kicked off the builds again, seemed to be transient failures.

fommil:
hmm, seems that the cache directory doesn't have the right version of sbt in `project/build.properties`, which is really slowing things down.

fommil:
can you please rebase or merge from the 2.0 branch? That should solve the CI issues

fommil:
oh, what the hell.... YOLO!

fommil:
I got paranoid that we didn't have coverage around this so I started up the server in Emacs, and everything is ok :relaxed:

PR #1875, opened 2017-09-28
Convert the SearchService to use FS2

### LINKED ISSUE #1756: backpressure in the indexer
now that writing to the index (the graph) is really slow, we need something like #1351 because I'm loathe to reintroduce the #1739

to put some numbers on this, it means we need 4g to run ensime on ensime-server, but it used to be perfectly fine with 1g... and after initial index it's fine with a few hundred MB.

Workarounds are:

in `global.sbt`

```
import org.ensime.EnsimeKeys._

// to give more RAM
ensimeJavaFlags ++= Seq("-Xmx4g")

// point to a ram disk if you're on HDD
ensimeCachePrefix in ThisBuild := Some(file("/tmp/fommil-ensime"))
```

in `.ensime-server.conf`

```
akka.search-service-dispatcher.thread-pool-executor.core-pool-size-max = 1
```

to slow down the indexer.

Personally I don't really care because I have 64GB RAM so although this is a pretty important ticket, I am not going to be picking it up :stuck_out_tongue:

iravid:
Resolves #1756.

Summary of the changes:
- Added a dependency on fs2 0.9.7 (would love to use 0.10 but that means using cats - we need to decide separately)
- SearchService has been split up into three parts:
   - The plain request/response methods stay in `SearchService` - `search*`, `persist`, `delete`, etc.
    - The actor that managed the indexing is now gone and replaced with `IndexStream` - an FS2 stream that is fed via an `fs2.async.mutable.Queue` and manages the indexing in the same manner
    - The file change listener is replaced by a method on `IndexStream` that returns a listener that is hooked up to the stream's input queue

Other changes:
- The SearchService API has been converted to use `Task`. I'd love to do something more parametric in the effect type, but we need to discuss use of an FP library for that.
- The above change led to a lot of `.unsafe*` calls across the code. I think this is a good path that we can follow to continue converting more parts of the code to be safer.
- We need to pass around an fs2.Scheduler (this is gone in 0.10 of fs2), an fs2.Strategy (for various FS2 stuff) and an ExecutionContext (for suspending Futures into Tasks). We'll get rid of the execution contexts eventually.

### DIFF
--- a/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
+++ b/core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
@@ -3,6 +3,7 @@
 package org.ensime.indexer
 
 import org.apache.lucene.search.DisjunctionMaxQuery
+import fs2.Strategy
 
 import scala.concurrent._
 import scala.concurrent.duration._
@@ -66,11 +67,11 @@ class SearchServiceSpec
         (config.projects.head.targets.head.file / "org/example/Foo$.class").toFile
 
       classfile shouldBe 'exists
-      service.findUnique("org.example.Foo$") shouldBe defined
+      service.findUnique("org.example.Foo$").unsafeRun() shouldBe defined
 
       classfile.delete()
       refresh() shouldBe ((1, 0))
-      service.findUnique("org.example.Foo$") shouldBe empty
+      service.findUnique("org.example.Foo$").unsafeRun() shouldBe empty
     }
   }
 
@@ -211,7 +212,7 @@ class SearchServiceSpec
           )
       }
 
-      val bazHits = service.searchClasses("Baz", 10).map(_.fqn)
+      val bazHits = service.searchClasses("Baz", 10).unsafeRun().map(_.fqn)
       bazHits should contain theSameElementsAs (
         Seq(
           "org.example2.Baz",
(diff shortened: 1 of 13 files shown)


iravid:
@fommil, would you like to review?

fommil:
I'm on holiday for a week :-) can you maybe get somebody who knows something about fs2 to take a look? I'm happy for you to merge if there are no objections

iravid:
Enjoy your holiday! I’ll take care of finding a reviewer :-)

core/src/it/scala/org/ensime/indexer/SearchServiceSpec.scala
```diff
@@ -501,32 +512,39 @@ object SearchServiceTestUtils {
...
     fqn: String,
     hierarchyType: Hierarchy.Direction
-  )(implicit service: SearchService, p: Position): Hierarchy = {
+  )(implicit service: SearchService,
+    p: Position,
+    strategy: Strategy): Hierarchy = {
     val hierarchy =
-      Await.result(service.getTypeHierarchy(fqn, hierarchyType), Duration.Inf)
+      Await.result(
+        service.getTypeHierarchy(fqn, hierarchyType).unsafeRunAsyncFuture(),
```
lukasz-golebiewski: Is there any point in awaiting a future infinitely instead of running it synchronously with `unsafeRun`?

iravid: Looks pretty much equivalent from fs2's code. I was just trying to minimize introduced changes :-)

iravid:
Thanks for reviewing @lukasz-golebiewski. Will wait a few more days and merge afterwards.

fommil:
Are you too scared to merge? :-D

iravid:
Hehe, nah, entirely YOLO on this branch ;) Just a bit busy with other stuff and wondering if I should rewrite this on fs2 0.10.

aarshkshah1992:
Any particular reason this hasn't been merged till now ?

iravid:
Busy period for me :-) Will get back to this eventually.

fommil:
@iravid are you going to be at scalasphere or lambdaconf or scalaio this year? Would love to talk further about #1902 ... I think a full rewrite using metacp is the way to go.

iravid:
@fommil Slim chances, but maybe LambdaConf. I’ll let you know if so.

PR #1876, opened 2017-09-29
Support file names with '%' characters

oleastre:
Current EnsimeFile implementation handles url as file path, just removing the 'file://' part.
But urls can be encoded, '%' characters are converted to '%25', and this is ignored by the EnsimeFile constructor.
If your file contains a '%' in it's name, ensime will search for a file path with '%25' that does not exists.

This add a test case that shows the problem. I'm trying to get a proper fix for the 2.0 branch (ie without switching everything to java.nio)

### DIFF
--- a/util/src/main/scala/org/ensime/util/ensimefile.scala
+++ b/util/src/main/scala/org/ensime/util/ensimefile.scala
@@ -41,11 +41,11 @@ package object ensimefile {
     implicit val DefaultCharset: Charset = Charset.defaultCharset()
   }
 
-  private val ArchiveRegex = "(?:(?:jar:)?file:)?([^!]++)!(.++)".r
-  private val FileRegex = "(?:(?:jar:)?file:)?(.++)".r
+  private val ArchiveRegex = "((?:jar:)?file:)?([^!]++)!(.++)".r
+  private val FileRegex = "((?:jar:)?file:)?(.++)".r
   def EnsimeFile(path: String): EnsimeFile = path match {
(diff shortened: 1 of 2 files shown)


oleastre:
Proposed patch: url dcode file names if it starts with jar: or file: (capture the header part in the file and archive regex).
This solves my indexer problem I reported on gitter.

iravid:
Thanks for tackling this @oleastre! It looks like the integration tests are failing; you can see that in the link to the Drone CI.

Specifically, `MissingJarTargetTest` and `JarTargetTest`. You can see that at the bottom of the build output.

You can run those locally using `it:testOnly *JarTargetTest`. Can you look into why they're failing?

oleastre:
I did this quick fix to be able to continue work on my main project :) 
And I will have a look in the coming days to the failing tests.

iravid:
Sounds good!

oleastre:
New version, hopefully it will work on windows (I checked the failing tests in a vm).
Do you want me to squash the commits before merge ?

oleastre:
Seems everything is working except a timeout in travis. Is this expected ?
My new code path is maybe a lot slower that the old one, let me know if it's a blocking issue.

In my tests, I added some logging in the conversion code, and I'm was surprised by the number of times the same path is converted from String to Path. Maybe something can be done there.

oleastre:
Is there still something to do before this could be merged ?

yyadavalli:
@oleastre The failure is present on other 2.0 pull requests, so I think it's okay to merge.

I will look at the failure before tagging it.

oleastre:
Thanks a lot !

yyadavalli:
Welcome

PR #1880, opened 2017-10-06
Format using scalafmt

### LINKED ISSUE #1874: scalafmt on 2.0 branch
backport https://github.com/ensime/ensime-server/pull/1873 to 2.0

This will make it easier to move bugfixes and features between branches.

yyadavalli:
Resolve #1874


fommil:
thanks! I was half tempted to do this myself.

don't forget to run it in scala 2.10 and scala 2.11 too... and do a full compile locally because there seems to be a bug where some files get missed by the first pass.

Issue #1823, opened 2017-07-01
json/s-express cleanups


fommil:
- don't try to support BigDecimal, admit that Double is enough for our needs and avoid all the pitfalls discovered in circe.
- don't throw exceptions. Change the sig to return Either (or Validated)
- remove the RootJson typeclass
- use simulacrum

PR #1881, opened 2017-10-09
LSP support -- RPC library

vovapolu:
Rewritten version of https://github.com/dhpcs/scala-json-rpc

### DIFF
--- a/lsp/src/main/scala/org/ensime/lsp/JsonUtils.scala
+++ b/lsp/src/main/scala/org/ensime/lsp/JsonUtils.scala
@@ -1,3 +1,5 @@
+// Copyright: 2010 - 2017 https://github.com/ensime/ensime-server/graphs
+// License: http://www.gnu.org/licenses/gpl-3.0.en.html
 package org.ensime.lsp
 
 import spray.json._
--- a/lsp/src/main/scala/org/ensime/lsp/JsonUtils.scala
+++ b/lsp/src/main/scala/org/ensime/lsp/JsonUtils.scala
@@ -1,0 +1,12 @@
+package org.ensime.lsp
+
+import spray.json._
+
+object JsonUtils {
+  def wrapperFormat[A, B: JsonFormat](wrap: B => A,
+                                      unwrap: A => B): JsonFormat[A] =
+    new JsonFormat[A] {
+      def write(obj: A): JsValue = unwrap(obj).toJson
+      def read(j: JsValue): A    = wrap(j.convertTo[B])
+    }
+}
--- a/lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
+++ b/lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
@@ -9,10 +9,9 @@ import shapeless.tag.@@
 
 import scala.util.Try
 
-private object RpcConversions extends DefaultJsonProtocol with FamilyFormats {
+object JsInnerFormats {
 
-  implicit override def eitherFormat[A: JsonFormat, B: JsonFormat]
-    : JsonFormat[Either[A, B]] = super.eitherFormat[A, B]
+  trait JsInnerField
 
   implicit val innerJsValue: JsonFormat[JsValue @@ JsInnerField] =
     new JsonFormat[JsValue @@ JsInnerField] {
@@ -36,6 +35,14 @@ private object RpcConversions extends DefaultJsonProtocol with FamilyFormats {
       }
       def write(obj: JsArray @@ JsInnerField): JsValue = obj
     }
+}
+
+private object RpcConversions extends DefaultJsonProtocol with FamilyFormats {
+
+  import JsInnerFormats._
+
+  implicit override def eitherFormat[A: JsonFormat, B: JsonFormat]
+    : JsonFormat[Either[A, B]] = super.eitherFormat[A, B]
 
   implicit object CorrelationIdFormat extends JsonFormat[CorrelationId] {
     override def read(j: JsValue): CorrelationId = j match {
--- a/lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
+++ b/lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
@@ -9,7 +9,7 @@ import JsonRpcMessages._
 import org.ensime.lsp.JsonUtils
 import shapeless.tag.@@
 
-import scala.util.Try
+import scala.util.{ Failure, Success, Try }
 
 object JsInnerFormats {
 
@@ -121,10 +121,11 @@ private object RpcConversions extends DefaultJsonProtocol with FamilyFormats {
         Try(j.convertTo[JsonRpcRequestMessageBatch]) orElse
         Try(j.convertTo[JsonRpcResponseMessageBatch])
 
-      tryAll.fold(
-        _ => deserError[JsonRpcMessage]("Error during JsonRpcMessage parsing"),
-        x => x
-      )
+      tryAll match {
+        case Failure(_) =>
+          deserError[JsonRpcMessage]("Error during JsonRpcMessage parsing")
+        case Success(x) => x
+      }
     }
 
     def write(obj: JsonRpcMessage): JsValue = obj match {
--- a/lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
+++ b/lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
@@ -37,6 +37,21 @@ private object RpcConversions extends DefaultJsonProtocol with FamilyFormats {
       def write(obj: JsArray @@ JsInnerField): JsValue = obj
     }
 
+  implicit object CorrelationIdFormat extends JsonFormat[CorrelationId] {
+    override def read(j: JsValue): CorrelationId = j match {
+      case JsNull        => NullId
+      case JsNumber(num) => NumberId(num)
+      case JsString(str) => StringId(str)
+      case _             => deserError[CorrelationId]("Wrong CorrelationId format")
+    }
+
+    override def write(obj: CorrelationId): JsValue = obj match {
+      case NullId        => JsNull
+      case NumberId(num) => JsNumber(num)
+      case StringId(str) => JsString(str)
+    }
+  }
+
   implicit object JsonRpcResponseMessageFormat
       extends RootJsonFormat[JsonRpcResponseMessage] {
     def read(j: JsValue): JsonRpcResponseMessage = j match {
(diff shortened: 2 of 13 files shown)


lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
```diff
@@ -0,0 +1,134 @@
...
+import shapeless.tag.@@
+
+import scala.util.Try
+
+private object RpcConversions extends DefaultJsonProtocol with FamilyFormats {
+
+  implicit override def eitherFormat[A: JsonFormat, B: JsonFormat]
+    : JsonFormat[Either[A, B]] = super.eitherFormat[A, B]
+
+  implicit val innerJsValue: JsonFormat[JsValue @@ JsInnerField] =
```
vovapolu: I couldn't serialize simple `JsValue` (it was ambiguous implicit resolution), so added tag to it. Maybe there is some simpler solution?

vovapolu:
Oops, I forgot to add headers. Also next step is to add tests.

lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
iravid: `JsonFormat` is invariant in the type parameter, so it's strange that this happens. Can you show a snippet that fails?

If it's actually problematic, you can try prioritizing the implicits in this object using inheritance (place lower priority implicits in another abstract class, and make `RpcConversions` extend that class)

vovapolu: Without tags it throws following exception:
```
ambiguous implicit values:
[error]  both object JsValueFormat in trait AdditionalFormats of type org.ensime.lsp.rpc.RpcConversions.JsValueFormat.type
[error]  and method familyFormat in trait LowPriorityFamilyFormats of type [T, Repr](implicit gen: shapeless.LabelledGeneric.Aux[T,Repr], implicit sg: shapeless.Cached[shapeless.Strict[org.ensime.lsp.rpc.RpcConversions.WrappedRootJsonFormat[T,Repr]]])spray.json.RootJsonFormat[T]
[error]  match expected type spray.json.JsonFormat[spray.json.JsValue]
[error]   implicit val JsonRpcResponseSuccessMessageFormat
[error]                ^
```

fommil:
@dragos any chance you could take a quick look?

lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
iravid: Ah. Hmm. I'm not sure what's the most elegant solution here. I can suggest:
1. Importing your formats at the call-site to prevent the `RootJsonFormat` from kicking in
2. Use your implicit explictly

Perhaps @fommil will have a nicer idea.

lsp/src/main/scala/org/ensime/lsp/rpc/companions.scala
```diff
@@ -0,0 +1,137 @@
...
+
+import spray.json.{ JsObject, JsonFormat }
+
+import org.ensime.lsp.rpc.messages._
+import JsonRpcMessages._
+import shapeless.tag
+
+import scala.util.{ Failure, Success, Try }
+
+sealed trait RpcCompaionError {
```
dragos: Typo, this should be `RpcCompanionError`

dragos:
I'm not an authority on the JSON-RPC protocol but this looks like a minimal implementation that could work. If I were to quibble, this isn't an LSP implementation yet (or you want to plug mine directly on top of this one?), just a JSON-RPC implementation. Also, a bunch of tests would be great to have.

vovapolu:
@dragos I split my changes into several branches to not create a one huge PR that no one would want to read. One of these branches has LSP implementation on top of this PR.
And tests are coming soon :)

dhpiggott:
I presume you're already going this way, but in case not, do feel free to pinch https://github.com/dhpcs/scala-json-rpc/blob/master/scala-json-rpc/src/test/scala/com/dhpcs/jsonrpc/JsonRpcMessageSpec.scala and https://github.com/dhpcs/scala-json-rpc/blob/master/scala-json-rpc/src/test/scala/com/dhpcs/jsonrpc/MessageCompanionsSpec.scala as basis for the tests.

vovapolu:
@dhpiggott Yep, I'm rewriting them now :)

lsp/src/main/scala/org/ensime/lsp/rpc/companions.scala
```diff
@@ -0,0 +1,137 @@
...
+  ): Either[RpcCompaionError, _ <: A] =
+    commands.find(_.method == jsonRpcRequestMessage.method) match {
+      case None => Left(UnknownMethod)
+      case Some(command) =>
+        jsonRpcRequestMessage.params match {
+          case None           => Left(NoParams)
+          case Some(Right(_)) => Left(NoNamedParams)
+          case Some(Left(obj)) =>
+            Try(command.format.read(obj)) match {
+              // We do this just to reset the path in the success case.
```
dhpiggott: I don't think these comment applies here.

fommil:
I'll try to take a look at the json implicits tomorrow... it's very fiddly, sorry, I'd like to fix that (for everybody) with https://gitlab.com/fommil/stalactite

lsp/src/main/scala/org/ensime/lsp/JsonUtils.scala
```diff
@@ -0,0 +1,14 @@
+// Copyright: 2010 - 2017 https://github.com/ensime/ensime-server/graphs
+// License: http://www.gnu.org/licenses/gpl-3.0.en.html
+package org.ensime.lsp
+
+import spray.json._
+
+object JsonUtils {
```
fommil: this would typically be called `.instance` on the companion of `JsonFormat`

lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
```diff
@@ -0,0 +1,176 @@
...
+import spray.json._
+import shapeless._
+import messages._
+import JsonRpcMessages._
+import org.ensime.lsp.JsonUtils
+import shapeless.tag.@@
+
+import scala.util.{ Failure, Success, Try }
+
+object JsInnerFormats {
```
fommil: I don't understand why this is needed. I'll try deleting it and see what fails to compile.

lsp/src/main/scala/org/ensime/lsp/rpc/companions.scala
```diff
@@ -0,0 +1,136 @@
+// Copyright: 2010 - 2017 https://github.com/ensime/ensime-server/graphs
+// License: http://www.gnu.org/licenses/gpl-3.0.en.html
+package org.ensime.lsp.rpc.companions
```
fommil: why are these called companions? That confuses me, because "companion" has a scala meaning.

lsp/src/main/scala/org/ensime/lsp/rpc/companions.scala
```diff
@@ -0,0 +1,136 @@
...
+import spray.json.{ JsObject, JsonFormat }
+import org.ensime.lsp.rpc.messages._
+import JsonRpcMessages._
+import org.ensime.lsp.rpc.JsInnerFormats.JsInnerField
+import shapeless.tag
+
+import scala.util.{ Failure, Success, Try }
+
+sealed trait RpcCompanionError {
+  val describe: String
```
fommil: best to make this a sealed abstract class that takes a parameter. vals in traits are evil.

lsp/src/main/scala/org/ensime/lsp/rpc/companions.scala
```diff
@@ -0,0 +1,136 @@
...
+}
+case class OtherError(err: String) extends RpcCompanionError {
+  override val describe = err
+}
+
+object RpcCompanionError {
+  def apply(err: String): OtherError = OtherError(err)
+}
+
+case class RpcCommand[A](method: String)(implicit val format: JsonFormat[A])
```
fommil: taking the JsonFormat here is bad form... it is mixing your data model with the polymorphism. Best put the requirement on the thing doing the marshalling. e.g. you wouldn't want `Option` to take implicit evidence for everything its `A` would want to do.

lsp/src/main/scala/org/ensime/lsp/rpc/companions.scala
vovapolu: Originally there were `*Companion` classes. So I named file with them as `companions` :)

lsp/src/main/scala/org/ensime/lsp/rpc/messages.scala
```diff
@@ -0,0 +1,222 @@
...
+package org.ensime.lsp.rpc.messages
+
+import org.ensime.lsp.rpc.JsInnerFormats.JsInnerField
+import shapeless.tag
+import shapeless.tag.@@
+import spray.json._
+
+import scala.collection.immutable.Seq
+
+object JsonRpcMessages {
```
fommil: putting ADTs in an `object` tends to amplify bugs in the scala compiler. Best put them in a package.

lsp/src/main/scala/org/ensime/lsp/rpc/RpcFormats.scala
fommil: ok, I understand why you did this. Sorry, I could have saved you the effort of doing this overly complex workaround. The solution is the same thing you did for `either`...  look at the comments in `JerkyFormats` to understand what the problem is and the workaround.

stalactite / scalaz-deriving solves this trivially. In fact, refactoring our json formatter to use export-hook style solves it too. I meant to do that before you started but I had no time :disappointed:

fommil: i.e. https://github.com/ensime/ensime-server/issues/1823

fommil: the fact that the formatters are `implicit object` instead of `implicit val` could be the core problem, I may just fix that now.

lsp/src/main/scala/org/ensime/lsp/rpc/companions.scala
vovapolu: I added implicit val to `RpcCommand` to explicitly denote what `JsonFormat`s connected with corresponding methods. Format is generated only in `RpcCommand` definition, not when the writing/reading is performed.

lsp/src/main/scala/org/ensime/lsp/rpc/messages.scala
vovapolu: I can't write top-level type binding for `Params`

PR #1882, opened 2017-10-13
LSP support -- LSP API

vovapolu:
*Depends on https://github.com/ensime/ensime-server/pull/1881 PR.*
Rewritten version of subset of LSP protocol methods from https://github.com/dragos/dragos-vscode-scala


fommil:
:smile: you are writing code faster than I can review. At some poitn I'll just merge it all

:fire: :fire: :fire:

PR #1897, opened 2017-11-29
rewrite the json library to be simpler and use scalaz.deriving

fommil:
@vovapolu I had to disable some of the lsp tests, but I think the tests are wrong. They are asserting things like "needs an id" yet the underlying objects have a decoder for `JsNull`.

If this is essential behaviour I can add an extra parameter into `JsConfig` to support "require null to exist" or something but I think it's fairly common for missing fields to be interpreted the same as if they had a `null` value.

Also, there was a "companions spec" test that I'm quite certain is just testing things that it is defining in the same file. Is there any point in keeping that file around?


lsp/src/main/scala/org/ensime/lsp/api/companions.scala
```diff
@@ -6,23 +6,7 @@ import org.ensime.lsp.rpc.companions._
...
-    : JsonFormat[TextDocumentHoverRequest] =
-    JsonFormat[TextDocumentPositionParams]
-      .xmap(TextDocumentHoverRequest(_), _.params)
-
-  implicit val textDocumentCompletionRequestFormat
-    : JsonFormat[TextDocumentCompletionRequest] =
-    JsonFormat[TextDocumentPositionParams]
-      .xmap(TextDocumentCompletionRequest(_), _.params)
-
+object ServerCommands extends {
```
vovapolu: extends what?

lsp/src/test/scala/org/ensime/lsp/rpc/JsonRpcMessageSpec.scala
```diff
@@ -434,22 +377,6 @@ class JsonRpcMessageSpec extends FreeSpec {
                    |}""".stripMargin.parseJson
       willFailToDecode[JsonRpcResponseMessage, DeserializationException](json)
     }
-    "without an error or a result" - {
```
vovapolu: why delete this?

fommil: that might be an accident.

fommil: decodes as `JsonRpcResponseSuccessMessage(2.0,JsNull,NumberId(0))` which is valid.

vovapolu:
Maybe we could change removed tests to demonstrate behavior of parser?

vovapolu:
Oh, tests with `null` can't be tested by roundtrip. `null` is converted to empty field.

fommil:
I don't think there is much value in tests to assert that things *don't* deserialise. That is guaranteed by the library and the strength of the ADT. If we are allowing too much then it means our ADT is wrong, not that serialisation is broken. (i.e. we'd effectively be writing more tests for the `json` package and it's really very well covered).

The roundtrip tests of the protocol messages are useful because then we catch regressions AND it serves as good documentation / proof to downstream implementers. I don't think I've removed any of those.

lsp/src/main/scala/org/ensime/lsp/api/companions.scala
fommil: because, scala. *sigh*

PR #1901, opened 2017-12-03
introduce Canon typeclass

### LINKED ISSUE #1815: replace use of Poly with derived typeclasses
Especially if typeclass derivation gets faster via https://github.com/propensive/magnolia or new compiler features.

It feels safer to me. Less magical. Therefore simpler, and better.

fommil:
close #1815


fommil:
green lights locally, drone still breaks on maven central and desperately needs a cache.

PR #1905, opened 2017-12-07
added scalafix and ran it for a bunch of rules

fommil:


.scalafix.conf
```diff
@@ -0,0 +1,5 @@
+ExplicitResultTypes.unsafeShortenNames = true
+
+NoInfer.symbols = [
+  "scala.Predef.any2stringadd"
+]
```
ches: Be warned of the caveat in [the docs]—this completely overwrites defaults like `Any` and `Product` that are probably desirable.

[the docs]: https://scalacenter.github.io/scalafix/docs/rules/NoInfer

fommil: Damn, so I just turned it all off? Oh well, at least it's all wired up now.

PR #1908, opened 2018-01-29
Remove unused TCP server

zainab-ali:
The `TCPServer` class is unused as of [here](https://github.com/ensime/ensime-server/blob/3.0/server/src/main/scala/org/ensime/server/Server.scala).  I'm unsure as to whether it needs to be kept in the codebase.  Since I couldn't find any documentation to that effect I figured I'd open a PR for removal.


zainab-ali:
Interesting... there's probably something I'm missing here.  I'll debug and get back to this.

PR #1917, opened 2018-02-27
FileWatchService using sensitive watch event modifier, and a bit of refactoring.

wibisono:
The main change is using  SensitivityWatchEventModifier.HIGH when registering path in watchService. This will allow 2 secs polling in mac OS/X which then allow FileWatcherSpec to pass. The default value of MEDIUM sensitivity check every 10 seconds.

The other two changes are just to split FileWatchService into three files, separating WatchListener trait and WatchKeyManager object, so no longer 600 LOC of Scala. (I was originally thinking about incorporating https://github.com/lloydmeta/schwatcher to replace watch key manager, but then it seems a lot of work).

Small changes on build.sbt, (relative path to sbt)  to allow travis to pass when executing prewarm task.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -208,7 +208,7 @@ TaskKey[Unit](
   // would be good to be able to do this without exiting the JVM...
   val sv = scalaVersion.value
   val cmd =
-    if (sys.env.contains("APPVEYOR")) """C:\sbt\bin\sbt.bat""" else "sbt"
+    if (sys.env.contains("APPVEYOR")) """C:\sbt\bin\sbt.bat""" else "../../sbt"
   sys.process
     .Process(
       Seq(cmd, s"++$sv!", "ensimeConfig", "ensimeServerIndex"),
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ script:
   - export JAVA_HOME=$(/usr/libexec/java_home)
   - curl https://raw.githubusercontent.com/paulp/sbt-extras/master/sbt -o sbt ; chmod 755 sbt
   - ./sbt ++$SCALA_VERSION! ";prep ;cpl"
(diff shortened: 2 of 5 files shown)


build.sbt
```diff
@@ -208,7 +208,7 @@ TaskKey[Unit](
   // would be good to be able to do this without exiting the JVM...
   val sv = scalaVersion.value
   val cmd =
-    if (sys.env.contains("APPVEYOR")) """C:\sbt\bin\sbt.bat""" else "sbt"
+    if (sys.env.contains("APPVEYOR")) """C:\sbt\bin\sbt.bat""" else "../../sbt"
```
fommil: This looks specific to your machine

wibisono: That's actually to make it works in Travis. The line fetching sbt with curl 
```
curl https://raw.githubusercontent.com/paulp/sbt-extras/master/sbt -o sbt ; chmod 755 sbt
```

Will not work when executing the prewarm task, since sbt is two levels up, while the task is running under `testing/cahe`. Reverting back to `sbt` only gives error file not found. 

I am not sure to fix this here, or in .travis.yml, (by for example setting current path to be able to access sbt)

wibisono: About the too big pull request, I'll come back to this after work or tomorrow, and make smaller PR just with the sensitivity change.

fommil: aaah, I thought this was teh appveyor file. ok, that makes sense. But I think you need a `if (appveyor) ... else if (travis) ... else ...` or it'll break things for everybody else.

fommil:
@wibisono I added the CI fix, well spotted! That really speeds up the Travis CI.

We should probably start thinking about using travis for GNU / Linux tests now, instead of a custom drone.

build.sbt
wibisono: I will just close this one and make a smaller pull request just the sensitivity modifier. The refactoring of FileWatchService is a bit too big for me also.

fommil:
@wibisono did you mean to close the PR? I only added the travis fix, I didn't include your changes to the file watcher.

PR #1919, opened 2018-02-28
Remove debugger

### LINKED ISSUE #1907: separate the debugging component into its own server
will simplify the codebase, reduce deps, and be more micro servicey.

Will require a new server to be written and sbt-ensime / etc should add the necessary jars to enable launching it.

fommil:
close #1907 and all debugger tickets


chipsenkbeil:
LGTM

PR #1923, opened 2018-02-28
Just the change on sensitivity modifier for watch service

wibisono:
As you suggested, this PR is without refactoring, just to change the modifier to poll more frequently.

The change in .travis.yml is just to allow sbt to be access from testing/cache so I don't have to even modify build.sbt.


.travis.yml
```diff
@@ -11,6 +11,7 @@ script:
     echo "ensimeServerProjectJars := Nil" >> ensime.sbt ;
   - export JAVA_HOME=$(/usr/libexec/java_home)
   - curl https://raw.githubusercontent.com/paulp/sbt-extras/master/sbt -o sbt ; chmod 755 sbt
+  - export PATH=$PWD:$PATH
```
fommil: not needed anymore

core/src/main/scala/org/ensime/filewatcher/FileWatchService.scala
```diff
@@ -180,12 +182,8 @@ class FileWatchService { self =>
       log.trace(s"register ${dir} with WatchService")
     if (!observers.isEmpty) {
       val key: WatchKey = try {
-        dir.toPath.register(
-          watchService,
-          ENTRY_CREATE,
-          ENTRY_MODIFY,
-          ENTRY_DELETE
-        )
+        val events: Array[WatchEvent.Kind[_]] = Array(ENTRY_CREATE, ENTRY_MODIFY, ENTRY_DELETE)
```
fommil: you need to run `sbt fmt` ... this is beyond the line limit.

fommil:
if you do a reformat, this is good for merge. Thanks!

.travis.yml
wibisono: True, removed

core/src/main/scala/org/ensime/filewatcher/FileWatchService.scala
wibisono: sbt fmt'ed

wibisono:
I did reformatted with sbt fmt, and remove the line on travis that is no longer needed.

fommil:
unfortunately we no longer have any automatic publishing on merge as our drone server is out of action, sorry, but you can at least build your own version locally.

Issue #1927, opened 2018-03-23
Go to type definition not to Class.java


hzenginx:
Leave this list and include an `x` in the relevant boxes:

- [x] I have read http://ensime.org/getting_help/
- [x] I have followed the troubleshooting guide for my text editor
- [ ] I have asked in http://gitter.im/ensime/ensime-server
- [ ] I have included a link to an example project
- [x] I have included exact bug reproduction instructions

Steps to reproduce:
go to definition of a type parameter in a `classOf[A]` statement. 
example: `classOf[String]`, when I try to go to the definition of String it goes to Class.java.

PR #1928, opened 2018-03-23
#1927 return the type not Class when asked on classOf

hzenginx:
fix for: https://github.com/ensime/ensime-server/issues/1927

looks like emacs is not supported in 3.0 hence I couldn't verify the solution, I hope test is sufficient.


fommil:
Interesting, I don't fully understand what the compiler is doing here. I thought we supported going to definition for type parameters.

hzenginx:
Me neither. Ensime works quite good in terms of regular type parameters. May be because `classOf` is a stup method?(Predef.scala:126)

Issue #1922, opened 2018-02-28
use gitlab ci for gnu/linux tests


fommil:
we no longer have the debugger, which dramatically reduces our CI requirements.

We can probably go back to using Travis for GNU/Linux as well as OSX tests now, without relying drone.

fommil:
@nafg showed me that it is now possible to use gitlab ci for GitHub projects, this is a much better alternative! We can also have people contribute hardware easily.

nafg:
Yeah, being able to stick runners anywhere and use them in various
combinations is a great ability of gitlab.

fommil:
an example gitlab ci config file, for inspiration https://gitlab.com/fommil/scalaz-deriving/blob/master/.gitlab-ci.yml

@gabro ! ping

gabro:
I'm a bit behind this, and I've definitely overpromised on the times, but I'll carve some time off my next weekend to look into this

fommil:
I've created a connection, so it should pick up any files that are added. https://gitlab.com/ensime/ensime-server/

PR #1931, opened 2018-03-31
Use GitLab-CI

Tomahna:
Would something like that do the trick for #1922 ?

I can't seem to access https://gitlab.com/ensime/ensime-server/, are you sure the repository is public ?

Anyway I've created an integration to test, you can find it here https://gitlab.com/krauscher/ensime-server.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,55 @@
+image: fommil/docker-openjdk-sbt:latest
+cache:
+  key: default
+  paths:
+    - project/.sbt
+variables:
+  AKKA_TEST_TIMEFACTOR: 10
+  COURSIER_CACHE: "${CI_PROJECT_DIR}/project/.sbt/coursier"
+  SBT_TASK_LIMIT: 4
+  SCALA_VERSION: 2.12.4
+
+stages:
+  - test
+  - integration
+  - publish
+
+# Test tasks
+git:
+  cache: {}
+  script:
+    - if $(git grep -qE "TODO|FIXME" *) ; then
+        echo "Please remove TODO or FIXME." ;
+        exit 1 ;
+      fi
+
+format:
+  script:
+    - sbt scalafmtSbtCheck compile:scalafmtCheck test:scalafmtCheck it:scalafmtCheck
+
+lint:
+  script:
+    - sbt compile test:compile
+    - sbt "compile:scalafixCli --test" "test:scalafixCli --test"
+
+test:
+  script:
+    - sbt compile test:compile
+    - sbt test
+
+it-test:
+  stage: integration
+  script:
+    - echo "ensimeServerJars := Nil" > ensime.sbt
+    - echo "ensimeServerProjectJars := Nil" >> ensime.sbt
+    - sbt compile it:compile ensimeConfig assembly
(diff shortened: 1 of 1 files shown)


fommil:
nice, almost there. The caching strategy I use in scalaz-deriving should work better than this... most of what you're caching is already in the base image but the gitlab cache is better served by doing the COURSIER downloads 

https://gitlab.com/fommil/scalaz-deriving/blob/master/.gitlab-ci.yml

I *think* that works for merges doing the cache updates, but I am not 100% sure about it.

One big drawback of gitlabci is that it is really hard to setup a matrix build, so if we want to add the scala 2.11 build back in, it is going to be a real pain.

Tomahna:
Yeah, you're right, I didn't check your docker image

Well, there is no matrix build in GitLab. However, we do not need to run the entire pipeline against 2.11 and 2.12. test and it-test should be enough so I can just add the tasks manually.

Tomahna:
A few additional thoughts:
- I don't think there is any direct cache merge, the cache is just pulled at the beginning and pushed at the end (with new files). It seems fine as it means the cache will just converge to its final content.
- GitLab public shared runners are small (Digital Ocean 4GB 2vCPUs droplets) but autoscaling. Thus, it's better to have many smaller tasks than a few big ones.

So it should run better now, the first stage takes between 10 to 15 minutes to complete. The integration tests takes around 50 minutes. It's not 100% reliable though.

fommil:
in my experience the cache does not converge which just means the machine can't cache the cache and every run has to upload a huge cache to S3, which has the end result of being slower than not using the cache (plus costing me more money when using my own runner and S3). Hence why I only push the cache on merges to master and pull it on PRs.

fommil:
ouch, 50 minutes is pretty harsh. Maybe we should time travis-ci for GNU/Linux to see if it's as bad.

Tomahna:
I don't think I see your problem here. If you just cache coursier dependencies, the worst thing that might happen is you cache dependencies that are not yet or not anymore in use. You can solve it by invalidating the cache once in a while.

Concerning the integration tests, I might be missing a tweak here and there, but I've got a lot of OOM errors when trying to be faster.

Anyway, I agree we should probably test if travis behaves better.

fommil:
@Tomahna the problem is that the cache gets downloaded and uploaded to S3 on every single PR. It never converges. I don't know why.

.gitlab-ci.yml
```diff
@@ -6,7 +6,7 @@ cache:
 variables:
   AKKA_TEST_TIMEFACTOR: 10
   COURSIER_CACHE: "${CI_PROJECT_DIR}/project/.sbt/coursier"
-  SBT_TASK_LIMIT: 4
+  SBT_OPTS: "-XX:+CMSClassUnloadingEnabled  -XX:+UseConcMarkSweepGC -Xss2m -Xms3g -Xmx3g"
```
fommil: no need, we have a `.jvmopts` file

Tomahna:
Yeah i saw it, it's mostly a test at that point.

.gitlab-ci.yml
Tomahna: Yeah i saw it, it's mostly a test at that point.

PR #1933, opened 2018-04-05
add GNU/Linux tests to travis.yml

Tomahna:
A few notes about this:
- Using the vm based infrastucture (sudo: true) helps mitigate https://github.com/travis-ci/travis-ci/issues/3775.
- The approach using stage and several smaller tasks is not by itself incompatible with having a build matrix. Currently language: scala + os: osx is not supported which makes it a pain to use in build matrix. (but there are probably workarounds)

It should complete #1922.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,75 @@
-language: java
-os:
-  - osx
+os: linux
+language: scala
 env:
   global:
     - AKKA_TEST_TIMEFACTOR=10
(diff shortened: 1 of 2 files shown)


Tomahna:
Tasks 2054.5 and 2054.7 should be running on OSX

fommil:
nice, do you recommend merging?

PR #1937, opened 2018-04-18
Publishing

fommil:
the infinite digression of just wanting to make a code change but realising that the CI is not setup to publish yet... *sigh*


fommil:
@Tomahna FYI, if you're around a review would be :100:

fommil:
oh wow, I found the button for gitlab! We now have options.

fommil:
nope, gitlab didn't show up after all. I'm not sure how to get it working for PRs so travis seems the best choice.

fommil:
I think I made a mistake with the secret envavrs.

.travis.yml
```diff
@@ -2,78 +2,69 @@ os: linux
...
-      brew install sbt;
-      export JAVA_HOME=$(/usr/libexec/java_home);
-    fi
+- if [[ "$TRAVIS_OS_NAME" = "osx" ]]; then brew update; brew install sbt; export JAVA_HOME=$(/usr/libexec/java_home);
+  fi
 stages:
-  - name: test
-  - name: integration
-  - name: publish
-    if: branch = master
```
Tomahna: actually this is probably a mistake I introduced as ensime does not have a master branch

Tomahna: (should be 3.0)

fommil: ah, I can fix that here, well spotted

Tomahna:
have you tried scoping them to the publish task ?
it's the only one which need them right ?

Tomahna:
But for secret variables, I'd go with the web ui though.

fommil:
hmm, the publish task is running in the PR, not on merge, so I think there is a typo here. I'll disable the publish task until we get that fixed.

Tomahna:
It's weird cause https://github.com/Tomahna/ensime-server/pull/1 didn't trigger publish.

Tomahna:
nevermind

fommil:
well this is already a big improvement, I'll change the ticket to be "get it publishing" which should be a smaller scope!

PR #1938, opened 2018-04-18
[ADANDONED] Replace indexer with metacp

### LINKED ISSUE #1902: rewrite indexer to use metacp
If I have any time to work on ensime 3.0 in 2018 I would like to move more in the direction of the semanticdb. This would involve designing a spec for `semanticdb-bronze` (class files only, including scalap for scala output) and `semanticdb-silver` (reconciled with source jars).

The analysis we could perform would be the same as ensime 2.0, but major advantages include:

- indexing / reconciling once for shared binaries, instead of per-project
- a much smaller per-project db/index created that can survive rebuilds (currently the weakest part of ensime 2.0)
- make way for supporting semanticdb-gold (i.e. semanticdb as created by the scalac plugin)

// @xeno-by @olafurpg

fommil:
close #1902 

For anybody wanting to pick this up (I've abandoned this, for the avoidance of doubt), you just need to implement `SearchService.scala` by calling out to metacp's API.

You don't even need to invoke metacp's indexer, we can do that in the build tool and assume the semanticdb compiler plugin is enabled for the project so you get indexing for free.


timcharper:
So much red! 😍

fommil:
heh, yeah... Rory is going to add some green.

DanielPorter:
@rorygraves hey Rory, I'm looking at picking this up. Sam mentioned you and shane might have done some work at Scalasphere. Any prior art here?

rorygraves:
Awesome - we did some prototype work generating the semantic db files from the classpath - will try and find the branch this evening

fommil:
You don't need to generate the semanticdb files if you assume the user has it as part of their build. ensime-sbt will need a small update to run metacp

fommil:
I have abandoned this PR... would be very happy for somebody to pick it up.

Tomahna:
I've started looking into it and plan to continue as soon as I have time for it. Might not be super soon so if anybody wants to pick it up before go ahead.

PR #1940, opened 2018-04-18
Prevent PR from triggering publish

Tomahna:
This fix the problem from #1937.
**BUT** how do we prevent someone to just remove this filter, open a PR and trigger the publish with his modification through the PR. GitLab solves this with protected variables which are visible only on protected branches but I'm not sure at all how travis does fix that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ before_install:
 stages:
   - name: test
   - name: integration
-#  - name: publish
-#    if: branch = 3.0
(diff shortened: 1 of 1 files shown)


Tomahna:
Ah nevermind encrypted variable are not accessible to untrusted builds. Retrospectively it's obvious

Tomahna:
should be good to go then

Tomahna:
Ok now it seems to behave properly

PR #1943, opened 2018-04-21
remove ENJIME (java support)

fommil:
there are Java LSP servers, so we no longer have a strong reason to support java in the server.

If anybody wanted to create a very minimal standalone Java LSP, these deleted files are exactly what you'd need.


albfan:
That is not the last word on this, hehe!

Issue #1935, opened 2018-04-18
get the LSP working in some clients


fommil:
There was some interest at ScalaSphere to maintain and use the LSP support. @rorygraves sounded vaguely interested and I swear I heard him say that @zainab-ali promised to write it all :wink: 

The following people voted to keep the LSP in ensime which, according to the way it works around here, means that they have volunteered themselves to help out:

- @gabro
- @laughedelic
- @wifasoi
- @markusa380
- @gruggiero
- @bonabeavis
- @chipsenkbeil
- @ktonga
- @everson

rorygraves:
Me too - I think this is the way to go.

everson:
oh, I didn't know those were the rules! But now I want to dance accordingly. 💃

fommil:
if the emacs lsp client turns out to be good enough, perhaps we could abandon SWANKY as well. Less stuff to maintain.

fommil:
btw with regards to launching, you may need to edit sbt-ensime to insert the `ensime-lsp` jars into the `.ensime` file.

gabro:
> you've not seen the emacs lsp client 😉 (I have, it's not good)

Maybe, but it's still work that can be delegated and shared across N languages.

Also, Emacs floats around 5-10% adoption according to the latest StackOverlflow survey. Good LSP clients exist for most languages above 25-30% adoption, which would mean targeting a 80-something% of developers. It sounds like a pretty good deal to me ;)

laughedelic:
> The following people voted to keep the LSP in ensime which, according to the way it works around here, means that they have volunteered themselves to help out

Whatevs.. Nobody promised anything so people volunteer as much as they do in fact.

> (e.g. picking up what @vovapolu has done and see whether that can be made server-implementation-agnostic and shared across VSCode/Atom/Sublime/...).
> I know @laughedelic was also interested in this area.

As I claimed before, I am interesting in supporting Ensime LSP server in atom-ide-scala: https://github.com/laughedelic/atom-ide-scala/issues/13, but it's blocked by https://github.com/ensime/ensime-lsp-clients/issues/2. I understand that @vovapolu now doesn't have time to maintain it and I'm glad to contribute some minor fixes/improvements if it will be needed, but someone who owns that repo (@vovapolu or @fommil?) should publish it or something. Even though it's MIT-licensed, I don't want to copy-paste the code or publish it myself.

I'm also interested in helping with other clients which can be extended in JS and reuse this common launcher code.

[...]

PR #1945, opened 2018-04-28
Fix RequestMessage parsing

### LINKED ISSUE #1911: LSP notification json objects are incorrectly decoded into requests
LSP notification json objects get incorrectly decoded into requests with a null id:

For example:
```scala
JsObject(Map(
  jsonrpc -> JsString(2.0),
  method -> JsString("textDocument/didOpen"),
  params -> JsObject(Map(
    "textDocument" -> JsObject(Map(
      "uri" -> JsString("file:///path/to/file.scala"),
      "languageId" -> JsString("scala-yay"),
      "version" -> JsNumber(0),
      "text" -> JsString("someFileText")))))))
```
Is decoded into:
```scala
JsonRpcRequestMessage(
  2.0,
  "textDocument/didOpen",
  Some(ObjectParams(JsObject(Map(
    "textDocument" ->
      JsObject(Map(
        "uri" -> JsString("file:///path/to/file.scala"),
        "languageId" -> JsString("scala-yay"),
        "version" -> JsNumber(0),
        "text" -> JsString("someFileText")
      )))))),NullId)
```
This uses the `JsReader[JsonRpcMessage].read(json)` for decoding.  The `Connection` uses this decoder for attempting to decode notifications.

laughedelic:
Fixes #1911. @zainab-ali was right from the beginning: https://github.com/ensime/ensime-server/issues/1911#issuecomment-362096344

> A quick glance at the LSP spec indicates that requests should always have a non-null id parameter. This isn't enforced within the `JsonRpcRequestMessage` decoder, although an attempt is made to do this in the `JsonRpcRequestOrNotificationMessage` decoder.

After that comment there is some confusing discussion about null ids which I don't really understand (in particular, how it is related to the problem). But anyway LSP spec is unambiguous about [`RequestMessage`](https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#request-message) id type:

```
id: number | string;
```

It's **not optional** and **cannot be `null`**. 

I've tested it locally and now server recognizes notifications and basic features like type on hover, symbols outline and go to definition work well 👌

### DIFF
--- a/lsp/src/main/scala/org/ensime/lsp/core/Connection.scala
+++ b/lsp/src/main/scala/org/ensime/lsp/core/Connection.scala
@@ -86,6 +86,7 @@ class Connection(inStream: InputStream,
       case Some(jsonString) =>
         readJsonRpcMessage(jsonString) match {
           case Left(e) =>
+            log.error(s"Invalid message: ${e} - ${jsonString}")
             msgWriter.write(e)
 
           case Right(message) =>
@@ -105,6 +106,7 @@ class Connection(inStream: InputStream,
               case request: JsonRpcRequestMessage =>
                 unpackRequest(request) match {
                   case Left(e) =>
+                    log.error(s"Invalid request: ${e} - ${request}")
                     msgWriter.write(e)
                   case Right(command) =>
                     msgWriter.write(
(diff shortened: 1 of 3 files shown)


fommil:
Thanks! I'm not keen on the way we've modelled lsp with runtime exceptions, but we are where we are.

rorygraves:
Thanks @laughedelic !

laughedelic:
> Did you test it on the shutdown request? I remember it contained null id.

@vovapolu I think the problem you probably had with the `shutdown` request is about empty `params`, not the `id`. See my comment in https://github.com/ensime/ensime-server/issues/1935#issuecomment-385140935 and the fix in https://github.com/scalameta/metals/pull/53.

PR #1946, opened 2018-04-29
Fix shutdown routine

laughedelic:
Adapted some pieces from https://github.com/scalameta/metals/pull/53. 

Server should respond to the shutdown request and exit on the exit notification. This procedure is required for the correct communication with the client.

This seems to fix problems with the server shutdown: _if_ server starts well, then it will also stops well when the client asks. It still doesn't always start well, but that's another problem.


laughedelic:
Thanks for the merge rights! :shipit:

PR #1947, opened 2018-04-30
Two or three kinds of null

dickmao:
There is nothing logically amiss about #1945 but please consider the following.

Via http://www.jsonrpc.org/specification,

> **Id**: An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification.

While LSP explicitly disallows null in requests, the underlying json-rpc spec advises making the distinction between a null value and an absent value.  Unfortunately, as written, https://github.com/ensime/ensime-server/blob/3.0/json/src/main/scala/spray/json/DerivedJsReader.scala#L32 cannot.

I am sympathetic to fommill and zainab-ali's reservations #1911 about classifying messages via trial-and-error runtime exceptions, and this PR does nothing to alleviate that.


json/src/main/scala/spray/json/JsReader.scala
```diff
@@ -59,8 +59,9 @@ object JsReader {
   }
 
   implicit def option[A: JsReader]: JsReader[Option[A]] = {
-    case JsNull => None
-    case a      => Some(JsReader[A].read(a))
+    case JsAbsent => None
+    case JsNull   => None
```
fommil: minor point, but you can do `case JsAbsent | JsNull`  here and avoid repeating the RHS (which is trivial, so feel free to ignore me).

lsp/src/main/scala/org/ensime/lsp/api/companions.scala
```diff
@@ -1,91 +0,0 @@
-// Copyright: 2010 - 2017 https://github.com/ensime/ensime-server/graphs
-// License: http://www.gnu.org/licenses/gpl-3.0.en.html
-package org.ensime.lsp.api.companions
```
fommil: kermit celebrating

lsp/src/main/scala/org/ensime/lsp/ensime/EnsimeLanguageServer.scala
```diff
@@ -255,10 +255,12 @@ class EnsimeLanguageServer(in: InputStream, out: OutputStream)
...
     log.info("Shutdown request")
-    system.terminate()
-    log.info("Shutting down actor system.")
-    Await.result(system.whenTerminated, Duration.Inf)
-    log.info("Actor system down.")
+    if (!(system eq null)) {
+      system.terminate()
+      log.info("Shutting down actor system.")
+      Await.result(system.whenTerminated, Duration.Inf)
+      log.info("Actor system down.")
```
fommil: feel free to stick a `sys.exit` in here

fommil: actually, gerp the codebase for the other place where we do that. We have a system property guard so we don't exit the JVM when running tests.

dickmao: As in [Server.scala](https://github.com/ensime/ensime-server/blob/3.0/server/src/main/scala/org/ensime/server/Server.scala#L211-L215)?  The bifurcation of servers [here](https://github.com/ensime/ensime-server/blob/3.0/server/src/main/scala/org/ensime/server/Server.scala#L160-L162) guarantees the existence of the actor system in the "regular" but not the "lsp".  A similar null check can be found  [here](https://github.com/ensime/ensime-server/blob/3.0/lsp/src/main/scala/org/ensime/lsp/ensime/EnsimeLanguageServer.scala#L193) and [here](https://github.com/ensime/ensime-server/blob/3.0/lsp/src/main/scala/org/ensime/lsp/ensime/EnsimeLanguageServer.scala#L202)

fommil: I mean a check to see if you're in test mode or not, and then just exit the JVM.

json/src/main/scala/spray/json/JsReader.scala
dickmao: Noob disclaimer applies.

lsp/src/main/scala/org/ensime/lsp/ensime/EnsimeLanguageServer.scala
laughedelic: I think `sys.exit` should be called in the `exit` notification handler, not the `shutdown` request. See https://github.com/ensime/ensime-server/pull/1946.

json/src/main/scala/spray/json/JsReader.scala
```diff
@@ -59,8 +59,8 @@ object JsReader {
   }
 
   implicit def option[A: JsReader]: JsReader[Option[A]] = {
-    case JsNull => None
-    case a      => Some(JsReader[A].read(a))
+    case JsAbsent | JsNull => None
```
laughedelic: Could you also modify the JSON writer to distinguish between an absent value and a `null`?

dicksbu: I'm not sure how to do that.  `JsAbsent` is not writable.

json/src/main/scala/spray/json/JsValue.scala
```diff
@@ -22,6 +22,8 @@ object JsArray {
 
 case object JsNull extends JsValue
 
+case object JsAbsent extends JsValue
```
fommil: oh, hold on, I just realised what happened here. I thought this code was isolated to the lsp layer. This is quite a big change.

I think the better way to do this would be with a custom data type that has special handling in the `DerivedJsReader`. A more general way to do this would be with scalaz tags.

If we're going down this route, it might be best to just start moving efforts towards creating the scalabus, using Alexy's lsp implementation, and have the bus talk to ensime with JERKY.

I don't really want to start hacking the JSON library ... I plan on moving it out into a separate repository as an example of how to do typeclass derivation with scalaz-deriving and the scope of the library is very much a "simple" json library, whereas circe is appropriate for more complex use cases.

Does what I proposed above make sense?

laughedelic: An offtopic sidenote:

> using Alexy's lsp implementation

I see it already second time so I'd like to clarify: if you mean the [lsp4s](https://github.com/scalameta/lsp4s) library, I'm just a mere contributor there, not the primary author.

fommil: to add some more detail into this advice, I mean adding an ADT like

```scala
sealed abstract class Possibly[A]
final case class There[A](a: A) extends Possibly[A]
final case class Not() extends Possibly[A]
final case class Nullary() extends Possibly[A]
```

and then adding a method to `DerivedJsReader` (and the writer) like

```scala
  implicit def hconsPossibly[A, Key <: Symbol, Value, Remaining <: HList](
    implicit Key: Witness.Aux[Key],
    LV: Lazy[JsReader[Value]],
    DR: DerivedJsReader[A, Remaining]
  ): DerivedJsReader[A, FieldType[Key, Possibly[Value]] :*: Remaining] = { m =>
    val read = ... special handling of the three cases goes here
    field[Key](read) :: DR.readObject(m)
  }
```

I know it's shapeless magic, but it's not so bad :smile:

dickmao: I hear you.  I merely wanted to call attention to the distinction between absent values and null values, and that the current implementation will populate a JsNull for both cases.  I did try to come up with a solution that didn't involve `JsAbsent` but this is the first time I've seen shapeless code.  I'll close the PR without further ado.

fommil: it's actually a great point! And something I will add to the JSON library when I generalise it. `null` and absent handling is definitely one of the more frustrating things about json. Nothing quite at the level of XML.

Issue #1900, opened 2017-12-01
finishing touches on s-express and json


fommil:
- all methods should be total, no more `require` or `DeserializationException`
- rename json package to `org.ensime.json`
- use scalaz data types internally instead of stdlib
- define typeclass instances in terms of scalaz typeclasses instead of stdlib wrappers (`Traverse`, `Bitraverse` instead of `Traversable`, `Either`, etc)
- compiletime optimisations with `LowPriorityX` traits? (measure)

PR #1949, opened 2018-06-02
Json total

Tomahna:
So this should go toward the goals of #1900.

I don't really see the point of maintaining a specific json parser for ensime though.


fommil:
Cool! I was planning on doing something like this. I'll take a look in to next few weeks.

Re: fork of json. It is important to aim for zero deps in scala tooling so we can move fast to support new versions of Scala. Especially when the json ast mess was imminent. However, since Scala 3 are going to break everything it's kind of a moot point. We may as well use circe now but I was planning on using scalaz-deriving for fun.

Tomahna:
I can push a PR with Circe. If you have other improvement like this in mind, you can ping me. I'll have a bit of time the next few weeks.

fommil:
There are other battles :-D maybe try nothing up with Daniel on the metacp PR...

PR #1950, opened 2018-06-02
Make s-express reader total

Tomahna:


Tomahna:
Btw the test failed doesn't seem to be due to the PR. So if you can just relaunch it before I investigate further it would be great.
The pipeline worked correctly on my repo.

fommil:
Awesome. I'll probably make some changes soon in this area as I'm planning to use these in my book.

Incidentally, the book uses public domain license on the code. Are you ok if anything you've done ends up being published that way? If you do, that's fine, I can base the code on the old version I wrote.

Tomahna:
No problem, go ahead it's perfectly fine.

PR #1951, opened 2018-06-06
WIP: Replace lsp implementation with lsp4s

zainab-ali:
This is a WIP and shouldn't be reviewed yet, but I wanted to give it some visibility so others didn't repeat the work.

Since last month's [tooling hackday](https://twitter.com/syntaxSugarLdn) I, with the help of @peterneyens, have been working on integrating [lsp4s](https://github.com/scalameta/lsp4s/).  It's a much simpler task than I anticipated, partly due to lsp4s's clean structure (logging seems to be the most complicated aspect).  I expect to have a working implementation over the next few weeks.

There are some changes that may be worth discussing, even in these early stages, notably the dependency on monix.  According to some discussion in #1935, this might be problematic.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -83,6 +83,7 @@ lazy val core = project
       "org.apache.lucene" % "lucene-analyzers-common" % luceneVersion,
       "org.ow2.asm"       % "asm-commons"             % "5.2",
       "org.ow2.asm"       % "asm-util"                % "5.2",
+      // TODO remove the locally built snapshot in favour of a release
       "org.scalameta" %% "lsp4s" % "0.1.0+5-de927af3-SNAPSHOT",
       "org.scala-lang"    % "scalap"                  % scalaVersion.value,
       "com.typesafe.akka" %% "akka-actor"             % akkaVersion,
@@ -221,6 +222,7 @@ addCommandAlias("fix", "all compile:scalafixCli test:scalafixCli")
 addCommandAlias("tests", "all test it:test")
 // not really what is used in CI, but close enough...
 addCommandAlias("ci", ";check ;prep ;cpl ;doc ;tests")
+// TODO The logging between lsp4s (scribe) conflicts with this
 // private val slf4jVersion = "1.7.25"
 //   val logback = Seq(
 //     "ch.qos.logback" % "logback-classic"  % "1.2.3",
--- a/build.sbt
+++ b/build.sbt
@@ -83,6 +83,7 @@ lazy val core = project
       "org.apache.lucene" % "lucene-analyzers-common" % luceneVersion,
       "org.ow2.asm"       % "asm-commons"             % "5.2",
       "org.ow2.asm"       % "asm-util"                % "5.2",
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -83,6 +83,8 @@ lazy val core = project
       "org.apache.lucene" % "lucene-analyzers-common" % luceneVersion,
       "org.ow2.asm"       % "asm-commons"             % "5.2",
       "org.ow2.asm"       % "asm-util"                % "5.2",
+      // TODO remove the locally built snapshot in favour of a release
+      "org.scalameta" %% "lsp4s" % "0.1.0+5-de927af3-SNAPSHOT",
```
olafurpg: There is a v0.1.0 release on Maven http://search.maven.org/#artifactdetails%7Corg.scalameta%7Clsp4s_2.12%7C0.1.0%7Cjar

fommil:
Cool. Just a heads up that this breaks from the convention of minimising the dependency list, which is typically the biggest pain when supporting new versions of scala... but that's not a hill I care to die on.

olafurpg:
For the record, lsp4s brings in a lot of dependencies

- circe
- monix
- scribe
- shapeless (via circe-generic)
- jawn (via circe-parser)
- cats (via circe-core, and soon monix as well)

There's a discussion in https://github.com/scalameta/lsp4s/issues/8 to slim it down (and I gave it an honest try in https://github.com/scalameta/lsp4s/pull/4) but it's not something I expect to prioritize in the coming months.

zainab-ali:
We were aware of it's dependencies, but didn't think that any of them were a sticking point.  Even if they are, I think the plan would be to minimise them.

@fommil would you be happy with this PR depending on the current form of lsp4s, or do you think there are some dependencies we should try to rip out of there first?  I'm not too familiar with the pains of scala version support.

fommil:
@zainab-ali yeah it's great! :smile:

We might want to just use circe everywhere. @Tomahna was suggesting similar recently. Our internal json lib will probably just morph into a case study in my book.

You don't need to check anything with me :grin: ... I'm not really using ensime anymore and I'm not watching the repo (except in a few work projects that benefit from autocomplete) so please merge / release stuff when you think it's ready. I can share sonatype creds for the `ensime` user with any of you, just send me your public gpg key. Of course, I'm happy to help out if you have questions.

olafurpg:
FYI https://github.com/scalameta/lsp4s/releases/tag/v0.2.0

fommil:
@zainab-ali btw I have been writing a jsonformat encoder/decoder typeclass  for my book with examples using magnolia and shapeless. It is able to handle default arguments and user config for `null`. That means users can define this kind of thing

```scala
sealed abstract class Possibly[A]
case class Missing[A]() extends Possibly[A]
case class Found[A](a: A) extends Possibly[A]
```

With `JsonDecoder[Possibly[A]]` always producing a `Found` via `xmap`. And have

```scala
final case class(foo: Possibly[Option[String]] = Missing())
```

which will distinguish between getting a null, not getting anything, and getting a valid value.

Anyways, I don't have plans to bring it into ensime but just food for thought.

PR #1952, opened 2018-06-22
Bump scalazio effect version

Jacoby6000:
Looked like we were quite a bit out of date.

On the bright side, now improperly using `IO` feels about 100 times worse, since you have to extend the runtime.

### DIFF
--- a/api/src/main/scala/org/ensime/io/Canon.scala
+++ b/api/src/main/scala/org/ensime/io/Canon.scala
@@ -18,7 +18,7 @@ import Scalaz._
  * but those that do not must perform IO.
  */
 @typeclass trait Canon[A] { self =>
-  def canon(a: A): IO[A]
+  def canon(a: A): IO[Throwable, A]
 
   final def xmap[B](f: A => B, g: B => A): Canon[B] =
     b => self.canon(g(b)).map(f)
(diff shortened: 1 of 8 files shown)


Jacoby6000:
This is the first of a series of PRs I'll be making to clean up as many of the side-effects as I can..

Tomahna:
Looks good to me.

api/src/main/scala/org/ensime/io/Canon.scala
```diff
@@ -18,15 +18,17 @@ import Scalaz._
  * but those that do not must perform IO.
  */
 @typeclass trait Canon[A] { self =>
-  def canon(a: A): IO[A]
+  def canon(a: A): IO[Throwable, A]
```
fommil: you can use `Task`

api/src/main/scala/org/ensime/io/Canon.scala
```diff
@@ -18,15 +18,17 @@ import Scalaz._
...
 
   final def xmap[B](f: A => B, g: B => A): Canon[B] =
     b => self.canon(g(b)).map(f)
 }
 object Canon extends LowPriorityCanon {
 
   implicit val file: Canon[File] =
-    f => IO.sync(f.getCanonicalFile).catchAll(_ => IO.sync(f.getAbsoluteFile))
+    f =>
+      IO.syncThrowable(f.getCanonicalFile)
```
fommil: `Task.apply` => `IO.syncThrowable`

build.sbt
```diff
@@ -4,7 +4,7 @@ lazy val api = project
     licenses := Seq(LGPL3),
     resourcesOnCompilerCp(Compile),
     libraryDependencies ++= Seq(
-      "org.scalaz" %% "scalaz-ioeffect" % "1.0.0"
+      "org.scalaz" %% "scalaz-ioeffect" % "2.5.0",
```
fommil: I think I'll cut another release in the next few hours

PR #1953, opened 2018-06-27
Get back ENJIME

albfan:
This mostly seems to work. I cannot get autocompletion for:

    Syste

but I get for

    System.ou

captura de pantalla de 2018-06-26 23-55-04

I would like to know what blockers make to decide to remove this feature


fommil:
I vote no. The reason why this was reverted was to encourage somebody (maybe you! :-D) to create a standalone java LSP so that ensime no longer needs to be concerned with java matrix builds.

It's best that LSPs serve one language and other LSPs already exist for Java.

albfan:
That totally makes sense. Should it be a project under ensime group?

fommil:
@albfan only if you think that would help with discoverability. If you get something stood up and working in one of your own repos, I'd be happy to accept it as `ensime` group and give you full admin to it.

albfan:
Nice, I will go for that. Anyway I need to study the codebase (don't know if it share code really) don't want to reinvent the wheel

laughedelic:
I didn't get it, are you going to create another Java LSP server? This would be a reinvention of the wheel: https://github.com/eclipse/eclipse.jdt.ls

fommil:
Not necessarily. The Eclipse LSP uses the Eclipse / IBM AST, whereas the ENSIME codebase uses the one that comes with the JDK.

There would be value in a Java LSP that uses the JDK API.

In addition, the Eclipse LSP has an abhorrent license that was designed to specifically be hostile towards SUN Microsystems (now Oracle) and OpenJDK.

albfan:
I want a barebone LSP for java. That seems to me like start eclipse in background to have a LSP. I don't want to reinvent in terms of ENSIME design

PR #1954, opened 2018-06-30
Handle exceptions thrown by asm

mberndt123:
asm likes to throw exceptions for some class files, like the
module-info.class files generated by Java 9. It's probably
better to log these and carry on than to just abort indexing.


mberndt123:
I strongly suggest to use GitHub's "Hide whitespace changes" feature when reviewing this. It's really not a large change at all.

PR #1960, opened 2018-09-20
Update orientVersion to work with Java 9+

udalrich:
The old version of orient throws an exception trying to do a privileged access which is not allowed in Java 9. That is fixed in the new version.


udalrich:
The travis build failed, but it didn't look like something related to my changes.  Was the build already broken?

Tomahna:
I've restarted the travis build and I'll check if the build is broken on the main branch or not.

